PR #1, opened 2017-02-24
Initial cloudconfig library

nhlfr:
Fixes giantswarm/aws-operator#15
Fixes giantswarm/aws-operator#16


templates/master.yaml
```diff
@@ -0,0 +1,858 @@
...
+      [Unit]
+      Description=k8s-setup-network-env Service
+      Wants=network-online.target docker.service
+      After=network-online.target docker.service
+
+      [Service]
+      Type=oneshot
+      RemainAfterExit=yes
+      TimeoutStartSec=0
+      Environment="IMAGE={{.Spec.Cluster.Docker.ImageNamespace}}/setup-network-env:{{.Spec.Cluster.Operator.NetworkSetupVersion}}
```
xh3b4sd: FYI I think in the long run we want to pin the full image here. It will probably never change. See https://github.com/giantswarm/k8s-setup-network-environment.

templates/master.yaml
```diff
@@ -0,0 +1,858 @@
...
+      [Unit]
+      Description=Timer
+
+      [Timer]
+      OnCalendar=13:00
+      Unit=etcd2-restart.service
+
+      [Install]
+      WantedBy=multi-user.target
+  # TODO(nhlfr): Set up Calico on Kubernetes, in example by http://docs.projectcalico.org/v2.0/getting-started/kubernetes/installation/hosted/kubeadm/calico.yaml.
```
xh3b4sd: +1 go for it. <3

cloudconfig.go
```diff
@@ -0,0 +1,71 @@
...
+	"encoding/base64"
+	"html/template"
+	"io/ioutil"
+
+	"github.com/giantswarm/awstpr"
+	tpraws "github.com/giantswarm/awstpr/aws"
+	"github.com/giantswarm/clustertpr/node"
+)
+
+type AWSNode struct {
```
nhlfr: The reason why we put AWS specific things here is that we get and decrypt keys from AWS blob store. See 

I can create an interface for operator-specific extensions and make things more abstract.

In case if you folks really really don't want to have operator-specific things in this repo, then we will need to create a separate one and provide some way of providing extensions. I don't have any clear picture of this yes - I will need to think about it today.

cloudconfig.go
```diff
@@ -0,0 +1,71 @@
...
+
+type CloudConfig struct {
+	config   string
+	params   CloudConfigTemplateParams
+	path     string
+	template string
+}
+
+func NewCloudConfig(templatePath string, params CloudConfigTemplateParams) (*CloudConfig, error) {
+	cloudConfigTempl, err := ioutil.ReadFile(templatePath)
```
nhlfr: Sorry if I will sound a little bit harsh, but TBH, I don't care very much - I see nothing wrong in having a template file and just reading it. I also don't get the argument about "having hard runtime dependencies", since reading the file which is in the repo brings no runtime dependencies - even on OS X if that's what you meant.

I think that having the content of template in go code isn't a very good idea since it's very large and we will lose benefits of having a separate yaml file - first of them is that text editors will not provide yaml-specific colors or features if we put this in Go code.

If that will make you happier, I can move reading file out from this function and use afero, but as I wrote above, I'm struggling to see any point of that.

templates/master.yaml
```diff
@@ -0,0 +1,858 @@
...
+          #  args:
+          #  - --v=2
+          #  - --logtostderr
+          #  ports:
+          #  - containerPort: 10054
+          #    name: metrics
+          #    protocol: TCP
+          #  resources:
+          #    requests:
+          #      memory: 10Mi
```
nhlfr: done

templates/master.yaml
```diff
@@ -0,0 +1,858 @@
...
+        --cacert /etc/kubernetes/ssl/apiserver-ca.pem --cert /etc/kubernetes/ssl/apiserver.pem --key /etc/kubernetes/ssl/apiserver-key.pem \
+        "https://{{.Spec.Cluster.Kubernetes.API.Domain}}:443/api/v1/namespaces/kube-system/services"
+
+      echo "K8S: Calico Policy"
+      curl -H "Content-Type: application/json" \
+        -XPOST -d"$(cat /srv/calico-system.json)" \
+        --cacert /etc/kubernetes/ssl/apiserver-ca.pem --cert /etc/kubernetes/ssl/apiserver.pem --key /etc/kubernetes/ssl/apiserver-key.pem \
+        "https://{{.Spec.Cluster.Kubernetes.API.Domain}}:443/api/v1/namespaces/"
+
+      echo "Addons successfully installed"
```
nhlfr: done

cloudconfig.go
iaguis: I think the point was for example to have it get a `io.Reader` so you can use mock cloudconfigs and run unit tests against the package. Maybe afero is a bit overkill for this.

nhlfr: Ah, ok, yes that indeed makes sense to me.

nhlfr: However, while I have nothing against writing mock cloudconfigs to test, I think we should still test also the actual cloudconfigs which we have in this repo. That would be very helpful for checking whether the changes in cloudconfig files are correct.

nhlfr:
Back to WIP - I'm trying to move AWS- and KVM-specific things to extensions.

cloudconfig.go
xh3b4sd: I don't understand the problem. The interface is the template. The template contains a cloud config. The parts that are the same everywhere are hard coded in the template. The parts that need to be dynamically injected will be injected via template data. This is the interface. Lets talk if something is not clear.

nhlfr:
@xh3b4sd @JosephSalisbury @iaguis @asymmetric PTAL again

Maybe the current implementation of "extensions" isn't ideal, however feel free to comment it and share your thoughts.

operator.go
```diff
@@ -0,0 +1,149 @@
...
+
+type UnitAsset struct {
+	Metadata UnitMetadata
+	Content  []string
+}
+
+type OperatorExtension interface {
+	GetFiles() ([]FileAsset, error)
+	GetUnits() ([]UnitAsset, error)
+}
```
xh3b4sd: I like that interface. It makes sense for the idea of having an abstracted cloud config manager. If we like we can extend this in the future with authorized keys and things like this. Only two things toward that. 

- IMO there should not be anything AWS/KVM related inside of this project. I feel like my comments are unclear or something. We maybe should talk about this again. I do not understand why there is still code that only the AWS operator cares about in here. The implementation of the interface should be done inside of the AWS operator. 
- I would like to align the naming to the golang convention. Getters do not have the `get` prefix. See https://golang.org/doc/effective_go.html#Getters.

nhlfr: Just please propose clearly *where* you want to put AWS/KVM related templates and code. In a comment. :)

Makefile
```diff
@@ -0,0 +1,27 @@
...
+	glide-vc --use-lock-file
+
+update-vendor: verify-glide-installation
+	glide update --strip-vendor
+	glide-vc --use-lock-file
+
+verify-go-bindata-installation:
+	@which go-bindata || go get -u github.com/jteeuwen/go-bindata/...
+
+update-bindata: verify-go-bindata-installation
```
nhlfr: It's used manually by someone who edits the template files.

What we may check in CircleCI is whether bindata is up to date (so probably generate bindata and compare with the bindata.go file which is commited), but I think we shouldn't do it in this PR - it's already huge..

operator.go
xh3b4sd: Sure, see the comments I already wrote. 

- https://github.com/giantswarm/k8scloudconfig/pull/1#discussion_r102953453
- https://github.com/giantswarm/k8scloudconfig/pull/1#discussion_r102967916
- https://github.com/giantswarm/k8scloudconfig/pull/1#discussion_r103713751
- https://github.com/giantswarm/aws-operator/pull/14#discussion_r102486768

Ones again. Every code that is concerned about AWS specific logic, belongs to the AWS operator. So please put such code into this repository: https://github.com/giantswarm/aws-operator. Every code that is concerned about KVM specific logic, belongs to the KVM operator. So please put such code into this repository: https://github.com/giantswarm/kvm-operator. Right now I cannot give you a concrete file path, because this depends on the actual code. Anyway, this repository here is concerned with cloud config management. It should aim to provide an abstracted interface everybody can implement. So the implementation of the interface should be put into the separate operator repositories. That way this repository stays lightweight and abstracted. It does not care about AWS/KVM. It does only dumb an cares about cloud config generation. Every detail that is concerned about specialties of some operator should be provided as argument, where the argument represents an abstraction as well. You already defined the abstract interface `OperatorExtension`, which only cares about files and units. This is super awesome. So lets put the code that implements this interface into the operator that actually cares about the details of the files and units that it wants to see inside the cloud config. So this is my proposal. When you have another idea we can also talk about this.

templates/master.yaml
```diff
@@ -0,0 +1,858 @@
...
+
+- path: /etc/kubernetes/ssl/calico/client-ca.pem.enc
+  encoding: gzip+base64
+  content: {{.TLSAssets.CalicoClientCACert}}
+
+- path: /etc/kubernetes/ssl/calico/client-key.pem.enc
+  encoding: gzip+base64
+  content: {{.TLSAssets.CalicoClientKey}}
+
+- path: /etc/kubernetes/ssl/etcd/server.pem.enc
```
nhlfr: done, dunno why github still shows this diff

Makefile
xh3b4sd: I would say lets make smaller PRs. This was anyway what we wanted to do. So lets have smaller PRs.

operator.go
nhlfr: OK, got it. I removed all AWS/KVM-related stuff and left only:
- interface
- a function for templating assets (+ written a test for it)

xh3b4sd:
I noticed that there is no circle build. We have to sort this out as well.

operator.go
xh3b4sd: I would really like to fix the naming if there is no better reason to keep it as it is. See

> I would like to align the naming to the golang convention. Getters do not have the get prefix. See https://golang.org/doc/effective_go.html#Getters.

nhlfr:
@xh3b4sd fixed the naming of methods in the interface

PR #3, opened 2017-03-02
operator: Use external asset contents instead of internal assets

nhlfr:
External extensions implementes in operator repos (like aws-operator) will use their own templates, that's why we need to allow to template them here.

### DIFF
--- a/operator.go
+++ b/operator.go
@@ -35,13 +35,8 @@ type OperatorExtension interface {
 	Units() ([]UnitAsset, error)
 }
 
-func RenderAssetContent(assetPath string, params interface{}) ([]string, error) {
-	rawContent, err := Asset(assetPath)
-	if err != nil {
-		return nil, err
-	}
-
-	tmpl, err := template.New(assetPath).Parse(string(rawContent[:]))
+func RenderAssetContent(rawAssetContent []byte, params interface{}) ([]string, error) {
+	tmpl, err := template.New("").Parse(string(rawAssetContent[:]))
 	if err != nil {
 		return nil, err
 	}
--- a/operator_test.go
+++ b/operator_test.go
@@ -24,7 +24,12 @@ func TestRenderAssetContent(t *testing.T) {
 	}
 
 	for _, tc := range tests {
-		content, err := RenderAssetContent(tc.assetPath, tc.params)
+		rawAssetContent, err := Asset(tc.assetPath)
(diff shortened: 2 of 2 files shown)


nhlfr:
That's a quick fix which will unlock the aws-operator extension.

operator.go
```diff
@@ -35,13 +35,8 @@ type OperatorExtension interface {
...
 }
 
-func RenderAssetContent(assetPath string, params interface{}) ([]string, error) {
-	rawContent, err := Asset(assetPath)
-	if err != nil {
-		return nil, err
-	}
-
-	tmpl, err := template.New(assetPath).Parse(string(rawContent[:]))
+func RenderAssetContent(assetContent []byte, params interface{}) ([]string, error) {
```
iaguis: Can you call it `rawAssetContent`? Same for the tests.

nhlfr: I want to expose this function for `aws-operator`, so not really.

iaguis: I mean the parameter, not the function name.

nhlfr: Ah, sorry. Done.

operator.go
```diff
@@ -35,13 +35,8 @@ type OperatorExtension interface {
...
 
-func RenderAssetContent(assetPath string, params interface{}) ([]string, error) {
-	rawContent, err := Asset(assetPath)
-	if err != nil {
-		return nil, err
-	}
-
-	tmpl, err := template.New(assetPath).Parse(string(rawContent[:]))
+func RenderAssetContent(rawAssetContent []byte, params interface{}) ([]string, error) {
+	tmpl, err := template.New("").Parse(string(assetContent[:]))
```
iaguis: missed this bit ;)

P.S.: We need CircleCI ASAP

operator_test.go
```diff
@@ -24,7 +24,12 @@ func TestRenderAssetContent(t *testing.T) {
 	}
 
 	for _, tc := range tests {
-		content, err := RenderAssetContent(tc.assetPath, tc.params)
+		assetContent, err := Asset(tc.assetPath)
```
iaguis: please change the name here too to `rawAssetContent`

operator.go
nhlfr: done

operator_test.go
nhlfr: done

PR #6, opened 2017-03-09
Use text/template, not html/template

iaguis:
If we use html/template, text will be escaped to make it safe.

This is not what we want, we're not dealing with html and it breaks
things like base64-encoded keys and such.

### DIFF
--- a/cloudconfig.go
+++ b/cloudconfig.go
@@ -4,7 +4,7 @@ import (
 	"bytes"
 	"compress/gzip"
 	"encoding/base64"
-	"html/template"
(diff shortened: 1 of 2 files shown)


iaguis:
This bit me today and I spent some time trying to figure out the issue 😞

iaguis:
Ping.

iaguis:
CircleCI seems broken :/

JosephSalisbury:
@iaguis try merging master plz

iaguis:
Done

PR #9, opened 2017-03-13
Changes to master.yaml

asymmetric:
**NOTE:** This PR is used to separate changes which apply to the AWS cloudconfig only from those that apply more generally. As such, there are changes which might not make a lot of sense. Those changes will be reverted and removed from this branch. Those that apply more generally will be left in.

--

These are the changes to `master.yaml` as of https://github.com/kinvolk/aws-operator/commit/7f04717c17db314bdfaa7cb5b240859d25532e2d

This PR is a WIP and not ready to be merged.

### DIFF
--- a/templates/master.yaml
+++ b/templates/master.yaml
@@ -413,11 +413,11 @@ write_files:
 
 - path: /etc/kubernetes/ssl/apiserver.pem.enc
   encoding: gzip+base64
-  content: {{.TLSAssets.APIServerCert}}
+  content: {{.TLSAssets.APIServerCrt}}
 
 - path: /etc/kubernetes/ssl/apiserver-ca.pem.enc
   encoding: gzip+base64
-  content: {{.TLSAssets.APIServerCACert}}
+  content: {{.TLSAssets.APIServerCACrt}}
 
 - path: /etc/kubernetes/ssl/apiserver-key.pem.enc
   encoding: gzip+base64
@@ -425,11 +425,11 @@ write_files:
 
 - path: /etc/kubernetes/ssl/calico/client.pem.enc
   encoding: gzip+base64
-  content: {{.TLSAssets.CalicoClientCert}}
+  content: {{.TLSAssets.CalicoClientCrt}}
 
 - path: /etc/kubernetes/ssl/calico/client-ca.pem.enc
   encoding: gzip+base64
-  content: {{.TLSAssets.CalicoClientCACert}}
+  content: {{.TLSAssets.CalicoClientCACrt}}
 
 - path: /etc/kubernetes/ssl/calico/client-key.pem.enc
   encoding: gzip+base64
@@ -437,11 +437,11 @@ write_files:
 
 - path: /etc/kubernetes/ssl/etcd/server.pem.enc
   encoding: gzip+base64
-  content: {{.TLSAssets.EtcdServerCert}}
+  content: {{.TLSAssets.EtcdServerCrt}}
 
 - path: /etc/kubernetes/ssl/etcd/server-ca.pem.enc
   encoding: gzip+base64
-  content: {{.TLSAssets.EtcdServerCACert}}
+  content: {{.TLSAssets.EtcdServerCACrt}}
 
 - path: /etc/kubernetes/ssl/etcd/server-key.pem.enc
   encoding: gzip+base64
--- a/templates/master.yaml
+++ b/templates/master.yaml
@@ -507,9 +507,8 @@ coreos:
     content: |
       [Unit]
       Description=etcd2
-      # TODO can we move this to a Wants in the decrypt-tls-certs.service ?
-      Requires=k8s-setup-network-env.service decrypt-tls-certs.service
-      After=k8s-setup-network-env.service decrypt-tls-certs.service
+      Requires=k8s-setup-network-env.service
+      After=k8s-setup-network-env.service
       Conflicts=etcd.service
       Wants=calico-node.service
       StartLimitIntervalSec=0
@@ -571,8 +570,8 @@ coreos:
     content: |
       [Unit]
       Description=Calico per-host agent
-      Requires=etcd2.service decrypt-tls-certs.service
-      After=etcd2.service decrypt-tls-certs.service
+      Requires=etcd2.service
+      After=etcd2.service
       Wants=k8s-api-server.service k8s-controller-manager.service k8s-scheduler.service
       StartLimitIntervalSec=0
 
@@ -944,17 +943,5 @@ coreos:
 
       [Install]
       WantedBy=multi-user.target
-  - name: decrypt-tls-certs.service
-    enable: true
-    command: start
-    content: |
-      [Unit]
-      Description=Decrypt TLS certificates
-
-      [Service]
-      ExecStart=/opt/bin/decrypt-tls-assets
-      ExecStartPost=/usr/bin/chown -R etcd:etcd /etc/kubernetes/ssl/etcd
-      # TODO this is a hack - which user needs to read this? etcd already has read
-      ExecStartPost=/usr/bin/chmod go+r /etc/kubernetes/ssl/etcd
   update:
     reboot-strategy: off
--- a/templates/master.yaml
+++ b/templates/master.yaml
@@ -943,5 +943,12 @@ coreos:
 
       [Install]
       WantedBy=multi-user.target
+
+  {{range .Units}}- name: {{.Metadata.Name}}
+    enable: {{.Metadata.Enable}}
+    command: {{.Metadata.Command}}
+    content: |
+      {{range .Content}}{{.}}
+      {{end}}{{end}}
   update:
     reboot-strategy: off
--- a/templates/master.yaml
+++ b/templates/master.yaml
@@ -6,7 +6,7 @@ write_files:
   owner: root
   content: |
     127.0.0.1 localhost
-    127.0.0.1 {{.Node.Hostname}}
+    127.0.0.1 {{.Node.Node.Hostname}}
     127.0.0.1 etcd.giantswarm
 - path: /etc/resolv.conf
   permissions: 0644
@@ -64,9 +64,9 @@ write_files:
             args:
             # command = "/kube-dns
             - --dns-port=10053
-            - --domain={{.Cluster.Kubernetes.Domain}}
+            - --domain={{.Spec.Kubernetes.Domain}}
             - --kubecfg-file=/etc/kubernetes/config/kubelet-kubeconfig.yml
-            - --kube-master-url=https://{{.Cluster.Kubernetes.API.Domain}}
+            - --kube-master-url=https://{{.Spec.Kubernetes.API.Domain}}
             ports:
             - containerPort: 10053
               name: dns-local
@@ -108,6 +108,32 @@ write_files:
               requests:
                 cpu: 150m
                 memory: 10Mi
+            #### Commented cause we need Prometheus integrated in Kubernetes ####
+            #env:
+            #- name: PROMETHEUS_PORT
+            #  value: "10055"
+          #### Commented cause we need Prometheus integrated in Kubernetes ####
+          #- name: dnsmasq-metrics
+          #  image: gcr.io/google_containers/dnsmasq-metrics-amd64:1.0
+          #  livenessProbe:
+          #    httpGet:
+          #      path: /metrics
+          #      port: 10054
+          #      scheme: HTTP
+          #    initialDelaySeconds: 60
+          #    timeoutSeconds: 5
+          #    successThreshold: 1
+          #    failureThreshold: 5
+          #  args:
+          #  - --v=2
+          #  - --logtostderr
+          #  ports:
+          #  - containerPort: 10054
+          #    name: metrics
+          #    protocol: TCP
+          #  resources:
+          #    requests:
+          #      memory: 10Mi
           - name: healthz
             image: gcr.io/google_containers/exechealthz-amd64:1.2
             resources:
@@ -118,7 +144,7 @@ write_files:
                 cpu: 10m
                 memory: 50Mi
             args:
-            - -cmd=nslookup kubernetes.default.svc.{{.Cluster.Kubernetes.Domain}} 127.0.0.1 >/dev/null && nslookup kubernetes.default.svc.{{.Cluster.Kubernetes.Domain}} 127.0.0.1:10053 >/dev/null
+            - -cmd=nslookup kubernetes.default.svc.{{.Spec.Kubernetes.Domain}} 127.0.0.1 >/dev/null && nslookup kubernetes.default.svc.{{.Spec.Kubernetes.Domain}} 127.0.0.1:10053 >/dev/null
             - -port=8080
             - -quiet
             ports:
@@ -148,7 +174,7 @@ write_files:
     spec:
       selector:
         k8s-app: kube-dns
-      clusterIP: {{.Cluster.Kubernetes.DNS.IP}}
+      clusterIP: {{.Spec.Kubernetes.DNS}}
       ports:
       - name: dns
         port: 53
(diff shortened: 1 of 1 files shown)


templates/master.yaml
```diff
@@ -179,198 +205,31 @@ write_files:
...
-        port: 443
-        nodePort: 30011
-      selector:
-        app: ingress-controller
 - path: /opt/k8s-addons
   permissions: 0544
   content: |
       #!/bin/bash
-      while ! curl --output /dev/null --silent --head --fail --cacert /etc/kubernetes/ssl/apiserver-ca.pem --cert /etc/kubernetes/ssl/apiserver.pem --key /etc/kubernetes/ssl/apiserver-key.pem "https://{{.Cluster.Kubernetes.API.Domain}}:443"; do sleep 1 && echo 'Waiting for master'; done
+      while ! curl --output /dev/null --silent --head --fail "http://{{.Spec.Kubernetes.API.Domain}}:{{.Spec.Kubernetes.API.InsecurePort}}"; do sleep 1 && echo 'Waiting for master'; done
```
nhlfr: Good change, but s/Spec/Cluster

templates/master.yaml
```diff
@@ -415,52 +274,91 @@ write_files:
         user: scheduler
       name: service-account-context
     current-context: service-account-context
-
-- path: /etc/kubernetes/ssl/apiserver-crt.pem.enc
+- path: /opt/bin/decrypt-tls-assets
```
nhlfr: aws-specfic thing - it was already moved to the aws-operator

templates/master.yaml
```diff
@@ -415,52 +274,91 @@ write_files:
...
+           mv -f $f ${encKey%.enc}
+         done;
+         echo done.'
+
+    rkt rm --uuid-file=/var/run/coreos/decrypt-tls-assets.uuid || :
+
+- path: /etc/kubernetes/ssl/apiserver.pem.enc
   encoding: gzip+base64
-  content: {{.TLSAssets.APIServerCrt}}
+  content: {{.TLSAssets.APIServerCert}}
```
nhlfr: s/Cert/Crt

templates/master.yaml
```diff
@@ -415,52 +274,91 @@ write_files:
...
-{{range .Files}}- path: {{.Metadata.Path}}
-  owner: {{.Metadata.Owner}}
-  permissions: {{printf "%#o" .Metadata.Permissions}}
-  content: |
-    {{range .Content}}{{.}}
-    {{end}}{{end}}
-
 coreos:
   units:
+  - name: add-route-bridge.service
```
nhlfr: that's kvm-specific thing for flannel, should be moved to kvm-operator

summon @rossf7 @xh3b4sd

templates/master.yaml
```diff
@@ -482,7 +381,7 @@ coreos:
     - name: 10-giantswarm-extra-args.conf
       content: |
         [Service]
-        Environment="DOCKER_CGROUPS=--exec-opt native.cgroupdriver=systemd {{.Cluster.Docker.Daemon.ExtraArgs}}
+        Environment="DOCKER_CGROUPS=--exec-opt native.cgroupdriver=systemd {{.Spec.Docker.Daemon.ExtraArgs}}
```
nhlfr: s/Spec/Cluster

templates/master.yaml
```diff
@@ -658,16 +547,16 @@ coreos:
...
+      StartLimitIntervalSec=0
 
       [Service]
       Restart=always
       RestartSec=0
       TimeoutStopSec=10
-      StartLimitIntervalSec=0
       EnvironmentFile=/etc/network-environment
-      Environment="IMAGE={{.Cluster.Kubernetes.Hyperkube.Docker.Image}}"
+      Environment="IMAGE={{.Spec.Docker.ImageNamespace}}/hyperkube:{{.Spec.Kubernetes.Hyperkube.Version}}"
```
nhlfr: Path to that thing is probably much different now, let me find it later (or you can try to find faster :D )

templates/master.yaml
```diff
@@ -944,11 +848,17 @@ coreos:
 
       [Install]
       WantedBy=multi-user.target
-  {{range .Units}}- name: {{.Metadata.Name}}
```
nhlfr: please give the templating parts back ;)

templates/master.yaml
```diff
@@ -944,11 +848,17 @@ coreos:
...
-    enable: {{.Metadata.Enable}}
-    command: {{.Metadata.Command}}
+  - name: decrypt-tls-certs.service
+    enable: true
+    command: start
     content: |
-      {{range .Content}}{{.}}
-      {{end}}{{end}}
+      [Unit]
```
nhlfr: ant that was moved to aws-operator

templates/master.yaml
```diff
@@ -179,198 +205,31 @@ write_files:
...
       },
-      "description": "fication for a network isolation policy",
+      "description": "Specification for a network isolation policy",
       "versions": [
         {
           "name": "v1alpha1"
         }
       ]
     }
-- path: /srv/fallback-server-dep.yml
```
xh3b4sd: IMO we still need the ingress controllers and their fallback servers inside the guest clusters. So I think we have to keep them. @JosephSalisbury has a better overview about this.

templates/master.yaml
```diff
@@ -179,198 +205,31 @@ write_files:
...
-
-      echo "K8S: Ingress Controller"
-      curl -H "Content-Type: application/yaml" \
-        -XPOST -d"$(cat /srv/ingress-controller-dep.yml)" \
-        --cacert /etc/kubernetes/ssl/apiserver-ca.pem --cert /etc/kubernetes/ssl/apiserver.pem --key /etc/kubernetes/ssl/apiserver-key.pem \
-        "https://%%K8S_MASTER_DOMAIN_NAME%%:443/apis/extensions/v1beta1/namespaces/kube-system/deployments"
-      curl -H "Content-Type: application/yaml" \
-        -XPOST -d"$(cat /srv/ingress-controller-svc.yml)" \
-        --cacert /etc/kubernetes/ssl/apiserver-ca.pem --cert /etc/kubernetes/ssl/apiserver.pem --key /etc/kubernetes/ssl/apiserver-key.pem \
-        "https://%%K8S_MASTER_DOMAIN_NAME%%:443/api/v1/namespaces/kube-system/services"
```
xh3b4sd: @JosephSalisbury please have a look because of the IC and FS changes.

templates/master.yaml
xh3b4sd: +1 this should not be in here.

templates/master.yaml
xh3b4sd: Good catch. See https://github.com/giantswarm/kvm-operator/issues/64.

templates/master.yaml
```diff
@@ -496,7 +395,7 @@ coreos:
       Type=oneshot
       RemainAfterExit=yes
       TimeoutStartSec=0
-      Environment="IMAGE={{.Cluster.Operator.NetworkSetup.Docker.Image}}
+      Environment="IMAGE={{.Spec.Docker.ImageNamespace}}/setup-network-env:{{.Spec.Operator.NetworkSetupVersion}}"
```
xh3b4sd: It should be `.Cluster.Operator.NetworkSetup.Docker.Image`. See https://github.com/giantswarm/clustertpr/blob/47bae764f0a168b9dac27e274515b1a5d03eea6b/operator/networksetup/docker/docker.go#L4.

templates/master.yaml
```diff
@@ -512,31 +411,26 @@ coreos:
     content: |
       [Unit]
       Description=etcd2
-      Requires=k8s-setup-network-env.service
-      After=k8s-setup-network-env.service
+      # TODO can we move this to a Wants in the decrypt-tls-certs.service ?
+      Requires=k8s-setup-network-env.service decrypt-tls-certs.service
```
xh3b4sd: I thought the `decrypt-tls-certs.service` is AWS specific and thus has to be injected via the extensions API. A dependency the other way around then might work. So `decrypt-tls-certs.service` has a `Wants` and an `After` on the `k8s-setup-network-env.service` if this is necessary.

templates/master.yaml
```diff
@@ -512,31 +411,26 @@ coreos:
...
       EnvironmentFile=/etc/network-environment
-      ExecStartPre=/usr/bin/etcdctl --endpoint=http://{{.Node.Hostname}}:2379 set /k8s/master/{{.Cluster.Etcd.Prefix}} '${DEFAULT_IPV4}'
       ExecStartPre=/bin/bash -c "while [ ! -f /etc/kubernetes/ssl/etcd/server-ca.pem ]; do echo 'Waiting for /etc/kubernetes/ssl/etcd/server-ca.pem to be written' && sleep 1; done"
       ExecStartPre=/bin/bash -c "while [ ! -f /etc/kubernetes/ssl/etcd/server.pem ]; do echo 'Waiting for /etc/kubernetes/ssl/etcd/server.pem to be written' && sleep 1; done"
       ExecStartPre=/bin/bash -c "while [ ! -f /etc/kubernetes/ssl/etcd/server-key.pem ]; do echo 'Waiting for /etc/kubernetes/ssl/etcd/server-key.pem to be written' && sleep 1; done"
-      ExecStart=/usr/bin/etcd2 --advertise-client-urls=https://{{.Cluster.Etcd.Domain}}:443,http://127.0.0.1:2383 \
-                               --data-dir=/etc/kubernetes/data/etcd/ \
-                               --initial-advertise-peer-urls=https://{{.Cluster.Etcd.Domain}}:443 \
-                               --listen-client-urls=https://0.0.0.0:2379,http://127.0.0.1:2383 \
-                               --listen-peer-urls=https://${DEFAULT_IPV4}:2380 \
```
xh3b4sd: This all should not change. Why do you want it to change? If we change how etcd works for the guest clusters we should only work on separating it and run etcd clusters via the etcd-operator. This can be done in a next step, but until then we should keep it as it is.

templates/master.yaml
```diff
@@ -512,31 +411,26 @@ coreos:
...
       [Service]
       User=etcd
       Type=notify
       Restart=always
       RestartSec=0
       TimeoutStopSec=10
-      StartLimitIntervalSec=0
       LimitNOFILE=40000
       EnvironmentFile=/etc/network-environment
-      ExecStartPre=/usr/bin/etcdctl --endpoint=http://{{.Node.Hostname}}:2379 set /k8s/master/{{.Cluster.Etcd.Prefix}} '${DEFAULT_IPV4}'
```
xh3b4sd: This is necessary to be able to update the Kubernetes endpoints of the master and worker nodes of the guest clusters. So this should also not change. Why do you want to get rid of it?

templates/master.yaml
```diff
@@ -576,39 +469,35 @@ coreos:
 
       [Install]
       WantedBy=multi-user.target
-  # TODO(nhlfr): Set up Calico on Kubernetes, in example by http://docs.projectcalico.org/v2.0/getting-started/kubernetes/installation/hosted/kubeadm/calico.yaml.
-  # Or at least use anything which doesn't download binaries in systemd unit...
```
xh3b4sd: This TODO is pretty valid. We should only remove it when it is addressed. There should also be an issue for this. Feel free to work on this topic next. For now I think we should keep it.

templates/master.yaml
```diff
@@ -729,22 +617,36 @@ coreos:
...
+      [Unit]
+      Description=k8s-key-generator Service
+
+      [Service]
+      Type=oneshot
+      ExecStartPre=/usr/bin/mkdir -p /etc/kubernetes/secrets
+      ExecStart=/usr/bin/openssl genrsa -out /etc/kubernetes/secrets/token_sign_key.pem 2048
+
+      [Install]
+      WantedBy=multi-user.target
```
xh3b4sd: What is this used for? Why do we need it?

templates/master.yaml
```diff
@@ -729,22 +617,36 @@ coreos:
...
+      StartLimitIntervalSec=0
 
       [Service]
       Restart=always
       RestartSec=0
       TimeoutStopSec=10
-      StartLimitIntervalSec=0
       EnvironmentFile=/etc/network-environment
-      Environment="IMAGE={{.Cluster.Kubernetes.Hyperkube.Docker.Image}}"
+      Environment="IMAGE={{.Spec.Docker.ImageNamespace}}/hyperkube:{{.Spec.Kubernetes.Hyperkube.Version}}"
```
xh3b4sd: This is backwards again. Please use the master of the clustertpr.

templates/master.yaml
```diff
@@ -756,10 +658,12 @@ coreos:
       -v /etc/kubernetes/secrets/token_sign_key.pem:/etc/kubernetes/secrets/token_sign_key.pem \
       $IMAGE \
       /hyperkube controller-manager \
-      --master=https://{{.Cluster.Kubernetes.API.Domain}}:443 \
+      --master=http://{{.Spec.Kubernetes.API.Domain}}:{{.Spec.Kubernetes.API.InsecurePort}} \
```
xh3b4sd: This is also backwards. We should keep the secured connection.

templates/master.yaml
rossf7: kvm-operator is still using k8s-vm. If this needs to be removed I can raise a kvm-operator issue so we don't forget to add it when we switch to k8scloudconfig. @xh3b4sd WDYT?

templates/master.yaml
asymmetric: This wasn't actively removed by me, it just wasn't present in [the copy of cloudconfig I based my work on](https://github.com/kinvolk/aws-operator/blob/iaguis/aws/master-cloudconfig.yaml#L215).

So not every change in this PR has reasoning behind it.

asymmetric: Unfortunately, I should add :). That's why this is a WIP. We are trying to work out a thorny situation in which we worked on the same code across multiple branches (and repos!).

templates/master.yaml
nhlfr: Good catch, shame on me that I missed that. :o

templates/master.yaml
nhlfr: +1

nhlfr:
@xh3b4sd 
>I do not understand why we want to go back to unencrypted HTTP communication between the guest cluster k8s components.

That's right. We have two choices regarding that:

- talk to the `localhost` in master services, and leave the traffic unecryptedm, via HTTP
- go back to HTTPS

I prefer the first option, because it would be much better in HA setups. When talking to the domain instead of localhost, you send a traffic through the load balancer, which may use another node. Redirecting a traffic from the one machine to another (or the same!) through LA sounds bad to me, it's better to just talk to the loopback device.

templates/master.yaml
asymmetric: @xh3b4sd @JosephSalisbury I am gonna readd that, as it was removed as a result of trying to merge branches that have diverged long time ago - not because we thought it should be removed.

JosephSalisbury: Yeah, we still need the ingress controller / fallback stuff.

templates/master.yaml
asymmetric: @xh3b4sd in our setup we don't currently have an outer etcd instance to talk to, so this fails.

templates/master.yaml
asymmetric: @xh3b4sd this is another case where we have conflicting needs. I removed those settings in my branch to revert to the defaults, because in our setup, the etcd instance is running on the master, so it can't use port 443 (which is used by the k8s API).

I'll open an issue to discuss how we should deal with situations like this one.

templates/master.yaml
xh3b4sd: I have this on my list. See https://github.com/giantswarm/kvm-operator/issues/65 for tracking. <3

templates/master.yaml
xh3b4sd: How do we want to do this then? I think the IP propagation and the endpoint updates have to be done as well on AWS. Is this right? I think we should use the same approaches across operators and only make differences where the underlying provider enforces this by its owm primitives. Like AWS has KMS. Fine. But in this case I think we have to update the endpoints with the actual IPs anyway:

templates/master.yaml
xh3b4sd: IMO we should not diverge solutions here like I stated in another comment above. The way etcd works within guest clusters is ugly but it works and should be the same for each operator. The story to succeed the current etcd situation is to remove the single etcd node from the master and run an etcd cluster via the etcd-operator. This should then apply in general and count for AWS and KVM.

xh3b4sd:
I am fine with doing the communication of local components via localhost. This does not imply to fallback to HTTP. We can talk HTTPS locally as well. This is not critical, but would be nice at the end.

templates/master.yaml
```diff
@@ -414,7 +411,7 @@ write_files:
       name: service-account-context
     current-context: service-account-context
 
-- path: /etc/kubernetes/ssl/apiserver-crt.pem.enc
+- path: /etc/kubernetes/ssl/apiserver.pem.enc
```
asymmetric: @xh3b4sd what's the correct naming scheme here? [This issue](https://github.com/giantswarm/aws-operator/issues/18) says without `-crt`, so just checking to make sure.

xh3b4sd: We want it to be consistent with this scheme. 

```
foo-ca.pem
foo-crt.pem
foo-key.pem
```

```
fooCA
fooCrt
fooKey
```

templates/master.yaml
asymmetric: Those keys are [already present](https://github.com/giantswarm/k8scloudconfig/blob/58aa6022a5209fecfdb0b6e85f1e2af50a080cd2/templates/master.yaml#L678) on the machines in the KVM setup I suppose. We need to generate them.

nhlfr:
@asymmetric Is everything from this branch already included in the other pull requests? If yes, let's close it. ;)

PR #17, opened 2017-03-16
bindata: correct files

iaguis:
By mistake, they didn't contain the right data. For example, the master
cloudconfig file started with "#cloud-configz", which caused
coreos-cloudinit to reject it.


iaguis:
I'll follow up with a PR that makes sure this doesn't happen again by adding a new test on CircleCI.

PR #18, opened 2017-03-16
Makefile: make bindata file deterministic

iaguis:
And check on CircleCI that it has the right data.

Depends on #17

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ verify-go-bindata-installation:
 
 update-bindata: verify-go-bindata-installation
 	find templates/ -type f -exec perl -i -pe "chomp if eof" {} \;
-	go-bindata -pkg cloudconfig templates/...
+	go-bindata -modtime 1 -mode 420 -pkg cloudconfig templates/...
 
 check:
 	go test `glide novendor`
--- a/circle.yml
+++ b/circle.yml
@@ -6,6 +6,7 @@ machine:
         GOOS: "linux"
         GOARCH: "amd64"
         GOPATH: "$HOME/go"
+        PATH: "$PATH:$HOME/go/bin"
         GS_WD: "$HOME/go/src/github.com/$CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME"
 
 dependencies:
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -21,7 +21,7 @@ verify-go-bindata-installation:
 
 update-bindata: verify-go-bindata-installation
 	find templates/ -type f -exec perl -i -pe "chomp if eof" {} \;
-	go-bindata -pkg cloudconfig templates/...
+	go-bindata -modtime 1 -mode 420 -pkg cloudconfig templates/...
```
xh3b4sd: What does `-modtime 1` actually do? Would it make sense to set the modification timestamp to a real timestamp or how does this work?

circle.yml
```diff
@@ -20,3 +20,7 @@ test:
     override:
         - cd $GS_WD && go test -v $(glide novendor)
         - cd $GS_WD && go build -a -v -tags netgo -ldflags "-w -X main.gitCommit=$(git rev-parse --short HEAD)"
+        - cd $GS_WD &&
+          cp bindata.go bindata-repo.go &&
+          make update-bindata &&
+          cmp bindata.go bindata-repo.go
```
xh3b4sd: AFAIK `cmp` fails when the files are not equal. `bindata.go` and `bindata-repo.go` should be different since `bindata.go` was regenerated by `update-bindata`. I feel like I miss something.

iaguis: I'll try to explain.

First, we can't generate `bindata.go` on the CI because it needs to be vendored in other repos, and for that to work we need the data to be commited to the repo.

https://github.com/giantswarm/k8scloudconfig/pull/18/commits/d8ef102f91db459a62c09441fb2e0efa76b29ccb makes bindata generation deterministic, that is, if the same data is in `master.yaml` and `worker.yaml`, the generated `bindata.go` file will not change.

We use this property to make sure the `bindata.go` that's submitted in the PR matches what's in `master.yaml` and `worker.yaml`.

Makefile
iaguis: We do this only to make `bindata.go` deterministic, it doesn't matter which timestamp we use, but it has to be the same always.

circle.yml
xh3b4sd: How does the workflow look like then? Do I have to manually regenerate the bindata every time I change the templates? This would make me super sad. I want this to be automated.

iaguis: Yes.

I can't think of a way to automate this. The only way I can think of is having a pre-commit hook that does that, but every developer needs to make sure the hook is there in their local repo clones.

Maybe somebody else has ideas how to do this better?

nhlfr: I can't provide any better proposal either. What @iaguis proposes here seems to be the only doable way IMO.

nhlfr: What's more, if you are modifying the code of Kubernetes (or their related projects), you need to generate protobufs - for the same reason, they should be vendorable. They didn't figure any better way as well, you need to generate everything each time.

PR #21, opened 2017-03-16
create etcd data dir before chowning it

asymmetric:
Not sure who/what was creating that directory in other setups.


asymmetric:
Builds fail, cfr #22

PR #28, opened 2017-03-22
encrypt: add Worker fields in CompactTLSAssets struct

iaguis:
We need this to store the worker certs.


iaguis:
This is done in https://github.com/giantswarm/k8scloudconfig/pull/35 https://github.com/giantswarm/k8scloudconfig/pull/36 and  https://github.com/giantswarm/k8scloudconfig/pull/37

PR #31, opened 2017-03-23
templates: Run calico-node service after decrypt-tls-assets

nhlfr:
Calico needs decrypted certificates to run successfully.

### DIFF
--- a/templates.go
+++ b/templates.go
@@ -586,8 +586,8 @@ coreos:
     content: |
       [Unit]
       Description=Calico per-host agent
-      Requires=etcd2.service
-      After=etcd2.service
+      Requires=etcd2.service decrypt-tls-assets.service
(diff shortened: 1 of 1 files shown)


nhlfr:
Good point.

We have two options regarding Calico:
- remove it from k8scloudconfig and make separate Calico definitions in aws- and kvm-operator
- make Calico self-hosted

I'd prefer the second option. Self-hosted Calico is fine for us, because even if it will fail, k8s will restart it, so we don't have to make any logic around waiting for keys.

JosephSalisbury:
+1 for self-hosted calico

nhlfr:
We agreed to:
- improve ugly bash magic for waiting for certs to make it working before kubecon
- do self-hosted calico after kubecon

PR #32, opened 2017-03-23
templates: Include port in the etcd URL

nhlfr:


iaguis:
When we have ingress controllers set up we will be using port 443, that's why the port wasn't specified (implicit 443 with https).

I don't think we should do this here.

PR #36, opened 2017-03-24
Add worker assets to template

asymmetric:
Make the TLS assets available to the workers.

### DIFF
--- a/encrypt.go
+++ b/encrypt.go
@@ -4,7 +4,7 @@ type CompactTLSAssets struct {
 	APIServerCACrt    string
 	APIServerKey      string
 	APIServerCrt      string
-	WorkerCACrt       string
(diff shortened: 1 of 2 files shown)


encrypt.go
```diff
@@ -4,6 +4,9 @@ type CompactTLSAssets struct {
 	APIServerCACrt    string
 	APIServerKey      string
 	APIServerCrt      string
+	WorkerCACrt       string
```
xh3b4sd: I am not sure where the wording `CACrt` comes from but we wanted to have the scheme `ca`, `crt`, `key`. So it should be `WorkerCA` IMO.

asymmetric: True. I'll fix the naming for the other keys in a separate PR.

PR #39, opened 2017-03-31
templates: Add Calico IP pool on workers

nhlfr:
IP pool has to be added and defined on the every node.


nhlfr:
@xh3b4sd It's necessary, because nodes using Calico need to be aware of the IP address pool they're going to use for k8s pods. I don't have any idea how it worked in kvm-operator without that (or whether it worked :trollface: ).

PR #51, opened 2017-04-21
Moves etcd to only serve on localhost

JosephSalisbury:
Towards https://github.com/giantswarm/aws-operator/issues/133

We don't currently need to serve etcd on a domain,
so this changeset moves etcd to only serve on localhost.

### DIFF
--- a/templates.go
+++ b/templates.go
@@ -531,13 +531,13 @@ coreos:
       ExecStartPre=/bin/bash -c "while [ ! -f /etc/kubernetes/ssl/etcd/server-ca.pem ]; do echo 'Waiting for /etc/kubernetes/ssl/etcd/server-ca.pem to be written' && sleep 1; done"
       ExecStartPre=/bin/bash -c "while [ ! -f /etc/kubernetes/ssl/etcd/server-crt.pem ]; do echo 'Waiting for /etc/kubernetes/ssl/etcd/server-crt.pem to be written' && sleep 1; done"
       ExecStartPre=/bin/bash -c "while [ ! -f /etc/kubernetes/ssl/etcd/server-key.pem ]; do echo 'Waiting for /etc/kubernetes/ssl/etcd/server-key.pem to be written' && sleep 1; done"
-      ExecStart=/usr/bin/etcd2 --advertise-client-urls=https://{{.Cluster.Etcd.Domain}}:2379,http://127.0.0.1:2383 \
+      ExecStart=/usr/bin/etcd2 --advertise-client-urls=https://127.0.0.1:2379,https://127.0.0.1:2380 \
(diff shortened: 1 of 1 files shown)


templates.go
```diff
@@ -531,13 +531,13 @@ coreos:
       ExecStartPre=/bin/bash -c "while [ ! -f /etc/kubernetes/ssl/etcd/server-ca.pem ]; do echo 'Waiting for /etc/kubernetes/ssl/etcd/server-ca.pem to be written' && sleep 1; done"
       ExecStartPre=/bin/bash -c "while [ ! -f /etc/kubernetes/ssl/etcd/server-crt.pem ]; do echo 'Waiting for /etc/kubernetes/ssl/etcd/server-crt.pem to be written' && sleep 1; done"
       ExecStartPre=/bin/bash -c "while [ ! -f /etc/kubernetes/ssl/etcd/server-key.pem ]; do echo 'Waiting for /etc/kubernetes/ssl/etcd/server-key.pem to be written' && sleep 1; done"
-      ExecStart=/usr/bin/etcd2 --advertise-client-urls=https://{{.Cluster.Etcd.Domain}}:2379,http://127.0.0.1:2383 \
+      ExecStart=/usr/bin/etcd2 --advertise-client-urls=https://127.0.0.1:2379,https://127.0.0.1:2380 \
```
JosephSalisbury: I might be wrong here, but I think that:
`http://127.0.0.1:2383` is incorrect, and it should be `https://127.0.0.1:2380`

So, we use `2380` instead of `2383`, and also https to cover https://github.com/giantswarm/aws-operator/issues/132

rossf7: That makes sense to me and https://github.com/giantswarm/k8scloudconfig/pull/48 is related.

xh3b4sd: Here I come with my half knowledge. Isn't etcd required to listen on the domain? How does ingress to etcd via domain work when the server is not listening on the domain? Somebody will request the domain and the certificates are valid for the domain, when the server does not provide the domain because it is not listening on it, how does TLS then work?

JosephSalisbury: We're not currently exposing etcd outside.

xh3b4sd: Why did we do this before?

JosephSalisbury: Some weird kvm operator legacy that I can't remember the details of? :D 

Currently, all the master components are on one vm, so we don't need to expose it, we can just use localhost. Makes things a bit simpler.

xh3b4sd: Ok but lets keep the etcd certificates because when we rip out etcd to its own cluster we need this anyway for e.g. the apiserver etc.

JosephSalisbury: Yeah yeah, I reckon we'll need to do it in the future, just dropping it for now.

PR #54, opened 2017-04-21
properly install systemd unit on multi user target

xh3b4sd:
Fixes https://github.com/giantswarm/aws-operator/issues/131.

### DIFF
--- a/templates.go
+++ b/templates.go
@@ -1384,6 +1384,9 @@ coreos:
 
       [Service]
       ExecStart=/opt/bin/decrypt-tls-assets
+
+      [Install]
+      WantedBy=multi-user.target
   {{range .Units}}- name: {{.Metadata.Name}}
     enable: {{.Metadata.Enable}}
     command: {{.Metadata.Command}}


templates.go
```diff
@@ -1384,6 +1384,9 @@ coreos:
 
       [Service]
       ExecStart=/opt/bin/decrypt-tls-assets
+
+      [Install]
+      WantedBy=multi-user.target
```
xh3b4sd: See https://www.freedesktop.org/software/systemd/man/systemd.unit.html#%5BInstall%5D%20Section%20Options.

JosephSalisbury:
huh

we also have https://github.com/giantswarm/aws-operator/blob/5bb51ace983957697d6ad8afc85cc7b479d040ae/service/create/templates.go

which one of these is source of truth?

JosephSalisbury:
Having both kind of worries me, as two people have worked on two different ones :D

xh3b4sd:
Well, I am pretty lost on this organization of code now. Don't know. We should get this and the concept/idea behind it straight with @nhlfr and @asymmetric.

JosephSalisbury:
see https://github.com/giantswarm/aws-operator/issues/143 for tracking

xh3b4sd:
Having a closer look at the original issue I do not think this will fix it because the install target is defined as you just saw on the CoreOS.

nhlfr:
This service was removed by the another PR, closing.

PR #57, opened 2017-04-24
templates: Configure SSL connection to etcd in Calico

### LINKED ISSUE #56: Calico cannot connect to etcd with ssl
Towards https://github.com/giantswarm/giantswarm/issues/1388

```
Apr 24 11:05:18 ip-10-0-139-24.eu-west-1.compute.internal calicoctl[29512]: get_ip_pools: Error accessing etcd (Server response was not valid JSON:
Apr 24 11:05:18 ip-10-0-139-24.eu-west-1.compute.internal calicoctl[29512]: ValueError('No JSON object could be decoded',)).  Is etcd running?
```

nhlfr:
Fixes #56

### DIFF
--- a/templates.go
+++ b/templates.go
@@ -588,7 +588,11 @@ coreos:
       RestartSec=0
       TimeoutStopSec=10
       EnvironmentFile=/etc/network-environment
-      Environment="ETCD_AUTHORITY=127.0.0.1:2380"
+      Environment="ETCD_AUTHORITY=127.0.0.1:2379"
+      Environment="ETCD_SCHEME=https"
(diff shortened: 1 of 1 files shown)


templates.go
```diff
@@ -589,6 +589,10 @@ coreos:
       TimeoutStopSec=10
       EnvironmentFile=/etc/network-environment
       Environment="ETCD_AUTHORITY=127.0.0.1:2380"
```
JosephSalisbury: Do we want to use `2379` here?

nhlfr: yes, we want :)

PR #60, opened 2017-04-25
Moves extension units to top of cloudconfig

JosephSalisbury:
Towards https://github.com/giantswarm/aws-operator/issues/131

coreos-cloudinit starts unit sequentially, so the order of the
units needs to avoid deadlocks. In the case where the
aws-operator adds decrypt-tls-assets which is a requirement of
the later units, this causes a deadlock.

This changset moves the extension units to the top of the
cloudconfig, to avoid these deadlocks.

### DIFF
--- a/templates.go
+++ b/templates.go
@@ -455,6 +455,12 @@ write_files:
 
 coreos:
   units:
+  {{range .Units}}- name: {{.Metadata.Name}}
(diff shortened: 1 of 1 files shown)


JosephSalisbury:
This should fix it.

PR #68, opened 2017-04-28
templates: Don't override /etc/resolv.conf

nhlfr:
By default, the DHCP server in AWS points instances to their internal DNS servers, which makes our Route 53 domains available faster.

Fixes giantswarm/aws-operator#178


nhlfr:
@asymmetric done

PR #73, opened 2017-05-02
templates: Add wait-for-domains service

### LINKED ISSUE #72: Wait for DNS records to be available
Creating of Reoute 53 records takes some time, due to ELB health checks. We need to wait for the domains availability before setting up services.

nhlfr:
Fixes #72

### DIFF
--- a/templates.go
+++ b/templates.go
@@ -318,6 +318,19 @@ write_files:
         nodePort: 30011
       selector:
         app: ingress-controller
+- path: /opt/wait-for-domains
+  permissions: 0544
+  content: |
+      #!/bin/bash
+      until nslookup {{.Cluster.Etcd.Domain}}; do
+          echo "Waiting for domain {{.Cluster.Etcd.Domain}} to be available"
+          sleep 5
+      done
+
+      until nslookup {{ .Cluster.Kubernetes.API.Domain}}; do
+          echo "Waiting for domain {{.Cluster.Kubernetes.API.Domain}} to be available"
+          sleep 5
+      done
 - path: /opt/k8s-addons
   permissions: 0544
   content: |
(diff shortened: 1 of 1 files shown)


templates.go
```diff
@@ -455,6 +468,19 @@ coreos:
     content: |
       {{range .Content}}{{.}}
       {{end}}{{end}}
+  - name: wait-for-domains.service
+    enable; true
```
xh3b4sd: `;` -> `:`. Same issue in the worker.

rossf7:
The problem with removing the existing checks is that nslookup will just check if the record is there. But I think the ELB will still be unhealthy due to the healthcheck.

screen shot 2017-05-02 at 12 40 53 pm

If we want to check if the etcd service is ready we need to decouple it from apiserver. As it needs to come up first.

nhlfr:
@xh3b4sd I don't think so. The other checks which we have in the cloudconfig are more like `curl --cacert something.crt https://{{.Cluster.Etcd.Domain}}/version`, so they are checking explicitly checking whether etcd is working, rather than only domains.

The service I'm introducing here will check domains before running etcd.

nhlfr:
@rossf7 At this point, I don't want to check whether etcd is ready. I want to check the domain availability before. Checking etcd availability is already done in the other places, but the overall process doesn't work properly, because etcd starts when domain isn't available yet.

templates.go
```diff
@@ -318,6 +318,19 @@ write_files:
         nodePort: 30011
       selector:
         app: ingress-controller
+- path: /opt/wait-for-domains
+  permissions: 0544
+  content: |
+      #!/bin/bash
+      until nslookup {{.Cluster.Etcd.Domain}}; do
+          echo "Waiting for domain {{.Cluster.Etcd.Domain}} to be available"
+      sleep 5
```
asymmetric: You got an indentation problem here :)

Issue #69, opened 2017-04-30
calico-node unit fails to start due to invalid IP


rossf7:
Towards giantswarm/giantswarm#1388

Unit fails to start due to the error below. It looks like its because the env var BRIDGE_IP is not set. This is need to add a bgp peer using calicoctl.

```
Apr 30 15:50:11 master-1 bash[2332]: Invalid IP address specified.
Apr 30 15:50:11 master-1 systemd[1]: calico-node.service: Control process exited, code=exited status=1
```

See https://github.com/giantswarm/k8scloudconfig/blob/master/templates.go#L609

calvix:
`BRIDGE_IP` is on our barematel setup IP of flannel bridge.
In AWS setup I think it should have value of ip address which is used for communication between workers and master. So one of the internal ip.

rossf7:
Using the DEFAULT_IPV4 env var instead of BRIDGE_IP works for adding the bgp peer. Although the cloudconfig needs to work for aws and bare metal.

There is also a permissions issue. The unit needs to save the key /calico/v1/host/{{.Node.Hostname}}-flannel/bird_ip to etcd but this fails with an invalid cert error. However calling etcdctl manually with the cert args works so the cert is valid. We set the owner of the certs files to etcd so that might be the problem.

Lets discuss tomorrow when everyone is back.

rossf7:
BRIDGE_IP issue was fixed by creating a oneshot unit that creates /etc/calico-environment and sets the bridge IP to the internal IP.

PR #75, opened 2017-05-02
Provide bridge IP to calico node unit

rossf7:
Towards giantswarm/giantswarm#1388 and #69

This PR adds an env file in /etc/calico-environment that sets the bridge IP. On AWS this will use the internal IP for the instance. On KVM it needs to be the bridge IP.

Also adds missing cert files when saving the IP to etcd in the master template.


templates.go
```diff
@@ -453,7 +453,8 @@ write_files:
   encoding: gzip+base64
   content: {{.TLSAssets.EtcdServerKey}}
 
-{{range .Files}}- path: {{.Metadata.Path}}
+{{range .Files}}
+- path: {{.Metadata.Path}}
```
rossf7: This newline is needed or the 2nd file has incorrect indentation in the cloudconfig. This means that the file contents is wrong.

templates.go
```diff
@@ -1087,7 +1089,8 @@ write_files:
...
```
rossf7: Same issue with the worker cloudconfig.

PR #76, opened 2017-05-03
Evaluate calico env file in pre command

rossf7:
Towards giantswarm/giantswarm#1388 and #69 

Environment files are not evaluated in systemd units but on AWS we need to set BRIDGE_IP to DEFAULT_IPV4. So this is done in a pre command instead.

### DIFF
--- a/templates.go
+++ b/templates.go
@@ -614,7 +614,6 @@ coreos:
       RestartSec=0
       TimeoutStopSec=10
       EnvironmentFile=/etc/network-environment
-      EnvironmentFile=/etc/calico-environment
       Environment="ETCD_AUTHORITY={{ .Cluster.Etcd.Domain }}:2379"
       Environment="ETCD_SCHEME=https"
       Environment="ETCD_CA_CERT_FILE=/etc/kubernetes/ssl/calico/client-ca.pem"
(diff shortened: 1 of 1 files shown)


templates.go
```diff
@@ -626,6 +625,7 @@ coreos:
       ExecStartPre=/bin/bash -c "while [ ! -f /etc/kubernetes/ssl/calico/client-crt.pem ]; do echo 'Waiting for /etc/kubernetes/ssl/calico/client-crt.pem to be written' && sleep 1; done"
       ExecStartPre=/bin/bash -c "while [ ! -f /etc/kubernetes/ssl/calico/client-key.pem ]; do echo 'Waiting for /etc/kubernetes/ssl/calico/client-key.pem to be written' && sleep 1; done"
       ExecStartPre=/bin/bash -c "while ! curl --output /dev/null --silent --fail --cacert /etc/kubernetes/ssl/calico/client-ca.pem --cert /etc/kubernetes/ssl/calico/client-crt.pem --key /etc/kubernetes/ssl/calico/client-key.pem https://{{ .Cluster.Etcd.Domain }}:2379/version; do sleep 1 && echo 'Waiting for etcd master to be responsive'; done"
+      ExecStartPre=/bin/bash -c "source /etc/calico-environment"
```
xh3b4sd: That does not make sense to me. I do not understand it. When we need this we also would need it for `/etc/network-environment`, which we don't. So I presume the root cause is something else.

rossf7: Network environment only has values so there is no evaluation issue.

```
cat /etc/network-environment
LO_IPV4=127.0.0.1
ETH0_IPV4=10.0.134.68
DEFAULT_IPV4=10.0.134.68
DOCKER0_IPV4=172.17.0.1
```

In the calico env file I need to set BRIDGE_IP to DEFAULT_IPV4 but the file isn't evaluated so it doesn't work.

```
cat /etc/calico-environment
# On AWS use internal IP as bridge IP
BRIDGE_IP=$DEFAULT_IPV4
```

It may be we need a better way to set this env var.

nhlfr: Unfortunately, I don't think there is any better way for evaluating. Sad but true, systemd doesn't have any functionality for that.

rossf7:
Closing in favour of giantswarm/aws-operator/pull/187

PR #79, opened 2017-05-04
Service account

### LINKED ISSUE #70: apiserver unit fails to start due to missing service account key
Towards giantswarm/giantswarm#1388

Unit fails to start because there is no key file at /etc/kubernetes/secrets/token_sign_key.pem

```
Apr 30 15:31:53 master-1 docker[3449]: I0430 15:31:53.171367       1 config.go:498] Will report 10.0.135.53 as public IP address.
Apr 30 15:31:53 master-1 docker[3449]: F0430 15:31:53.171721       1 server.go:244] Invalid Authentication Config: read /etc/kubernetes/secrets/token_sign_key.pem: is a directory
Apr 30 15:31:53 master-1 systemd[1]: k8s-api-server.service: Main process exited, code=exited, status=255/n/a
```

See https://github.com/giantswarm/k8scloudconfig/blob/master/templates.go#L705

rossf7:
Towards giantswarm/giantswarm#1388 and fixes #70 

This PR adds service account to the TLS assets and creates the encoded files in the master cloudconfig.

### DIFF
--- a/encrypt.go
+++ b/encrypt.go
@@ -1,16 +1,19 @@
 package cloudconfig
 
 type CompactTLSAssets struct {
-	APIServerCA     string
-	APIServerKey    string
-	APIServerCrt    string
-	WorkerCA        string
-	WorkerKey       string
-	WorkerCrt       string
-	CalicoClientCA  string
-	CalicoClientKey string
-	CalicoClientCrt string
-	EtcdServerCA    string
-	EtcdServerKey   string
-	EtcdServerCrt   string
+	APIServerCA       string
+	APIServerKey      string
+	APIServerCrt      string
+	WorkerCA          string
+	WorkerKey         string
+	WorkerCrt         string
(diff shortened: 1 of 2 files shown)


encrypt.go
```diff
@@ -1,16 +1,19 @@
...
-	EtcdServerCA    string
-	EtcdServerKey   string
-	EtcdServerCrt   string
+	APIServerCA       string
+	APIServerKey      string
+	APIServerCrt      string
+	WorkerCA          string
+	WorkerKey         string
+	WorkerCrt         string
+	ServiceAccountCA  string
```
rossf7: Even though we're just using the keypair I'm adding the CA and cert so they are downloaded and decrypted. That feels correct but not 100% sure. WDYT?

xh3b4sd: Having it for completeness sounds good to me.

PR #92, opened 2017-05-08
fixing missing templating changes

puja108:
we had forgotten the 4 IC API calls


puja108:
I also did a search for more `%%` but couldn't find any, so I hope we're good 😉

Issue #80, opened 2017-05-04
move certificate file to corresponding operators


xh3b4sd:
Right now we have certificate file management in here which is necessary for AWS. The aws-operator leverages KMS to encrypt/decrypt certificate data because we are storing the cloud config on S3. This is not necessary for KVM. So the aws-operator should define the list of files on its own as should the kvm-operator. For the kvm-operator the certificate file management is a bit more easy because we do not need to store the cloud config anywhere, but can directly apply the secret data as received from Kubernetes. We should do this rather sooner than later. I am already starting on improving the kvm-operator. 

See https://github.com/giantswarm/k8scloudconfig/blob/250143cb667b43239154618419bc35b6361e3053/templates.go#L420-L466. 

FYI @rossf7 @nhlfr

PR #95, opened 2017-05-10
Move TLS assets from k8scloudconfig to cloud config plugin

nhlfr:
Ref giantswarm/k8scloudconfig#80

### DIFF
--- a/cloudconfig.go
+++ b/cloudconfig.go
@@ -11,11 +11,11 @@ import (
 )
 
 type CloudConfigTemplateParams struct {
-	Cluster   clustertpr.Cluster
-	Node      node.Node
-	TLSAssets CompactTLSAssets
-	Files     []FileAsset
-	Units     []UnitAsset
+	Cluster     clustertpr.Cluster
+	Node        node.Node
+	MasterFiles []FileAsset
+	WorkerFiles []FileAsset
+	Units       []UnitAsset
 }
 
 type CloudConfig struct {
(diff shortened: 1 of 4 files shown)


nhlfr:
I made a distinction between MasterFiles and WorkerFiles there.

MasterFiles in operators is going to contain:
- apiserver certs
- service account certs
- calico certs
- etcd server certs

WorkerFiles:
- worker certs
- calico certs
- etcd certs

cloudconfig.go
```diff
@@ -11,11 +11,11 @@ import (
...
 type CloudConfigTemplateParams struct {
-	Cluster   clustertpr.Cluster
-	Node      node.Node
-	TLSAssets CompactTLSAssets
-	Files     []FileAsset
-	Units     []UnitAsset
+	Cluster     clustertpr.Cluster
+	Node        node.Node
+	MasterFiles []FileAsset
+	WorkerFiles []FileAsset
```
xh3b4sd: Why are we wanting to separate the files? As I see it the input is quite the same concept wise, no matter if it is executed against the master or the worker template. If there are differences, that I do not see, then the separation should be made towards the whole input. So there would be something like  `MasterCloudConfigTemplateParams` and `WorkerCloudConfigTemplateParams`. That would make sense in case we need to separate it, but this change made here feels like we are getting the abstraction wrong.

nhlfr: Good idea, will re-implement that.

PR #98, opened 2017-05-11
Remove TLSAssets from cloudconfig

nhlfr:
TLS files need to be injected by operators (aws/kvm) as regular files.

Ref giantswarm/k8scloudconfig#80


nhlfr:
@xh3b4sd Towards https://github.com/giantswarm/k8scloudconfig/pull/95#discussion_r115987255

I realized that separating master and worker files isn't needed at all, since those are separate templates. So we can just make a distinction between parameters passed to them in aws-operator. In this PR, I'm only removing TLS files. Everything else will be implemented in aws-operator.

nhlfr:
I'm also thinking where we should keep this structure https://github.com/giantswarm/k8scloudconfig/blob/master/encrypt.go

It will be reused both in aws- and kvm-operator, so keeping it in operators doesn't seem to be a good idea. On the other hand, it isn't a part of cloudconfig, so this repo also doesn't seem to be a right place. Any suggestions? Maybe we need an another library for keeping commons for operators?

xh3b4sd:
> I realized that separating master and worker files isn't needed at all, since those are separate templates. So we can just make a distinction between parameters passed to them in aws-operator. In this PR, I'm only removing TLS files. Everything else will be implemented in aws-operator.

Exactly. That was my thinking. 

> I'm also thinking where we should keep this structure https://github.com/giantswarm/k8scloudconfig/blob/master/encrypt.go

Sounds like certifictetpr to me. It is about the asset management. There are already types and structures in there that also use the same naming scheme.

nhlfr:
Cool, I will remove CompactTLSAssets and move them to certificatetpr then.

Issue #100, opened 2017-05-12
Secure CoreOS SSH config


kopiczko:
https://github.com/giantswarm/vodafone/issues/128

calvix:
this should be done right @kopiczko ?

if yes please close

PR #101, opened 2017-05-12
Harden SSH config (#100)

kopiczko:
#100
towards giantswarm/giantswarm#1419 and in detail giantswarm/vodafone#128

### DIFF
--- a/templates.go
+++ b/templates.go
@@ -420,6 +420,7 @@ write_files:
 - path: /etc/ssh/sshd_config
   owner: root
   permissions: 0600
+  content: |
     ClientAliveCountMax 2
     MaxAuthTries 3
     PermitRootLogin no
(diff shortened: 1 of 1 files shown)


kopiczko:
First I'd like to log in to one of the machines and see `/etc/ssh/sshd_config` content first -- just to make sure sth is not elided.
This *shouldn't* ssh connections other than root.

calvix:
cool we have some security experts :)I Like that

Just few notes:
1) please try to provide info to PR, it should include reason why you are changing stuff and also info what each change does (just some short summary)
2) as this is for `k8s-vm` we should completely disable password login for all users, as we only use ssh keys  and also that would make MaxAuthTries redundant

puja108:
re 1. I guess a following should be enough:
towards https://github.com/giantswarm/giantswarm/issues/1419 and in detail https://github.com/giantswarm/vodafone/issues/128

re 2. that does make sense, we can do more than they asked for 😄

kopiczko:
@calvix it's #100 - it was in the title, I updated also description. That also makes me less of an security expert :)

calvix:
@kopiczko thanks for the update :) we are lazy poeple and if there is no link  ver are sad,:)

kopiczko:
@calvix does setting `MaxAuthTries` make sense while `PasswordAuthentication no`?

calvix:
@kopiczko 
```
    MaxAuthTries
    Specifies the maximum number of authentication attempts permitted per connection. Once the number of failures reaches half this value, additional failures are logged. The default is 6.

```
from manual, 
So no it does not make sense.

kopiczko:
I just wasn't sure if non-password auth counts.

kopiczko:
@puja108 all good now?

PR #105, opened 2017-05-15
make calico-node dependent on the DNS records

asymmetric:
this way, we ensure that the whole kubernetes stack waits until the records are there


rossf7:
Towards #109

PR #106, opened 2017-05-15
remove domains requirement from etcd2

asymmetric:
etcd doesn't need the DNS records to be present. it's etcd's clients who need it, like calico-node.


rossf7:
Towards #109

PR #108, opened 2017-05-15
extensions: Add "encoding" field to file metadata

nhlfr:
TLS assets need to use gzip+base64 encoding, so we need to allow to specify encoding by extensions.

Ref #80

### DIFF
--- a/operator.go
+++ b/operator.go
@@ -6,10 +6,26 @@ import (
 	"text/template"
 )
 
+type encoding string
+
+const (
+	UTF8       encoding = "utf-8"
+	GzipBase64 encoding = "gzip+base64"
+)
+
+func (e encoding) String() string {
+	if e == "" {
+		return string(UTF8)
+	}
+
+	return string(e)
+}
+
 type FileMetadata struct {
 	AssetContent string
 	Path         string
 	Owner        string
+	Encoding     encoding
 	Permissions  int
 }
 
--- a/operator_test.go
+++ b/operator_test.go
@@ -10,6 +10,26 @@ type FakeParams struct {
 	Foo string
 }
 
+func TestEncodingType(t *testing.T) {
+	tests := []struct {
+		encodingInstance encoding
+		expectedString   string
+	}{
+		{
+			encodingInstance: encoding(""),
+			expectedString:   "utf-8",
+		},
+		{
+			encodingInstance: encoding("gzip+base64"),
+			expectedString:   "gzip+base64",
+		},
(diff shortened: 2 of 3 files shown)


operator.go
```diff
@@ -6,10 +6,21 @@ import (
 	"text/template"
 )
 
+type encoding string
```
xh3b4sd: What do you think about this: https://github.com/giantswarm/architect/blob/master/configuration/provider/provider.go? 

Having constants with the specific private type allow decent godoc and easy use out of the box. IMO we do not need a String() method, we only need the allowed options protected by a package type. That also prevents calling some method with empty strings like `encoding("")`.

nhlfr: I don't get your idea, to be honest.

The goal I wanted to achieve by writing my own type, was to make `utf-8` a default value. That's because if we don't specify encoding for some file, like i.e.:

```
FileMetadata{
    Path: /opt/something,
    Owner: "root:root",
    Permissions: 0700,
    Content: "foobar",
}
```

I would like to render:

```
- path: /opt/something
  owner: "root:root"
  encoding: utf-8
  permissions: 0700
  content: |
    foobar
```

instead of:

```
- path: /opt/something
  owner: "root:root"
  encoding:
  permissions: 0700
  content: |
    foobar
```

which would surely result in an error. Of course I could use if statement in template to not define encoding at all if it's empty, but my solution with custom type providing default values is IMO much better.

And that's why we need String() method, because template library is using that method for types not being a `string` (even if you just define your own string type). I've tested that, it works fine and generates good cloudconfig.

The custom type you linked seems to be unrelated to this problem. Am I wrong?

xh3b4sd: I somehow can explain this in words apparently. Let me explain in code. I would like to have this. 

```
type encoding string

const (
	UTF8 encoding = "utf-8"
	GzipBase64 encoding = "gzip+base64"
)
```

- We do not need a string method. 
- We do not need to use empty strings to get some default value. 
- We use the private type to protect our choices. 
- We provide all available choices via constants. 

The constant has only to be applied to the template params, which you would also do with your current encoding type. I only think what I suggest is more strict and more straight forward. At the end, do what you think is right.

nhlfr: The point is that I don't use empty strings explicitly. I just want to have a `utf-8` value if I don't give any value to `Encoding` field at all.

But sure, consts are good, but I would like to keep String() method as well.

nhlfr:
@xh3b4sd PTALA

PR #112, opened 2017-05-17
updating and fixing nginx IC

puja108:
- update to version 0.9beta5
- adds configmap with hash bucket increase
- fixes ssl port


puja108:
this is implemented and tested manually on `l3l85` cluster

PR #114, opened 2017-05-17
removing bridge ip from calico on master node too

puja108:
to get master node calico working on aws


puja108:
testing this manually still had no results, we might need more

PR #117, opened 2017-05-19
enable cloud configs in KVM

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/1197. I did several things here. Thanks for bearing with me. The PR is not that big though. 

- Updated vendor deps to get clustertpr changes for the hostname replacement. 
- Simplifying some variable naming. This is nicer from a client point of view. 
- Added a `String()` method to the cloud config. KVM does not expect the cloud config to be base64 encoded.

### DIFF
--- a/templates.go
+++ b/templates.go
@@ -2,7 +2,6 @@ package cloudconfig
 
 const (
 	MasterTemplate = `#cloud-config
-hostname: "{{.Node.Hostname}}"
 write_files:
 - path: /srv/kubedns-dep.yaml
   owner: root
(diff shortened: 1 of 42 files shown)


templates.go
```diff
@@ -1359,7 +1357,7 @@ coreos:
       --register-node=true \
       --allow-privileged=true \
       --kubeconfig=/etc/kubernetes/config/kubelet-kubeconfig.yml \
-      --node-labels="kubernetes.io/hostname={{.Node.Hostname}},ip=${DEFAULT_IPV4},{{.Cluster.Kubernetes.Kubelet.Labels}}" \
+      --node-labels="kubernetes.io/hostname=${HOSTNAME},ip=${DEFAULT_IPV4},{{.Cluster.Kubernetes.Kubelet.Labels}}" \
```
rossf7: <3

PR #118, opened 2017-05-19
put extension to params to simplify interface

xh3b4sd:
This PR aims to ease the creation of cloud configs.


xh3b4sd:
Agreed. I do not plan to change more. The changes made today are not that big so this will not be an issue. Let me know when I should take it.

Issue #116, opened 2017-05-19
run kubelet on master (e.g. for self-hosted calico)


puja108:
We need to run a kubelet on master to be able to schedule things like Calico there.

Not sure if we also need to run kube-proxy. Might need to check my local vagrant setup (or directly https://github.com/coreos/coreos-kubernetes)

The kubelet on master should be tainted. From http://docs.projectcalico.org/v2.1/getting-started/kubernetes/installation/hosted/calico.yaml there is a typical toleration for master kubelets:

```
scheduler.alpha.kubernetes.io/tolerations: |
          [{"key": "dedicated", "value": "master", "effect": "NoSchedule" },
           {"key":"CriticalAddonsOnly", "operator":"Exists"}]
```

If our taint looks similar we should be good.

puja108:
This is how we do it in g8s (should be mostly transferrable):
kubelet:
https://github.com/giantswarm/hive/blob/init-bootstrap-node/template_snippets/cloudconfig/g8s_master.yaml#L36
kubeproxy:
https://github.com/giantswarm/hive/blob/init-bootstrap-node/template_snippets/cloudconfig/g8s_master.yaml#L289

puja108:
This should be done, just needs to be validated

PR #121, opened 2017-05-23
adding kubelet and proxy to master

puja108:
WIP

towards https://github.com/giantswarm/k8scloudconfig/issues/116

### DIFF
--- a/templates.go
+++ b/templates.go
@@ -399,6 +399,48 @@ write_files:
         "https://{{.Cluster.Kubernetes.API.Domain}}:443/api/v1/namespaces/kube-system/services"
 
       echo "Addons successfully installed"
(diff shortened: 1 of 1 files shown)


puja108:
Todos:

- [ ] add kubeconfig.yaml for kubelet to `/etc/kubernetes/config/`
- [ ] add kubeconfig.yaml for proxy to `/etc/kubernetes/config/`

puja108:
not sure we need all that waiting for calico and things in there, but for now I'd like to just get that kubelet working, and we can handle calico move afterwards.

RFR

rossf7:
Thanks @puja108 I'll launch a cluster with this cloudconfig to test. Agree on removing the calico waits later.

puja108:
first testing looks good, next would be launching a full cluster on aws with this

Issue #40, opened 2017-03-31
Make Calico self-hosted


nhlfr:
Make Calico self-hosted

http://docs.projectcalico.org/v2.0/getting-started/kubernetes/installation/hosted/

Currently we set up Calico with systemd units, without using containers, and with downloading binaries as a part of systemd unit, which really should be changed!

puja108:
+1

I'm fiddling around looking into upgrading to a recent version (2.1.1+ needed for k8s 1.6+) and it's quite a mindfuck like it is right now.

calvix:
We already running calico as k8s pod in g8s itself on aws so i would recommend copy anything necessary for that configuration !

https://github.com/giantswarm/gsegg/blob/initial_version/template-manifests/calico-node.yaml

puja108:
cool

puja108:
Looked further into it and etcdless might not be good for now, there's stuff missing like calico bgp itself and lot of other functionality, it only does policy enforcement. We should use http://docs.projectcalico.org/v2.1/getting-started/kubernetes/installation/hosted/hosted for now

puja108:
I can get this running already. Some issues I still have:

- There's one thing we need to check, currently the etcd TLS setup is through a secret, I manually fiddled our certs in there, but we need to automate that secret creation (the secret needs to be in `kube-system` inside guest k8s). 
- Currently, the DS only schedules to one node, not sure what is goign wrong with the discovery of the second node.
- The DS should additionally schedule to the master node, but as we currently do not have a kubelet running there it can't, so we would need to have at least a kubelet running, even if we don't move to self-hosted yet.

puja108:
I tried it out on a local 2 worker vagrant setup with k8s 1.5.5 and got it working, so in general it should work with our setup considering the above points (especially 1 and 3) solved

calvix:
@puja108 we run kubelet on core machines so that is fine, we jsut ahve to ahve different labels so on workers kubelet  an core kubelet

nhlfr:
Unfortunately, in k8scloudconfig we don't run kubelet on masters yet, we need to change that when implementing hosted installation.

puja108:
The master kubelet should be tainted, and the calico DS has a toleration for that taint.

puja108:
just as a sidenote: we could run a kubelet on masters, even without goign full self-hosted k8s

puja108:
needs https://github.com/giantswarm/k8scloudconfig/issues/116

[...]

PR #123, opened 2017-05-24
making calico hosted (and updating to latest)

puja108:
Towards https://github.com/giantswarm/k8scloudconfig/issues/40

WIP

Running Calico 2.2.1

This is already implemented and tested manually on `wopre` cluster on AWS FRA

TODO:
- [x] add calico yamls
- [x] check/rename etcd cert directory/filename
- [ ] test
- [ ] retag images

### DIFF
--- a/templates.go
+++ b/templates.go
@@ -310,7 +310,7 @@ write_files:
                   value: "true"
               volumeMounts:
                 # Mount in the etcd TLS secrets.
-                - mountPath: /etc/kubernetes/etcd
+                - mountPath: /etc/kubernetes/ssl/etcd
                   name: etcd-certs
           volumes:
             # Mount in the etcd TLS secrets.
--- a/templates.go
+++ b/templates.go
@@ -71,9 +71,9 @@ write_files:
             }
         }
     
-      etcd_ca: "/etc/kubernetes/etcd/etcd-ca.pem"
-      etcd_cert: "/etc/kubernetes/etcd/etcd.pem"
-      etcd_key: "/etc/kubernetes/etcd/etcd-key.pem"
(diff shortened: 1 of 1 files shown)


puja108:
Open question: Do we still need to set MTU manually?

calvix:
@puja108  for KVM on-prem yes we need

puja108:
@calvix so could we make this optional? I don't feel good adjusting MTU when not needed on AWS

templates.go
```diff
@@ -3,27 +3,320 @@ package cloudconfig
...
+      namespace: kube-system
+    data:
+      # Configure this with the location of your etcd cluster.
+      etcd_endpoints: "https://{{ .Cluster.Etcd.Domain }}:2379"
+
+      # Configure the Calico backend to use.
+      calico_backend: "bird"
+    
+      # The CNI network configuration to install on each node.
+      # TODO: Do we still need to set MTU manually?
```
rossf7: Not urgent but would be good to know about MTU. As we can also remove from clustertpr if we don't need it.

templates.go
```diff
@@ -3,27 +3,320 @@ package cloudconfig
...
+      selector:
+        matchLabels:
+          k8s-app: calico-node
+      template:
+        metadata:
+          labels:
+            k8s-app: calico-node
+          annotations:
+            scheduler.alpha.kubernetes.io/critical-pod: ''
+            scheduler.alpha.kubernetes.io/tolerations: |
```
rossf7: This is really nice.

templates.go
puja108: @calvix mentioned we need it in KVM, so I created a ticket in operator board for making it optional

puja108: https://github.com/giantswarm/k8scloudconfig/issues/126

PR #124, opened 2017-05-25
update etcd version

calvix:
Upgrading etcd to etcdv3 and moving etcd to container.
Its working, but it cannot be fully tested because current calico does not support etcdv3 (etcdv3 forces TLS 1.2 and old calico is trying TLS 1.0 ), so we have to wait for https://github.com/giantswarm/k8scloudconfig/pull/123

### DIFF
--- a/templates.go
+++ b/templates.go
@@ -656,33 +656,6 @@ coreos:
 
       [Install]
       WantedBy=multi-user.target
-  - name: etcd3-restart.service
-    enable: true
-    content: |
-      [Unit]
-      Description=etcd3-restart
-
-      [Service]
-      Type=oneshot
-      ExecStartPre=/usr/bin/systemctl stop etcd3.service
-      ExecStartPre=/usr/bin/bash -c 'while systemctl is-active --quiet etcd3.service; do sleep 1 && echo waiting for etcd3 to stop; done'
-      ExecStart=/usr/bin/systemctl start etcd3.service
-
-      [Install]
-      WantedBy=multi-user.target
-  - name: etcd3-restart.timer
-    enable: true
-    command: start
-    content: |
-      [Unit]
-      Description=Timer
-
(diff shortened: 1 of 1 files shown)


templates.go
```diff
@@ -600,62 +600,77 @@ coreos:
...
+          --listen-peer-urls=https://${DEFAULT_IPV4}:2380 \
+          --initial-cluster-token k8s-etcd-cluster \
+          --initial-cluster etcd0=https://127.0.0.1:2380 \
+          --initial-cluster-state new \
+          --data-dir=/var/lib/etcd
 
       [Install]
       WantedBy=multi-user.target
-  - name: etcd2-restart.service
+  - name: etcd3-restart.service
```
JosephSalisbury: For my interest, why do we need this?

calvix: good question, i think it was there from times when we had cert units inside vm, and the restart renewed certs

calvix: i thnik we can remove that atm @JosephSalisbury @rossf7

JosephSalisbury: I'm a bit worried that it means that all etcd v3 nodes will restart at the same time. I would remove them.

JosephSalisbury: There's a cert renewal story, but I don't see it yet.

rossf7: I didn't query it because several other units have a restart service. But I don't know why its there so I should have done :(

I'm fine to remove it if its not needed. For restarts currently masters are single node. But I'd like to see etcd move to the host cluster, which I think is the plan?

JosephSalisbury: Provisionally yes, but that won't happen v. shortly, so let's make sure this is good.

calvix: on it

calvix: btw we had this for quite some time, :)

JosephSalisbury: oof

PR #125, opened 2017-05-25
remove restart units

calvix:
I dont see any reason why we should have restart units for:
 * k8s proxy
* kubelet
* k8s api server
* k8s controller manager
* k8s scheduler
(etcd is removed in separate PR - https://github.com/giantswarm/k8scloudconfig/pull/124)

WDYT @JosephSalisbury @rossf7 @xh3b4sd

### DIFF
--- a/templates.go
+++ b/templates.go
@@ -704,33 +704,6 @@ coreos:
       --v=2"
       ExecStop=-/usr/bin/docker stop -t 10 $NAME
       ExecStopPost=-/usr/bin/docker rm -f $NAME
-  - name: k8s-proxy-restart.service
-    enable: true
-    content: |
-      [Unit]
-      Description=k8s-proxy-restart
-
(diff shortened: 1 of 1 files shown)


JosephSalisbury:
For my interest, what is the story for rotating certs in `k8scloudconfig`?

calvix:
we dont  have one atm i think, which is not good, but even if we had one,  random restart is not the solution

calvix:
WDYT @puja108 @xh3b4sd

JosephSalisbury:
I have no issue removing the units till we sort out the rotation story - we can totally add them back later if we need them.

puja108:
Yeah, I was also thinking why we need those regular restarts, that doesn't feel very sophisticated, I thought it might be rotating keys, but it's not related, so 🔥

PR #134, opened 2017-05-29
IC update and sync with upstream

puja108:
based on:
https://github.com/kubernetes/ingress/blob/master/examples/deployment/nginx/nginx-ingress-controller.yaml
https://github.com/kubernetes/ingress/blob/master/examples/deployment/nginx/default-backend.yaml

- update to beta.7
- moved back service to port 443
- use default-backend from upstream
- change naming and healthchecks to be similar to upstream

this is rolled out and tested (with SSL passthrough) on `z5eru`


puja108:
Todo:
- [ ] change APi calls in addons

puja108:
RFR

PR #135, opened 2017-05-29
k8s addons v2

calvix:
redesign of k8s-addons script

* moving to `kubectl`
* doing retries when `apply` fails
* moving files to array so we don't have to repeat same shit hundred times
ping @giantswarm/team-positive

### DIFF
--- a/templates.go
+++ b/templates.go
@@ -670,15 +670,14 @@ write_files:
 
       for manifest in $MANIFESTS
       do
-      		while
-         		/opt/bin/kubectl --kubeconfig=/etc/kubernetes/config/kubelet-kubeconfig.yml apply -f /srv/$manifest
-         		[ "$?" -ne "0" ]
-        	do
-                	echo "failed to apply /src/$manifest, retrying in 5 sec"
-                	sleep 5s
-        	done
+          while
+              /opt/bin/kubectl --kubeconfig=/etc/kubernetes/config/kubelet-kubeconfig.yml apply -f /srv/$manifest
+              [ "$?" -ne "0" ]
+          do
+              echo "failed to apply /src/$manifest, retrying in 5 sec"
+              sleep 5s
+          done
       done
-
       echo "Addons successfully installed"
 - path: /etc/kubernetes/config/proxy-kubeconfig.yml
   owner: root
--- a/templates.go
+++ b/templates.go
@@ -658,70 +658,26 @@ write_files:
   permissions: 0544
   content: |
       #!/bin/bash
-      while ! curl --output /dev/null --silent --head --fail --cacert /etc/kubernetes/ssl/apiserver-ca.pem --cert /etc/kubernetes/ssl/apiserver-crt.pem --key /etc/kubernetes/ssl/apiserver-key.pem "https://{{.Cluster.Kubernetes.API.Domain}}:443"; do sleep 1 && echo 'Waiting for master'; done
-      curl -o /opt/bin/kubectl https://storage.googleapis.com/kubernetes-release/release/v1.6.1/bin/linux/amd64/kubectl 
+      # get kubectl
+      curl -o /opt/bin/kubectl https://storage.googleapis.com/kubernetes-release/release/v1.6.1/bin/linux/amd64/kubectl
       chmod +x /opt/bin/kubectl
+
+      # wait for healthy master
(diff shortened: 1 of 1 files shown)


templates.go
```diff
@@ -658,70 +658,26 @@ write_files:
   permissions: 0544
   content: |
       #!/bin/bash
-      while ! curl --output /dev/null --silent --head --fail --cacert /etc/kubernetes/ssl/apiserver-ca.pem --cert /etc/kubernetes/ssl/apiserver-crt.pem --key /etc/kubernetes/ssl/apiserver-key.pem "https://{{.Cluster.Kubernetes.API.Domain}}:443"; do sleep 1 && echo 'Waiting for master'; done
-      curl -o /opt/bin/kubectl https://storage.googleapis.com/kubernetes-release/release/v1.6.1/bin/linux/amd64/kubectl 
+      # get kubectl
+      curl -o /opt/bin/kubectl https://storage.googleapis.com/kubernetes-release/release/v1.6.1/bin/linux/amd64/kubectl
```
rossf7: Nitpick but any reason we don't use 1.6.4 here?

templates.go
```diff
@@ -658,70 +658,26 @@ write_files:
...
+      for manifest in $MANIFESTS
+      do
+      		while
+         		/opt/bin/kubectl --kubeconfig=/etc/kubernetes/config/kubelet-kubeconfig.yml apply -f /srv/$manifest
+         		[ "$?" -ne "0" ]
+        	do
+                	echo "failed to apply /src/$manifest, retrying in 5 sec"
+                	sleep 5s
+        	done
+      done
```
kopiczko: Spaces when nuts here

templates.go
calvix: good idea, can we somehow replace by some variable we have for this?

templates.go
```diff
@@ -658,70 +658,26 @@ write_files:
...
-      curl -H "Content-Type: application/yaml" \
-        -XPOST -d"$(cat /srv/default-backend-dep.yml)" \
-        --cacert /etc/kubernetes/ssl/apiserver-ca.pem --cert /etc/kubernetes/ssl/apiserver-crt.pem --key /etc/kubernetes/ssl/apiserver-key.pem \
-        "https://{{.Cluster.Kubernetes.API.Domain}}:443/apis/extensions/v1beta1/namespaces/kube-system/deployments"
-      curl -H "Content-Type: application/yaml" \
-        -XPOST -d"$(cat /srv/default-backend-svc.yml)" \
-        --cacert /etc/kubernetes/ssl/apiserver-ca.pem --cert /etc/kubernetes/ssl/apiserver-crt.pem --key /etc/kubernetes/ssl/apiserver-key.pem \
-        "https://{{.Cluster.Kubernetes.API.Domain}}:443/api/v1/namespaces/kube-system/services"
+      # apply manifests
+      MANIFESTS="calico-configmap.yaml calico-node-sa.yaml calico-policy-controller-sa.yaml calico-ds.yaml calico-policy-controller.yaml kubedns-dep.yaml kubedns-svc.yaml default-backend-dep.yml default-backend-svc.yml ingress-controller-cm.yml ingress-controller-dep.yml ingress-controller-svc.yml"
```
rossf7: This is really nice and cleans up the templates a lot <3

templates.go
calvix: ( i have no idea what is the name so feel free to replace or give me the name)

templates.go
calvix: yeah this is the `yaml` indentation vs `script indentation` hell :)

templates.go
xh3b4sd: This should actually be configured via the clustertpr. See https://github.com/giantswarm/kubernetesd/blob/a8528d72f2a2ff7b37c5d6674684a03c543830c1/kubernetes/configmap.yaml#L90. I realize now that this is not specific to KVM so the config should be moved in kubernetesd. Anyway, at the moment here the URL should be replaced by something like this. 

```
curl -o /opt/bin/kubectl {{ .Cluster.KVM.Kubectl.GoogleAPI.URL }}
```

Lets please create an issue for this and address it in the next steps.

calvix: @xh3b4sd good idea, issues created https://github.com/giantswarm/k8scloudconfig/issues/136

rossf7: @calvix I'd say `KUBECTL_VERSION` would work. For hyperkube etc the k8s version is specified in clustertpr. So I think the var can just be for kubectl.

PR #139, opened 2017-05-29
wait for calico to be ready

calvix:
waiting for calico pods to be ready 
script is checking DS `desired` vs `ready`

there is some bash magic :(

ping @giantswarm/team-positive

### DIFF
--- a/templates.go
+++ b/templates.go
@@ -674,15 +674,38 @@ write_files:
       # wait for healthy master
       while [ "$(/opt/bin/kubectl get cs | grep Healthy | wc -l)" -ne "3" ];do sleep 1 && echo 'Waiting for healthy k8s'; done
 
-      # apply manifests
-      MANIFESTS="calico-configmap.yaml calico-node-sa.yaml calico-policy-controller-sa.yaml calico-ds.yaml calico-policy-controller.yaml kubedns-dep.yaml kubedns-svc.yaml default-backend-dep.yml default-backend-svc.yml ingress-controller-cm.yml ingress-controller-dep.yml ingress-controller-svc.yml"
+      # apply calico CNI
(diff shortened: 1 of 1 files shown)


templates.go
```diff
@@ -687,6 +687,29 @@ write_files:
               sleep 5s
           done
       done
+
+      # wait for healthy calico - we check for pods - desired vs ready
+      while 
+          eval $(/opt/bin/kubectl --kubeconfig=/etc/kubernetes/config/kubelet-kubeconfig.yml -n kube-system get ds calico-node | tail -1 | awk '{print "[ \"" $2"\" -eq \""$3"\" ] "}')
+          [ "$?" -ne "0" ]
```
nhlfr: Why not
```
until /opt/bin/kubectl --kubeconfig=/etc/kubernetes/config/kubelet-kubeconfig.yml -n kube-system get ds calico-node | tail -1 | awk '{print "[ \"" $2"\" -eq \""$3"\" ] "}'
do
    echo "Waiting for calico to be ready . . "
    sleep 3s
done
```
?

templates.go
```diff
@@ -687,6 +687,29 @@ write_files:
...
+          sleep 3s
+      done
+
+      # apply k8s addons
+      MANIFESTS="kubedns-dep.yaml kubedns-svc.yaml default-backend-dep.yml default-backend-svc.yml ingress-controller-cm.yml ingress-controller-dep.yml ingress-controller-svc.yml"
+
+      for manifest in $MANIFESTS
+      do
+          while
+              /opt/bin/kubectl --kubeconfig=/etc/kubernetes/config/kubelet-kubeconfig.yml apply -f /srv/$manifest
```
nhlfr: ditto

templates.go
```diff
@@ -687,6 +687,29 @@ write_files:
...
+              [ "$?" -ne "0" ]
+          sdo
```
nhlfr: typo?

PR #145, opened 2017-05-29
Updating DNS

puja108:
Upstream: https://github.com/kubernetes/kubernetes/tree/master/cluster/addons/dns

RFR

### DIFF
--- a/templates.go
+++ b/templates.go
@@ -320,6 +320,29 @@ write_files:
             - name: etcd-certs
               hostPath:
                 path: /etc/kubernetes/ssl/etcd
+- path: /srv/kubedns-cm.yaml
+  owner: root
+  permissions: 0644
+  content: |
+    apiVersion: v1
+    kind: ConfigMap
+    metadata:
+      name: kube-dns
+      namespace: kube-system
+      labels:
+        addonmanager.kubernetes.io/mode: EnsureExists
+- path: /srv/kubedns-sa.yaml
+  owner: root
+  permissions: 0644
+  content: |
+    apiVersion: v1
+    kind: ServiceAccount
+    metadata:
+      name: kube-dns
+      namespace: kube-system
+      labels:
+        kubernetes.io/cluster-service: "true"
+        addonmanager.kubernetes.io/mode: Reconcile
 - path: /srv/kubedns-dep.yaml
   owner: root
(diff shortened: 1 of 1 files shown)


templates.go
```diff
@@ -320,6 +320,29 @@ write_files:
...
+  permissions: 0644
+  content: |
+    apiVersion: v1
+    kind: ServiceAccount
+    metadata:
+      name: kube-dns
+      namespace: kube-system
+      labels:
+        kubernetes.io/cluster-service: "true"
+        addonmanager.kubernetes.io/mode: Reconcile
```
rossf7: For my interest what does reconcile do?

puja108: AFAIK if addon-manager pod is running, it will check on this resource and reconcile it with its "source-of-truth" so if someone manually changed things there or something broke it will reconcile the addon

puja108: currently it does nothing, it is just preparing for addon-manager

templates.go
```diff
@@ -394,15 +429,32 @@ write_files:
                 path: /readiness
                 port: 8081
                 scheme: HTTP
-              initialDelaySeconds: 30
+              initialDelaySeconds: 3
               timeoutSeconds: 5
           - name: dnsmasq
-            image: gcr.io/google_containers/kube-dnsmasq-amd64:1.4
+            image: gcr.io/google_containers/k8s-dns-dnsmasq-nanny-amd64:1.14.2
```
rossf7: I like calling the container nanny ;)

templates.go
```diff
@@ -394,15 +429,32 @@ write_files:
...
+              initialDelaySeconds: 60
+              timeoutSeconds: 5
+              successThreshold: 1
+              failureThreshold: 5
             args:
+            - -v=2
+            - -logtostderr
+            - -configDir=/etc/k8s/dns/dnsmasq-nanny
+            - -restartDnsmasq=true
+            - --
```
rossf7: This line looks wrong to me. @puja108 WDYT?

puja108: should be right https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dns/kubedns-controller.yaml.base#L124
and this is also what I tried on cluster zero

puja108:
Already did the update

Issue #136, opened 2017-05-29
fetch kubectl URL from configmap


calvix:
this has to be dynamic not hardcoded
https://github.com/giantswarm/k8scloudconfig/pull/135/files#diff-44e6c54f43ac9be31121348d531e697fR662

message from Tim: 
This should actually be configured via the clustertpr. See https://github.com/giantswarm/kubernetesd/blob/a8528d72f2a2ff7b37c5d6674684a03c543830c1/kubernetes/configmap.yaml#L90. I realize now that this is not specific to KVM so the config should be moved in kubernetesd. Anyway, at the moment here the URL should be replaced by something like this.
``
curl -o /opt/bin/kubectl {{ .Cluster.KVM.Kubectl.GoogleAPI.URL }}
``
Lets please create an issue for this and address it in the next steps.

JosephSalisbury:
I feel like we should be using our kubectl docker image, as opposed to downloading it

asymmetric:
@JosephSalisbury I agree, since that's what we do for other binaries (Hyperkube). What do the others think? @giantswarm/team-positive

xh3b4sd:
My two cents. 

- We should configure the URL as described in my PR comment.
- We should use whatever kubectl is appropriate. In case we have another one than the official one, fine. Just do the do.

asymmetric:
Does it make sense to allow different `kubectl` versions for KVM and AWS? i.e. having two separate flags?

xh3b4sd:
I do not think so because the guest clusters should be the same. Is there a special reason why you think this should be different?

asymmetric:
Nope, just checking as I haven't had much to do with kubernetesd or KVM much. All good.

PR #149, opened 2017-05-30
Run kubectl as a docker container

asymmetric:
Towards  https://github.com/giantswarm/k8scloudconfig/issues/136


templates.go
```diff
@@ -740,20 +740,20 @@ write_files:
...
+      while [ "$(/usr/bin/docker run --net=host --rm $IMAGE get cs | grep Healthy | wc -l)" -ne "3" ]; do sleep 1 && echo 'Waiting for healthy k8s'; done
 
       # apply calico CNI
       CALICO_FILES="calico-configmap.yaml calico-node-sa.yaml calico-policy-controller-sa.yaml calico-ds.yaml calico-policy-controller.yaml"
 
       for manifest in $CALICO_FILES
       do
           while
-              /opt/bin/kubectl --kubeconfig=/etc/kubernetes/config/kubelet-kubeconfig.yml apply -f /srv/$manifest
+              /usr/bin/docker run --net=host --rm $IMAGE apply -f /srv/$manifest
```
calvix: you need to mount `/srv` folder in order to work

templates.go
```diff
@@ -780,7 +780,7 @@ write_files:
       for manifest in $MANIFESTS
       do
           while
-              /opt/bin/kubectl --kubeconfig=/etc/kubernetes/config/kubelet-kubeconfig.yml apply -f /srv/$manifest
+              /usr/bin/docker run --net=host --rm $IMAGE apply -f /srv/$manifest
```
calvix: you need to mount `/srv` folder in order to work

xh3b4sd:
Just to make sure, this has to be updated in the operators before https://github.com/giantswarm/kubernetesd/pull/164 is merged/deployed.

asymmetric:
I think either way, there will be a small window where clusters won't work, but I'll follow your suggested order.

xh3b4sd:
You are probably right. We just broke the contract. Never mind.

PR #151, opened 2017-06-06
changed port for etcd domain from 2379 to 443

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/1197. There was a Slack conversation around this issue. See https://gigantic.slack.com/archives/C2MS2VB37/p1496756711183060. 

The idea is to align the ports under which etcd is accessible via domain. KVM and AWS should use `etcd.domain:443`. At the moment AWS uses `etcd.domain:2379`, which does not work in KVM, because the Kemps are configured to listen on `443`. The easiest way to align this is to change it for AWS where ELB configuration has to be modified.

### DIFF
--- a/templates.go
+++ b/templates.go
@@ -43,7 +43,7 @@ write_files:
       namespace: kube-system
     data:
       # Configure this with the location of your etcd cluster.
-      etcd_endpoints: "https://{{ .Cluster.Etcd.Domain }}:2379"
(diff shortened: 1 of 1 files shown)


xh3b4sd:
The changes in the AWs operator can be made separately because the dependencies are vendored. See https://github.com/giantswarm/aws-operator/issues/305.

PR #152, opened 2017-06-08
adding storage mount for local SSD

puja108:
This PR adds a storage mount for the (first) local SSD of the instance. It gets mounted to the docker directory to be used for containers and images.


puja108:
The units are there now but we need to make them conditional only for AWS, and ideas? @rossf7 @asymmetric @nhlfr ?

teemow:
@puja108 seems that the `storage` key only works with ignition. this doesn't work with cloud-configs. guess this has to be done with regular units.

puja108:
closing as outdated

PR #155, opened 2017-06-14
configure nginx ingress tcp-services-configmap flag

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/1526

### DIFF
--- a/templates.go
+++ b/templates.go
@@ -671,6 +671,7 @@ write_files:
             - /nginx-ingress-controller
             - --default-backend-service=$(POD_NAMESPACE)/default-http-backend
             - --configmap=$(POD_NAMESPACE)/ingress-nginx
+            - --tcp-services-configmap=$(POD_NAMESPACE)/ingress-nginx-tcp-services
             env:
               - name: POD_NAME
                 valueFrom:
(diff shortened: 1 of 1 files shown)


templates.go
```diff
@@ -671,6 +671,7 @@ write_files:
             - /nginx-ingress-controller
             - --default-backend-service=$(POD_NAMESPACE)/default-http-backend
             - --configmap=$(POD_NAMESPACE)/ingress-nginx
+            - --tcp-services-configmap=$(POD_NAMESPACE)/ingress-nginx-tcp-services
```
kopiczko: I will align this name with https://github.com/giantswarm/g8s-ingress-controller/pull/10.
I wanted it to be aligned with `--configmap=$(POD_NAMESPACE)/ingress-nginx` above.

xh3b4sd:
Why are we doing this here? AFAIK the ingress controller within the guest clusters do not need this, but the ingress controller on the host cluster.

JosephSalisbury:
We're aiming to use k8scloudconfig for host clusters too

kopiczko:
@xh3b4sd very good point. @calvix @r7vme can you point me to the place where host cluster's ingress is configured?

r7vme:
@kopiczko Here https://github.com/giantswarm/hive/blob/master/template_snippets/cloudconfig/g8s_master.yaml#L1022-L1139

xh3b4sd:
> We're aiming to use k8scloudconfig for host clusters too

Fine, but how do we handle differences then? The TCP configmap flag isn't a common thing across guest and host clusters.

kopiczko:
Closing this. This should be changed here: https://github.com/giantswarm/k8scloudconfig/pull/155#issuecomment-308701854

If [this](https://github.com/giantswarm/k8scloudconfig/pull/155#issuecomment-308669006) is something we want, let's track it, and add properly.

PR #156, opened 2017-06-15
Switch to kubeadm

### LINKED ISSUE #42: Use kubeadm
Currently we are setting up Kubernetes components with systemd units. It would be nice to use kubeadm for that.

Pros:
- we can potentially get rid of the majority of the content of our cloudconfigs and don't define our own logic about deploying kubernetes
- apiserver, scheduler will become self-hosted
- it would take us closed to having fully self-hosted Kubernetes

Things to think about:
- [x] ~~Bootstraping master by kubeadm generated the token. We have two potential solutions:~~
  - ~~generate the token before and inject it somehow to kubeadm~~
  - ~~figure out the secure way of distributing the token to workers~~
This is solved, you can provide custom token to `kubeadm init`

Things to do:
- [x] Generate a token for kubeadm in cert-operator and put it into a secret
- [x] Fetch a token in aws/kvm-operator and put it as a parameter for cloudconfig generation in aws/kvm-operator, like we currently do with TLS assets
- [x] Change cloudconfig to use kubeadm
  - [x] Remove all services related to Kubernetes, etcd, Calico
  - [x] Create oneshot systemd unit for masters doing `kubeadm init --token ...`
  - [x] Create oneshot systemd unit for workers doing `kubeadm join --token --`
  - [x] Create oneshot systemd unit for masters waiting for workers availability (`kubectl get nodes` or API call), which will be a depencency of all the other units
  - [x] Create oneshot systemd unit for master for setting up self-hosted Calico
- [ ] Use the newest k8scloudconfig with kubeadm support in
  - [ ] kvm-operator
  - [ ] aws-operator

nhlfr:
This change removes all Kubernetes-related systemd services (except kubelet), changes format of configs from cloud-init to ignition and uses kubeadm for deployment of Kubernetes.

Fixes #42


cloudconfig.go
```diff
@@ -8,24 +8,36 @@ import (
...
+// this library.
+type internalParams struct {
+	Token string
+	Params
+}
+
 type CloudConfig struct {
 	config   string
-	params   Params
+	params   internalParams
```
xh3b4sd: Why not just keeping the params as they are and adding a private token field? Also, what kind of token is that? I would like the structure and naming to reflect its identity or purpose. Now I know that we talk about the kube admin token, but in half a year some new employee will not.

token.go
```diff
@@ -0,0 +1,54 @@
...
+	"k8s.io/client-go/pkg/api"
+	"k8s.io/client-go/pkg/api/v1"
+	"k8s.io/client-go/pkg/watch"
+)
+
+const (
+	WatchTimeout = 90 * time.Second
+)
+
+func SearchToken(k8sClient kubernetes.Interface, clusterID string) (string, error) {
```
xh3b4sd: I think this is logic the operator should provide. For instance we have something similar in the operatorkit repo, but we agreed at some point to move such logic to the cert-operator in this specific case back then.

cloudconfig.go
```diff
@@ -8,24 +8,36 @@ import (
...
 	template string
 }
 
-func NewCloudConfig(template string, params Params) (*CloudConfig, error) {
+func NewCloudConfig(k8sClient kubernetes.Interface, template string, params Params) (*CloudConfig, error) {
```
xh3b4sd: The PR description stated that the cloud config stuff moved to ignition. Why do we need this here then still? Also, the `k8sClient` is not used in here.

templates.go
```diff
@@ -1,1606 +1,411 @@
...
+        Type=oneshot
+        Environment=IMAGE={{.Cluster.Kubernetes.Kubectl.Docker.Image}}
+	Environment=NAME=%p.service
+	EnvironmentFile=/etc/network-environment
+	ExecStartPre=-/usr/bin/docker stop $NAME
+        ExecStartPre=-/usr/bin/docker pull $NAME
+        ExecStart=/usr/bin/docker run \
+            --name $NAME \
+            $IMAGE \
+            kubectl apply -f http://docs.projectcalico.org/v2.3/getting-started/kubernetes/installation/hosted/kubeadm/1.6/calico.yaml
```
xh3b4sd: Is this the recommended way to do it? Just wondering if it is cool to rely on some other source which is also HTTP. I think it would be better to put the manifests in here directly.

token.go
nhlfr: This is something that both aws- and kvm-operator will need to do. I'm fine with moving that to operatorkit. I'm fully against implementing that in operators.

templates.go
nhlfr: I have two solutions in my mind

- using HTTPS and checking sum
- putting manifests here in this repo

what is better in your opinion?

xh3b4sd: Having the manifests in place.

token.go
xh3b4sd: I see. Lets keep it in here for now.

kopiczko: Actually we were thinking about moving it to `*tpr` project. In this case `tokentpr`.

token.go
```diff
@@ -0,0 +1,54 @@
...
+				return string(token[:]), nil
+			case watch.Deleted:
+				// Noop. Ignore deleted events.
+			case watch.Error:
+				return "", microerror.MaskAny(tokenRetrievalFailedError)
+			}
+		case <-time.After(WatchTimeout):
+			return "", microerror.MaskAnyf(tokenRetrievalFailedError, "timed out waiting for token")
+		}
+	}
```
kopiczko: Shouldn't we use cache Informer here, like we do for TPRs?

token.go
xh3b4sd: I see. It was this way around. Perfect. Then lets do this. :D

cloudconfig.go
nhlfr: Private fields cannot be used in templates

templates.go
puja108: Is there any way to keep the manifest in the repo (or another repo), but not inside this file? can we import or do we need to host the file ourselves (which I'd also be ok with)?

I'd like to keep as much out of this as possible, so we can have a much more modular setup and upgrades (to e.g. calico) can be super easy)

nhlfr:
@puja108 
1. No, I will add it.
2. Even without the full self-hosted mode, kubeadm needs kubelet - by default, it runs everything except kubelet in kubernetes.
3. We can add a field in `Params` struct for kubeadm parameters and override it in operators.

nhlfr:
I moved watcher to https://github.com/giantswarm/kubeadmtokentpr/pull/2

Please review that, I will switch back to that pull request when the stuff in tpr gets merged

templates.go
nhlfr: Maybe let's get that file by `git`? That will ensure secure connection and at the same time we will not need to store the file ourselves.

https://github.com/projectcalico/calico/blob/master/v2.3/getting-started/kubernetes/installation/hosted/kubeadm/1.6/calico.yaml

xh3b4sd: When we want to fetch it from remote but want secure communication you can just make the URL HTTPS.

puja108: git works for me, we can also make it a repo we manage, similar to the retagging we do with images

nhlfr: The advantage of using git is that we will be sure that we have a proper file. To do that with HTTPS URL we would need to check a checksum to be absolutely sure.

nhlfr: done, PTAL - I created a script which clones the repo with manifest and applies it

PR #157, opened 2017-06-22
Adds dynamic storage prorvisioner and cloud provider capability

puja108:
This:
- adds default storage class admission control plugin
- adds cloudprovider flag
- applies default storage class if available

towards https://github.com/giantswarm/giantswarm/issues/1588


templates.go
```diff
@@ -784,7 +797,7 @@ write_files:
               /usr/bin/docker run --net=host --rm -v /srv:/srv $KUBECTL apply -f /srv/$manifest
               [ "$?" -ne "0" ]
           do
-              echo "failed to apply /src/$manifest, retrying in 5 sec"
+              echo "failed to apply /srv/$manifest, retrying in 5 sec"
```
puja108: this is a minor boyscouting fix

templates.go
```diff
@@ -1107,6 +1120,7 @@ coreos:
       --logtostderr=true \
       --machine-id-file=/rootfs/etc/machine-id \
       --cadvisor-port=4194 \
+      --cloud-provider={{.Cluster.Kubernetes.CloudProvider}} \
```
rossf7: This won't work. The templating is for the Kubernetes block in clustertpr and it doesn't have a CloudProvider.

puja108: should be solved by https://github.com/giantswarm/clustertpr/pull/28 ?

rossf7: Yes, that will fix it.

puja108: done

rossf7:
FYI @nhlfr since you're working on the move to Ignition :)

PR #159, opened 2017-06-23
Fix if statement for submitting storage class

rossf7:


templates.go
```diff
@@ -776,7 +776,7 @@ write_files:
       done
 
       # apply default storage class
-      if [ -f /srv/default-storage-class.yaml]; then
+      if [ -f /srv/default-storage-class.yaml ]; then
```
rossf7: Adds missing space for the conditional

PR #170, opened 2017-07-28
Add kubeadm config

nhlfr:


calvix:
closing legacy PR.  when we move to `kubeadm` we will do lots of different changes

PR #171, opened 2017-08-01
Ingress version

rossf7:
Towards giantswarm/giantswarm#1695

This PR gets the Ingress controller docker image from the cluster custom object.


rossf7:
Yes <3 single dep updates :)

PR #173, opened 2017-08-09
Making IC K8s 1.7 compatible and setting replicas=#workers

puja108:
Also set `maxUnavailable: 1` to have higher availability during updates.

Still not sure where to get the worker number from.

### DIFF
--- a/templates.go
+++ b/templates.go
@@ -633,7 +633,7 @@ write_files:
         prometheus.io/port: '10254'
         prometheus.io/scrape: 'true'
     spec:
-      replicas: {{len .Params.Cluster.Workers}}
+      replicas: {{len .Cluster.Workers}}
       strategy:
         type: RollingUpdate
         rollingUpdate:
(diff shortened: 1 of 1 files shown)


puja108:
RFR

Thanks @kopiczko for the go template help with `len`

templates.go
```diff
@@ -633,37 +633,28 @@ write_files:
         prometheus.io/port: '10254'
         prometheus.io/scrape: 'true'
     spec:
-      replicas: 3
+      replicas: {{len .Params.Cluster.Workers}}
```
kopiczko: `{{.Params.Cluster.Workers | len}} is also possible. I don't know which is better.

puja108:
ok, seems `Params` can't be accessed like that, getting:

```
cloudconfig_test.go:34: template: cloudconfig:633:29: executing "cloudconfig" at <.Params.Cluster.Work...>: can't evaluate field Params in type cloudconfig.Params
```

kopiczko:
Skip the `.Params` part: `replicas: {{len .Cluster.Workers}}`

puja108:
now it worked

puja108:
@rossf7 do we still do/need vendoring dance for these changes?

PR #175, opened 2017-08-11
Turn etcd auth on

r7vme:
we have only two services (k8s api and calico)
that use etcd and both of them have certificates
and keys for access.

Fixes https://github.com/giantswarm/giantswarm/issues/1550


r7vme:
cc: @rossf7 @kopiczko just to make sure you've saw this. this change enables etcd authin guest clusters.

PR #178, opened 2017-08-23
some quickwin fixes for security

puja108:
towards https://github.com/giantswarm/giantswarm/issues/1720

tested on `56x77` on ginger

these fixes are all recommended in the CIS benchmark and do not break anything, but make our setup more secure, this is the quickest win we can have in that area, other setting will be harder and will need more thorough testing (mostly cause they need bootstrapping steps)


puja108:
ping @r7vme @corest @calvix we should do similar changes in host cluster

PR #180, opened 2017-08-23
Adding RBAC and PSP by default

puja108:
towards https://github.com/giantswarm/giantswarm/issues/1331 and https://github.com/giantswarm/giantswarm/issues/1788

### DIFF
--- a/templates.go
+++ b/templates.go
@@ -567,6 +567,7 @@ write_files:
           labels:
             k8s-app: default-http-backend
         spec:
+          serviceAccountName: default-backend
           containers:
           - name: default-http-backend
             image: gcr.io/google_containers/defaultbackend:1.0
@@ -1915,10 +1916,6 @@ coreos:
       ExecStartPre=/usr/bin/docker pull $IMAGE
       ExecStartPre=-/usr/bin/docker stop -t 10 $NAME
       ExecStartPre=-/usr/bin/docker rm -f $NAME
-      ExecStartPre=/bin/bash -c "while [ ! -f /etc/kubernetes/ssl/worker-ca.pem ]; do echo 'Waiting for /etc/kubernetes/ssl/worker-ca.pem to be written' && sleep 1; done"
-      ExecStartPre=/bin/bash -c "while [ ! -f /etc/kubernetes/ssl/worker-crt.pem ]; do echo 'Waiting for /etc/kubernetes/ssl/worker-crt.pem to be written' && sleep 1; done"
-      ExecStartPre=/bin/bash -c "while [ ! -f /etc/kubernetes/ssl/worker-key.pem ]; do echo 'Waiting for /etc/kubernetes/ssl/worker-key.pem to be written' && sleep 1; done"
-      ExecStartPre=/bin/sh -c "while ! curl --output /dev/null --silent --head --fail --cacert /etc/kubernetes/ssl/worker-ca.pem --cert /etc/kubernetes/ssl/worker-crt.pem --key /etc/kubernetes/ssl/worker-key.pem https://{{.Cluster.Kubernetes.API.Domain}}; do sleep 1 && echo 'Waiting for master'; done"
       ExecStart=/bin/sh -c "/usr/bin/docker run --rm --net=host --privileged=true \
       --name $NAME \
       -v /usr/share/ca-certificates:/etc/ssl/certs \
@@ -1963,7 +1960,7 @@ coreos:
       -v /usr/lib/os-release:/etc/os-release \
       -v /usr/share/ca-certificates/:/etc/ssl/certs \
       -v /var/lib/docker/:/var/lib/docker:rw \
-      -v /var/lib/kubelet/:/var/lib/kubelet:rw,rslave \
+      -v /var/lib/kubelet/:/var/lib/kubelet:rw,shared \
       -v /etc/kubernetes/ssl/:/etc/kubernetes/ssl/ \
       -v /etc/kubernetes/config/:/etc/kubernetes/config/ \
       -v /etc/cni/net.d/:/etc/cni/net.d/ \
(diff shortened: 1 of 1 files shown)


puja108:
RFR

The flag, bindings and roles are what has been running for a few days now on the staging cluster on `asgard`.

Only additional thing here is the rbac bootstrap as a first step in the addons script. Applying the bindings first and then the roles worked out for me in my manual tests.

This needs to be tested for example on `gauss`, as my testing was limited to manually changing an already running cluster. Should work, but I'd rather see it coming up as a fresh cluster before rolling this out.

puja108:
If this works we can do a similar PR for PodSecurityPolicies.

xh3b4sd:
Talking about the dev side. This looks to me like we should have some addon operator and/or an permissions operator running inside the guest clusters. We will have to template all these things for AWS/KVM/Azure the same way so we should have one solution for this to not need to copy/paste a lot of crap in each and every operator for each and every provider. An operator would make sense to always ensure such resources are properly set up. We recently had an issue with a customer where the ingress controllers within a guest cluster had just be gone and nobody knew why. Having an operator taking care about these resources would also prevent such scenarios.

JosephSalisbury:
I really have no if this is "right" or not :D

puja108:
fuck, I forgot the templating...next commit incoming

puja108:
@xh3b4sd you are right, this we should do in the future, actually we should just not use our own addon shit and go with the community addon-manager, which already has reconcilliation loops. I actually already prepared some of our deployments for this (see annotations). But this is a different issue that I was hoping to avoid by moving to `kubeadm` but we shelved that so...

puja108:
should be alright now, anyone know how we can test this? @rossf7 usually does that part for my PRs, so I'm kinda clueless. I just know there's some vendoring dance to do with the operators

puja108:
I tested podsecuritypolicies together with this and decided to add this to the PR so we only test once and then get 2 security features that should be used together anyway with one PR

puja108:
PSPs are now added, with this setup only cluster-admin role can run privileged containers (and mount privileged directories like hostpath).

Next:
I will add roles for the 2 PSPs. This way a cluster admin can flexibly assign privileged access. I will add a binding that gives all authenticated users restricted access only.

puja108:
Done, now need someone to help me with testing or tell me how to do it.

puja108:
Running (manually set up) on `56x77` on `ginger`

puja108:
nearly there, I changed the upstream restricted PSP to be a little bit less restricted (allowing users in containers running as root) as it otherwise blocks nearly anything users want to deploy if its not customized to be more secure. If customers ask for stricter policy they can create another one.

doing some final testing now

puja108:
final test came up on `gauss` cluster `z6apr`

puja108:
conformance test ran through without degradation (still 124/125)

Can someone LGTM this? you can check above cluster on gauss to see if it is running.

ping @calvix @teemow at some point we should introduce that in host cluster, too.

puja108:
Next I will write some docs for this I guess

templates.go
```diff
@@ -1049,7 +1467,6 @@ coreos:
       ExecStartPre=/usr/bin/docker pull $IMAGE
       ExecStartPre=-/usr/bin/docker stop -t 10 $NAME
       ExecStartPre=-/usr/bin/docker rm -f $NAME
-      ExecStartPre=/bin/sh -c "while ! curl --output /dev/null --silent --head --fail --cacert /etc/kubernetes/ssl/apiserver-ca.pem --cert /etc/kubernetes/ssl/apiserver-crt.pem --key /etc/kubernetes/ssl/apiserver-key.pem https://{{.Cluster.Kubernetes.API.Domain}}; do sleep 1 && echo 'Waiting for master'; done"
```
teemow: This looks like it doesn't belong into this PR. 

Why did you remove this?

templates.go
```diff
@@ -1488,10 +1906,6 @@ coreos:
       ExecStartPre=/usr/bin/docker pull $IMAGE
       ExecStartPre=-/usr/bin/docker stop -t 10 $NAME
       ExecStartPre=-/usr/bin/docker rm -f $NAME
-      ExecStartPre=/bin/bash -c "while [ ! -f /etc/kubernetes/ssl/worker-ca.pem ]; do echo 'Waiting for /etc/kubernetes/ssl/worker-ca.pem to be written' && sleep 1; done"
```
teemow: Same here: https://github.com/giantswarm/k8scloudconfig/pull/180/files#r136763850

teemow:
Are the docs already updated? I think this shouldn't be merged until we know all our examples work with RBAC enabled.

templates.go
puja108: We actually removed this from all units long time ago, seems we forgot some units. I needed to remove this for RBAC/PSP as this would prevent stuff from coming up. We are not running into race conditions without it, the components just retry, so it's not needed.

puja108:
@teemow we have updates for the xamples already prepared for RBAC, I'll check on the rest.

puja108:
Docs TODO:

- [ ] update examples in docs-content (https://github.com/giantswarm/docs-content/pull/255)
- [ ] update examples outside docs-content (prom & elastic open)
- [ ] add docs for usage re usr mgmt

puja108:
Can be merged after https://github.com/giantswarm/kubernetes-elastic-stack/pull/34 and https://github.com/giantswarm/docs-content/pull/257

puja108:
Can I get an LGTM?

calvix:
also merge master before updating

puja108:
docs PRs are RFR

puja108:
waiting for #1798

jgsqware:
@puja108  Regarding https://github.com/giantswarm/giantswarm/issues/1719, **2.11** expect authorization plugins to validate. I think RBAC and PSP enforce it too. WDYT?

puja108:
closing in favor of https://github.com/giantswarm/k8scloudconfig/pull/203

Issue #179, opened 2017-08-23
Make Calico ranges configurable


r7vme:
Currently CALICO_IPV4POOL_CIDR set to default 192.168.0.0/16

JosephSalisbury:
Calico ranges are now fully configurable, and correct values are rolled out for all installations. Verified by the wonderful mr @r7vme. Closing.

PR #181, opened 2017-08-24
Makes calico cidr configurable

JosephSalisbury:
Towards https://github.com/giantswarm/k8scloudconfig/issues/179

### DIFF
--- a/cloudconfig_test.go
+++ b/cloudconfig_test.go
@@ -2,7 +2,11 @@ package cloudconfig
 
 import (
 	"encoding/base64"
+	"strings"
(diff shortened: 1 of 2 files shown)


JosephSalisbury:
I actually think this is all we need to change here, which is nice.

JosephSalisbury:
And, to help with reference:
```
  cluster:
    calico:
      cidr: 16
      domain: calico.y22nx.k8s.eu-central-1.aws.cps.vodafone.com
      mtu: 1430
      subnet: 172.20.0.0
```

PR #184, opened 2017-08-31
Docker CIS : 2.1

jgsqware:
2.1 - Ensure network traffic is restricted between containers on the default bridge


puja108:
Do you have this tested and working somewhere? We should check that it does not interfere with our networking setup

jgsqware:
I've tested with the DNS resolution. Have another test in mind?

jgsqware:
@calvix or @r7vme It disable the docker default bridge, `docker network inspect bridge` doesn't report any container attached to it. DNS resolution and cross container communication works.
Do you think to any other tests before merge?

r7vme:
Hold ON. It adds explicit drop to FORWARD chain.

let's discuss, as it will drop any unknow ip address traffic....

r7vme:
Okay `icc=false` should be ok as we are not using docker briodged network.

For me it set drop on following rule, not whole system.

`-A FORWARD -i docker0 -o docker0 -j DROP`

calvix:
ok fine :)

jgsqware:
Close in favor of #204

Issue #147, opened 2017-05-30
Use addon-manager


puja108:
The K8s addon-manager (https://github.com/kubernetes/kubernetes/tree/master/cluster/addons/addon-manager) can ensure that certain manifests (by default in '/etc/kubernetes/addons') are ensured to be running (created when not there) or even reconciled (changed back to defaults if changed).

The current version can be run like following: https://github.com/kubernetes/kubernetes/blob/master/test/kubemark/resources/manifests/kube-addon-manager.yaml

It runs in hostnetwork so it can be there before Calico and can actually even schedule calico and other things, not sure how it handles priorities so we still might want to have calico outside. I saw some things that EnsureExists has a higher prio than Reconcile. Otherwise it can just handle all the addons that we need running, incl. DNS, IC, StorageClasses, RBAC roles and bindings.

Things need to be in `kube-system` namespace and have a label `addonmanager.kubernetes.io/mode: EnsureExists` or `addonmanager.kubernetes.io/mode: Reconcile` in their yamls. The yamls should live in `/etc/kubernetes/addons` and don't need to be applied manually anymore AFAIK.

Not sure if it also handles updates, but that could be interesting to update things running in clusters without needing to roll anything.

WDYT @rossf7 @calvix but maybe not so urgent.

calvix:
That would be awesome, we still have to figure out how to do this with calico but anyway good direction.

rossf7:
Yes I think we should close. This is covered by giantswarm/giantswarm#1902 and we'll be using `chart-operator` to deploy addons.

calvix:
Yep we can close.

PR #186, opened 2017-09-12
kernel-hardening

calvix:
kernel hardening params

* sysctl kernel values see - https://github.com/giantswarm/vodafone/issues/136
* removing special groups from core user

from the ticket i have only removed - `net.ipv4.conf.all.forwarding = 0` becasue we need forwarding


calvix:
The last state is tested and conformance test is having a positive result and I tested all functionality I could think of and not found any issue.

calvix:
@xh3b4sd we could do it sam as we do it on `mayu` we have one general profile for all machines and then worker/master specific in separate files and this will get merged 
https://github.com/giantswarm/hive/blob/master/templates/last_stage_cloudconfig.yaml#L292-L296

xh3b4sd:
I would be in favour of some operator. I remember a case where a customer's ingress controllers have been vanished from his guest cluster and nobody noticed or knew why it was gone. An operator can ensure the desired state continuously. Separating templates does not help with this problem.

puja108:
@xh3b4sd for what you mean there's already a tool and we have an issue in sig operator: https://github.com/giantswarm/k8scloudconfig/issues/147

PR #187, opened 2017-09-12
fix coreos hardening -  sshd

calvix:


calvix:
so disabling sshd socket is basically disabling ssh service, it was working when I tested it but apparently after while no new connection can be established.

templates.go
```diff
@@ -914,12 +914,6 @@ coreos:
     content: |
       {{range .Content}}{{.}}
       {{end}}{{end}}
-  - name: sshd.service
-    enable: true
-  - name: sshd.socket
-    enable: false
-    mask: true
```
puja108: should we maybe leave the masking or is that also problematic

calvix: yeah, its also problematic

PR #189, opened 2017-09-14
updated calico to 2.5.1

puja108:
based on upstream https://docs.projectcalico.org/v2.5/getting-started/kubernetes/installation/hosted/calico.yaml

running succesfully on `totl0` on `ginger`

### DIFF
--- a/templates.go
+++ b/templates.go
@@ -28,12 +28,12 @@ write_files:
   owner: root
   permissions: 644
   content: |
-    # Calico Version v2.2.1
-    # http://docs.projectcalico.org/v2.2/releases#v2.2.1
+    # Calico Version v2.5.1
+    # https://docs.projectcalico.org/v2.5/releases#v2.5.1
     # This manifest includes the following component versions:
-    #   calico/node:v1.2.1
-    #   calico/cni:v1.8.3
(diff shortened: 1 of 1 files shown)


templates.go
```diff
@@ -145,6 +148,9 @@ write_files:
                 # Set Felix logging to "info"
                 - name: FELIX_LOGSEVERITYSCREEN
                   value: "info"
+                # Set MTU for tunnel device used if ipip is enabled
+                - name: FELIX_IPINIPMTU
+                  value: "1440"
```
puja108: Could this newly added MTU setting make our custom MTU setting in the config map obsolete? or do we need to set it to the same value? I just took the value they had in https://docs.projectcalico.org/v2.5/getting-started/kubernetes/installation/hosted/calico.yaml

calvix: no, i would rather not screw with MTU, it can look like it's working and then way later we find some hidden problems

calvix:
could you possible also update host cluster and aws?
https://github.com/giantswarm/hive/blob/master/template_snippets/cloudconfig/g8s_master.yaml#L676
https://github.com/giantswarm/aws-terraform/blob/master/resources/cloud-config/g8s-master.yaml.tmpl#L127

PR #191, opened 2017-09-19
increase-conntrack-limit

calvix:
towards https://github.com/giantswarm/adidas/issues/134

in future, we may wanna have this configurable, but for now, it's better if we don't have to configure it manually


calvix:
ip conntrack is for tracking connection for iptables
if this limit is reached iptables is not able to track more connection so the rest is dropped

the implication is just small increase of RAM usage

calvix:
closing in favor of https://github.com/giantswarm/k8scloudconfig/pull/192

PR #192, opened 2017-09-19
increase-conntrack-limit

calvix:
towards https://github.com/giantswarm/adidas/issues/134

in future, we may wanna have this configurable, but for now, it's better if we don't have to configure it manually

ip conntrack is for tracking connection for iptables
if this limit is reached iptables is not able to track more connection so the rest is dropped

the implication is just small increase of RAM usage

### DIFF
--- a/templates.go
+++ b/templates.go
@@ -1114,6 +1114,7 @@ coreos:
       --proxy-mode=iptables \
       --logtostderr=true \
       --kubeconfig=/etc/kubernetes/config/proxy-kubeconfig.yml \
+      --conntrack-max-per-core 131072 \
       --v=2"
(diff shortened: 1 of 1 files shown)


templates.go
```diff
@@ -1114,6 +1114,7 @@ coreos:
       --proxy-mode=iptables \
       --logtostderr=true \
       --kubeconfig=/etc/kubernetes/config/proxy-kubeconfig.yml \
+      --conntrack-max-per-core 131072 \
```
corest: do we need that on master nodes?

r7vme: Let's have it on all nodes for homogeneity.

calvix: exactly

PR #194, opened 2017-09-20
set --disable-legacy-registry to true

jgsqware:
This will disable Docker Legacy Registry v1


puja108:
@jgsqware also don't forget assigning the PR to yourself (I also forget that a lot)

PR #196, opened 2017-09-20
disable userland proxy

jgsqware:
Disable Userland proxy for Docker


calvix:
Can you always point a little bit more info about the changes?
For example, what does userland proxy for docker?

That can really speed up the process because we don't have to google the same thing you already googled.

jgsqware:
Close in favor of #206

PR #197, opened 2017-09-21
Make master unschedulable with taint

r7vme:
use correct label and taint for master (leave existing label for compatibility)
See source: https://github.com/kubernetes/kubernetes/pull/41835

use proper tolerations for Calico from kubeadm.
Default Calico manifests contain outdated toleraiotns
See source: https://github.com/projectcalico/calico/blob/master/v2.5/getting-started/kubernetes/installation/hosted/kubeadm/1.6/calico.yaml#L71-L77

Discussed in detail here: https://github.com/giantswarm/aws-terraform/pull/91

### DIFF
--- a/v_0_1_0/master_template.go
+++ b/v_0_1_0/master_template.go
@@ -107,10 +107,14 @@ write_files:
             k8s-app: calico-node
           annotations:
             scheduler.alpha.kubernetes.io/critical-pod: ''
-            scheduler.alpha.kubernetes.io/tolerations: |
-              [{"key": "dedicated", "value": "master", "effect": "NoSchedule" },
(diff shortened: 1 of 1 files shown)


templates.go
```diff
@@ -1196,10 +1200,10 @@ coreos:
       --cluster-domain={{.Cluster.Kubernetes.Domain}} \
       --network-plugin=cni \
       --register-node=true \
-      --register-schedulable=false \
+      --register-with-taints=node-role.kubernetes.io/master=:NoSchedule \
```
kopiczko: Why we need to add new label? Is it enough to add `role=master:NoSchedule`?

r7vme: This label is used in community and components like Calico rely on this label. See links in description

PR #198, opened 2017-09-21
started versioning by separate folder

xh3b4sd:
This is an effort to sort out how to version templates of this library. Please let me know what you think and lets iterate here in this PR. I am open to suggestions.


template.go
```diff
@@ -0,0 +1,26 @@
+package cloudconfig
+
+import (
+	"github.com/giantswarm/microerror"
+
+	"github.com/giantswarm/k8scloudconfig/v_0_1_0_3500632"
```
xh3b4sd: I thought about how to version this. There are certain limitations because of golang package name rules. For instance a package name has to start with a letter. The package name can be alphanumeric plus underscores. I want the package name and the import path to be equal. Here I tried to show a semver version, which in this case is `0.1.0`. I made this version up. The last part of the package name is the first 7 digits of the git commit hash. I am not sure we want or need this.

template.go
```diff
@@ -0,0 +1,26 @@
...
+)
+
+type Template struct {
+	Master string
+	Worker string
+}
+
+func NewTemplate(v string) (Template, error) {
```
xh3b4sd: This would be the interface for library users to get versioned templates.

v_0_1_0_3500632/cloudconfig.go
```diff
@@ -1,4 +1,4 @@
-package cloudconfig
+package v_0_1_0_3500632
```
xh3b4sd: Everything else moved to the versioned package.

template.go
kopiczko: How do you know what commit has it will be before committing?

template.go
kopiczko: I'd like to see `type Version string` with constants. It would introduce some level of type safety.

template.go
xh3b4sd: Right, the hash does not work that way. Good catch. I will just remove it.

xh3b4sd: Done.

template.go
xh3b4sd: Good idea. Done.

template.go
```diff
@@ -0,0 +1,32 @@
+package cloudconfig
+
+import (
+	"github.com/giantswarm/microerror"
+
+	"github.com/giantswarm/k8scloudconfig/v_0_1_0"
+)
+
+type version string
```
kopiczko: This should be exported. `NewTemplate` is exported.

kopiczko: Oh, I got it. you don't want to allow creating custom ones outside the package. Clever!

kopiczko: Not sure godoc-wise though. I leave it up to you :D

template.go
```diff
@@ -0,0 +1,32 @@
...
+
+const (
+	V_0_1_0 version = "v_0_1_0"
```
kopiczko: Maybe we want to add `Latest` alias. Uncertain.

rossf7: A latest alias could be useful for local development and test installations. IMO for production clusters it should always come from the cluster custom object.

xh3b4sd: I think we can add this later if we really want to. The update story is super big and I am still trying to wrap my head around. I want to make baby steps and get a look and feel. Thus I would keep the latest version out of this PR at least. Lets iterate. <3

xh3b4sd:
> I think we need to get the versioning scheme pinned down. As that's an important part of the total updates picture.

I agree that we need to pin the scheme down. I just think that it is good to go with a first version and let others think about the desired version scheme meanwhile. I would like to make some progress on our internal implementations and I am fine with having just some version. We can then adapt later as soon as we gained experiences.

xh3b4sd:
@puja108 even though this is merged it does not mean the version scheme is written in stone. We can just change it when we think it should be different. There are also open questions towards mapping and how to manage the user facing versions. We can align this as soon as we are certain how to roll. I just go with this for internal implementations.

PR #200, opened 2017-09-21
mask-flanneld

calvix:
* mask flanneld on guest cluster
* add mask for fleet also for worker ( I have forgotten it in the previous PR for hardening)

### DIFF
--- a/templates.go
+++ b/templates.go
@@ -1219,6 +1219,10 @@ coreos:
     enable: false
     mask: true
     command: stop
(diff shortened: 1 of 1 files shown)


xh3b4sd:
Please see https://github.com/giantswarm/k8scloudconfig/pull/198. We want to start with some versioned approach. It would probably make sense to recreate this PR to prevent weird merge conflict magic.

PR #203, opened 2017-09-25
adding RBAC and PSP

puja108:
redo of https://github.com/giantswarm/k8scloudconfig/pull/180 for new directory structure

### DIFF
--- a/v_0_1_0/master_template.go
+++ b/v_0_1_0/master_template.go
@@ -679,6 +679,7 @@ write_files:
                         values:
                         - nginx-ingress-controller
                   topologyKey: kubernetes.io/hostname
+          serviceAccountName: nginx-ingress-controller
           containers:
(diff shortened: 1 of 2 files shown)


puja108:
btw, that +/- 1 line difference to the original PR is because of line 234 (no actual change)

calvix:
yeah probably good idea to retest it on gauss

puja108:
Yeah, I'd also like to make sure after so many things have changed. Would `ginger` also be ready for testing this or do we still need changes to aws-operator to support the new format?

xh3b4sd:
The AWS operator has to be adapted to use the versioned k8scloudconfig. AFAIK Joe just pinned the old version. The task is not that difficult/big but let me know when you need help. Also may check the KVM operator how it is done there.

xh3b4sd:
AWS operator is done.

PR #204, opened 2017-09-26
Disable default bridge

jgsqware:
2.1 - Ensure network traffic is restricted between containers on the default bridge

This will disable default `bridge` network on docker daemon.

icc=false should be ok as we are not using docker bridged network.

For me it set drop on following rule, not whole system.

-A FORWARD -i docker0 -o docker0 -j DROP


jgsqware:
already been validated by @r7vme from #184

PR #210, opened 2017-09-29
Rolling update of Calico nodes

r7vme:
By default user needs manually delete pods
after new DaemonSet version applied.

related changes:
https://github.com/giantswarm/aws-terraform/pull/118
https://github.com/giantswarm/hive/pull/146


xh3b4sd:
How does the migration path look like? Will we take care manually?

PR #214, opened 2017-10-04
fix-etcd-defrag-cert-path

calvix:
my bad copy-paste from baremetal template

### DIFF
--- a/v_0_1_0/master_template.go
+++ b/v_0_1_0/master_template.go
@@ -1136,9 +1136,9 @@ coreos:
         $IMAGE \
         etcdctl \
         --endpoints https://{{ .Cluster.Etcd.Domain }}:443 \
-        --cacert /etc/etcd/etcd-ca.pem \
-        --cert /etc/etcd/etcd.pem \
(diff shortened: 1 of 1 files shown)


v_0_1_0/master_template.go
```diff
@@ -1129,16 +1129,16 @@ coreos:
       ExecStartPre=-/usr/bin/docker rm  $NAME
       ExecStartPre=-/usr/bin/docker pull $IMAGE
       ExecStart=/usr/bin/docker run \
-        -v /etc/giantswarm/g8s/ssl/etcd/:/etc/etcd \
+        -v /etc/kubernetes/ssl/etcd/:/etc/etcd \
```
xh3b4sd: Has this implications for the flannel-operator and the pods started for flanneld?

calvix: no, as this is the guest cluster only

we still have inconsistency  of this folder (red is bare-metal host cluster and green is guest cluster)

nothing really changed i just badly copied

PR #215, opened 2017-10-05
k8s-use-internalIP-for-kubelet

calvix:
fix for https://github.com/giantswarm/k8scloudconfig/pull/207
by default api is accessing machines via hostnames which are not existing in case of kvm


v_0_1_0/master_template.go
```diff
@@ -1723,6 +1723,7 @@ coreos:
       --insecure_bind_address=0.0.0.0 \
       --insecure_port={{.Cluster.Kubernetes.API.InsecurePort}} \
       --kubelet_https=true \
+      --kubelet-preferred-address-types=InternalIP \
```
xh3b4sd: How does the API server know about the internal IP?

calvix: kubelet register itself with that info

PR #216, opened 2017-10-06
k8s-use-localhost-for-etcd

calvix:
together with https://github.com/giantswarm/kubernetesd/pull/280

towards https://github.com/giantswarm/giantswarm/issues/1905

### DIFF
--- a/v_0_1_0/master_template.go
+++ b/v_0_1_0/master_template.go
@@ -1734,7 +1734,7 @@ coreos:
       --admission-control=NamespaceLifecycle,LimitRanger,ServiceAccount,ResourceQuota,DefaultStorageClass,PodSecurityPolicy \
       --cloud-provider={{.Cluster.Kubernetes.CloudProvider}} \
       --service-cluster-ip-range={{.Cluster.Kubernetes.API.ClusterIPRange}} \
-      --etcd-servers=https://{{ .Cluster.Etcd.Domain }}:443 \
+      --etcd-servers=https://127.0.0.1:2379 \
       --etcd-cafile=/etc/kubernetes/ssl/etcd/server-ca.pem \
       --etcd-certfile=/etc/kubernetes/ssl/etcd/server-crt.pem \
       --etcd-keyfile=/etc/kubernetes/ssl/etcd/server-key.pem \


v_0_1_0/master_template.go
```diff
@@ -1734,7 +1734,7 @@ coreos:
       --admission-control=NamespaceLifecycle,LimitRanger,ServiceAccount,ResourceQuota,DefaultStorageClass,PodSecurityPolicy \
       --cloud-provider={{.Cluster.Kubernetes.CloudProvider}} \
       --service-cluster-ip-range={{.Cluster.Kubernetes.API.ClusterIPRange}} \
-      --etcd-servers=https://{{ .Cluster.Etcd.Domain }}:443 \
+      --etcd-servers=https://127.0.0.1:2379 \
```
xh3b4sd: I am not sure which implications this has. Since we have etcd proxies on prem KVM is safe, but how does this work for AWS now? Is this safe to merge? What happens to existing clusters that get updated? I guess they will fail because they will miss the IPSANs in the certificates.

calvix: Will certificates not reconcile? If not then yes we will have an issue that is not easily solvable

xh3b4sd: Cert-operator was one of the first operators and is still quite dumb. Certificates are not really reconciled AFAIK. Even if they would be reconciled we still miss the push updates in general. E.g. you added the loopback IP to kubernetesd when it issues certificate TPOs. There is no mechanism to update existing TPOs yet. We need a migration path for this and a general idea of how to automate such situations.

calvix: I see. Than this is blocked for some time until we figure how to do it.

xh3b4sd: We could write some migration code in the cert operator. When it gets deployed it updates the certificates within the secrets and after that we just remove the migration code. Then we can merge this. Let me know when I can help with this.

r7vme:
Why we are not merging this?

calvix:
Because if existing cluster on KVM got updated by this the certificate won't have IP sans in the certs.
And that will fail.

calvix:
As all kvm clusters are manually updated and reconciled we can continue with this

https://github.com/giantswarm/giantswarm/issues/1905#issuecomment-341401252

PR #219, opened 2017-10-18
Update k8s to 1.8.1

puja108:
Towards https://github.com/giantswarm/giantswarm/issues/1901

To control k8s versions through our k8scc I integrated the hyperkube image and version instead of taking the templated variable, we should get rid of that after this PR. This helps us to move to multi-version support in the future.

Other than the update to k8s 1.8.1 I had to fix the kubeconfigs and remove the deprecated flags called api-server/master from the k8s components.

I also added an additional spec to our privileged PSP that came in with k8s 1.8

### DIFF
--- a/v_0_1_0/master_template.go
+++ b/v_0_1_0/master_template.go
@@ -1033,6 +1033,7 @@ write_files:
     metadata:
       name: privileged
     spec:
+      allowPrivilegeEscalation: true
       fsGroup:
         rule: RunAsAny
       privileged: true
@@ -1258,6 +1259,7 @@ write_files:
     - name: local
       cluster:
         certificate-authority: /etc/kubernetes/ssl/apiserver-ca.pem
+        server: https://{{.Cluster.Kubernetes.API.Domain}}
     contexts:
     - context:
         cluster: local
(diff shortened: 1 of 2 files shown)


puja108:
This is working and tested on gauss cluster `asrx8`.

@xh3b4sd added you to the review just FYI as this PR includes a move from installations-based k8s version to k8scloudconfig-based k8s version.

v_0_1_0/master_template.go
```diff
@@ -1803,7 +1805,7 @@ coreos:
       RestartSec=0
       TimeoutStopSec=10
       EnvironmentFile=/etc/network-environment
-      Environment="IMAGE={{.Cluster.Kubernetes.Hyperkube.Docker.Image}}"
+      Environment="IMAGE=quay.io/giantswarm/hyperkube:v1.8.1_coreos.0"
```
xh3b4sd: A constant for this would be super useful. We should maybe have a `version.go` in each folder which contains such information to have one place we can look at. I will think about this a bit more.

calvix: This make sense to me, we can update  much easier then

puja108: true, that would be cool, at least for hyperkube I would welcome this, good thinking @xh3b4sd <3

xh3b4sd: <3

PR #220, opened 2017-10-19
update calico to 2.6.2

puja108:
towards https://github.com/giantswarm/giantswarm/issues/1900

syncing with upstream https://docs.projectcalico.org/v2.6/getting-started/kubernetes/installation/hosted/hosted

policy controller has been renamed to kube-controllers

needs testing, will do so on gauss

### DIFF
--- a/v_0_1_0/master_template.go
+++ b/v_0_1_0/master_template.go
@@ -288,12 +288,6 @@ write_files:
       annotations:
         scheduler.alpha.kubernetes.io/critical-pod: ''
     spec:
-      tolerations:
-      - key: node-role.kubernetes.io/master
-        operator: Exists
-        effect: NoSchedule
-      - key: CriticalAddonsOnly
-        operator: Exists
       # The policy controller can only have a single active instance.
       replicas: 1
       strategy:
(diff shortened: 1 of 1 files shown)


calvix:
can calico 2.6.2 run on k8s 1.7 ?

puja108:
@calvix yes it can, but as guest clusters are already 1.8.1 I don't think we have any issue, or are you asking for host clusters?

calvix:
Yep i was thinking for host cluster ;)

puja108:
seems I need to update our kubectl version there

calvix:
we should always have kubectl equal to k8s version :)

puja108:
tested on gauss, but will need to do full from test 0 again after circle is done building

puja108:
> we should always have kubectl equal to k8s version :)

yes, we were using 1.6.4 btw, just cause nobody updated this thing since then, boyscouting FTW

calvix:
We for sure wanna use the same version for that, but it might be way better to put it outside of template and use const for it so we can easily change versions of all components in place and don't have to worry about forgetting it.

It may be out of the scope of the PR as it requires some code changes.

puja108:
I would like to handle the kubectl version together with hyperkube version, but it must be inside the k8scc version (folder), be that a version.go like @xh3b4sd suggested or whatever

xh3b4sd:
I think this is a good shout. We have to start bundling and versioning components and the business logic that manages these together. This will make our service quality more stable and our maintainability easier for us humans.

puja108:
tested and came up perfectly on `ocx0p` on ginger

Issue #162, opened 2017-07-04
clean calico config at worker start


calvix:
When for some reason worker is recreated (ie pod delete/scale or w/e reason) calico wont  start because there will already be records with same IP and different machine ID/hostname

we should add `systemd` unit that will clear etcd  

this is how i did that manually:
```
data=`sudo etcdctl --endpoints=https://etcd.d073w.g8s.kc.cg.internal:443 --cert-file /etc/kubernetes/ssl/etcd/client-crt.pem --key-file /etc/kubernetes/ssl/etcd/client-key.pem --ca-file /etc/kubernetes/ssl/etcd/client-ca.pem ls --recursive /calico | grep dkpen`

for i in $data; do sudo etcdctl --endpoints=https://etcd.d073w.g8s.kc.cg.internal:443 --cert-file /etc/kubernetes/ssl/etcd/client-crt.pem --key-file /etc/kubernetes/ssl/etcd/client-key.pem --ca-file /etc/kubernetes/ssl/etcd/client-ca.pem rm $i >/dev/null 2>/dev/null; done;

```
`dkpen` is the worker name taken from pod name `worker-dkpen-3853139292-n0m54`

r7vme:
I've started to work on this task.

After asking in Calico stack they recommened to take look on 

https://github.com/caseydavenport/calico-node-controller

I was able to test it with on-prem guest cluster. Died node were successfully removed from Calico state.

What is missing:
1) We need to unify node names in our kubernetes (Use hostnames everywhere, currently AWS use hostnames, onprem IP addresses). Calico by default uses hostnames.
2) "calico-node-controller" need to be properly packed to Docker/Helm
3) "calico-node-controller" is raw PoC. So i need evaluate it more and work on TODO items that mentioned in README.

calvix:
duplicate https://github.com/giantswarm/k8scloudconfig/issues/162#issuecomment-331082448

xh3b4sd:
@r7vme can you give a status update? AFAIK you worked on this and there have been PRs in the calico node controller fork.

r7vme:
Sure.

- Initial worked version https://github.com/giantswarm/calico-node-controller
- Installed in `gauss` (see kube-system namespace)
- [Manual test plan](https://docs.google.com/document/d/1_KcPdkmxH13A-Z13RdoYcOUiUd6ANCUcAq-IEtlOsUc/edit)

TODO:
- Execute test plan in `gauss`, if ok proceed further
[...]

PR #221, opened 2017-10-20
Adds calico-node-controller

r7vme:
Part of: https://github.com/giantswarm/k8scloudconfig/issues/162

### DIFF
--- a/v_0_1_0/master_template.go
+++ b/v_0_1_0/master_template.go
@@ -10,6 +10,79 @@ users:
        - "{{ $user.PublicKey }}"
 {{end}}
 write_files:
+- path: /srv/calico-node-controller-sa.yaml
+  owner: root
+  permissions: 644
+  content: |
+    apiVersion: v1
+    kind: ServiceAccount
+    metadata:
+      name: calico-node-controller
+      namespace: kube-system
+- path: /srv/calico-node-controller.yaml
+  owner: root
+  permissions: 644
+  content: |
+    apiVersion: extensions/v1beta1
+    kind: Deployment
+    metadata:
+      name: calico-node-controller
+      namespace: kube-system
+      labels:
+        k8s-app: calico-node-controller
+    spec:
+      replicas: 1
+      strategy:
+        type: Recreate
+      template:
+        metadata:
+          name: calico-node-controller
+          namespace: kube-system
+          labels:
+            k8s-app: calico-node-controller
+        spec:
+          serviceAccountName: calico-node-controller
+          containers:
+            - name: calico-node-controller
+              image: quay.io/giantswarm/calico-node-controller
+              env:
+                # The location of the Calico etcd cluster.
+                - name: ETCD_ENDPOINTS
+                  valueFrom:
+                    configMapKeyRef:
+                      name: calico-config
+                      key: etcd_endpoints
+                # Location of the CA certificate for etcd.
+                - name: ETCD_CA_CERT_FILE
+                  valueFrom:
+                    configMapKeyRef:
+                      name: calico-config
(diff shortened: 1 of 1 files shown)


v_0_1_0/master_template.go
```diff
@@ -10,6 +10,94 @@ users:
...
+- path: /srv/calico-node-controller-sa.yaml
+  owner: root
+  permissions: 644
+  content: |
+    apiVersion: v1
+    kind: ServiceAccount
+    metadata:
+      name: calico-node-controller
+      namespace: kube-system
+- path: /srv/calico-node-controller-cm.yaml
```
puja108: can't we just reuse the CM we already have for calico? looks pretty similar

v_0_1_0/master_template.go
```diff
@@ -10,6 +10,94 @@ users:
        - "{{ $user.PublicKey }}"
 {{end}}
```
kopiczko: Should it go to the v_0_1_0? That would mean we have inconsistent cluster version again. WDYT?

r7vme: i can not answer this. May be @xh3b4sd ? from what i see we are not using versioninig yet

xh3b4sd: I think we should start version our own code better. When a new feature is added lets have a new folder.

r7vme:
@puja108 done. Thanks good catch

v_0_1_0/master_template.go
puja108: We said we hold off on versioning until we have a agreed upon approach to it and can easily change/use it. Just adding versions in k8scc now would again break stuff I fear (e.g. operators have fixed version in code currently)

r7vme: Do we have a code that already use this v_X_Y_Z folders?

Thanks @kopiczko for raising this, but i think we should continue this discussion out of this PR.

r7vme:
tested in gauss all good. merging

v_0_1_0/master_template.go
xh3b4sd: I do not agree that "overdoing" versioning breaks something. The opposite is the case. We potentially just break stuff if we throw everything into the same branch of our code. Creating a new version folder here is more work but also makes us starting to gain experiences on how to actually version our software. There will not be the final solution that makes everything easier. We have to do it. We have to learn from it. There is no discussion necessary that we will have to copy code into versioned folders. We have to do it. So why not start now? How the folders are named is TBD though. But that we have to do it is not questionable AFAIK.

puja108: I would like to have full support in the operators for this (i.e. more than one folder) before we just add a folder. Full support means we just can choose based on TPR or whatever and have no hardcoded thing in the code that chooses a version. I am not a big fan of changing stuff in dependencies without having the support for it in the other components ready, especially if it's such central components like k8scc and cluster operators where we are rolling out new version a lot recently. These changes are hard breaks and they block others from getting on with their issues that they have priority on.
The case where we break stuff hasn't happened a lot lately, since we do actually test on gauss (and hopefully on lycan) when we change things. I don't see versioning helping a lot there TBH, QA and testing will and those can be done based on branches (and other mechanics) already without versioning.

r7vme: Fully agree with you that "only we are in charge of our own destinies".

I just don't know what and how treats this version folder. Let's may be start with that. If you can point me to the right code, then next time i'll try to bump the version. Or if there is no such code, then it may be good that we started this discussion and it's time to implement one or we even don't need one at all. And these are only my assumptions. But i know nothing about versioning, sorry.

PR #223, opened 2017-10-24
add AlwaysPullImages in api-server admission control

jgsqware:

Setting admission control policy to AlwaysPullImages forces every new pod to pull the
required images every time. In a multitenant cluster users can be assured that their private
images can only be used by those who have the credentials to pull them. Without this
admisssion control policy, once an image has been pulled to a node, any pod from any user
can use it simply by knowing the image’s name, without any authorization check against
the image ownership. When this plug-in is enabled, images are always pulled prior to
starting containers, which means valid credentials are required.

Related to : https://github.com/giantswarm/giantswarm/issues/2028 rule: 1.1.12 Ensure that the admission control policy is set to AlwaysPullImages (Scored) Follow-up: No Impact

### DIFF
--- a/v_0_1_0/master_template.go
+++ b/v_0_1_0/master_template.go
@@ -1850,7 +1850,7 @@ coreos:
       --repair-malformed-updates=false \
       --service-account-lookup=true \
       --authorization-mode=RBAC \
-      --admission-control=NamespaceLifecycle,LimitRanger,ServiceAccount,ResourceQuota,DefaultStorageClass,PodSecurityPolicy \
+      --admission-control=NamespaceLifecycle,LimitRanger,ServiceAccount,ResourceQuota,DefaultStorageClass,PodSecurityPolicy,AlwaysPullImages \
       --cloud-provider={{.Cluster.Kubernetes.CloudProvider}} \
       --service-cluster-ip-range={{.Cluster.Kubernetes.API.ClusterIPRange}} \
       --etcd-servers=https://{{ .Cluster.Etcd.Domain }}:443 \


puja108:
I think we need to discuss the impact of this with customers. This will significantly slow down (re-)scheduling of containers, especially if the containers are big, we are completely disabling the Docker cache here. @othylmann can you confirm with customers that this is what they want? I don't see a big threat in images being available throughout the cluster, for following 2 reasons:
1. Images should not contain any sensitive information (that's kept in secrets)
2. This is on K8s access level, so if you have K8s access and can schedule pods, scheduling pods that are private shouldn't be a big security issue, as they were meant to run in that cluster anyway.

othylmann:
I second @puja108 here if it creates slowness, and we need to explain it and be done with it.

jgsqware:
@puja108 good point. @othylmann should we discussed this week about this with customers to move forward on benchmarks?

othylmann:
We take decisions and explain them @jgsqware no need to discuss with customers. They might have an opinion later but we need to get this freaking done!

jgsqware:
Ok so we can not enbling this, let the cache up for the reason @puja108 explain, and tell the cusytomer why?

othylmann:
whatever makes sense. If you think it is stupid like that then we need to talk about it probably. If you think it is good, then yes, let's do that. You are the expert.

jgsqware:
Ok, as customer ar in Private and trusted environment, I don't see a issue to use local cache. So I will close this, and not enabling it.

PR #224, opened 2017-10-24
Disable anonymous access and insecure port

jgsqware:
Regarding k8s CIS benchmark, we need to disable anonymous access to the api and use 0 as insecure port.

Currently, k8s-addons use kubectl with anonymous access. We have to provide a kubeconfig for this. eg. proxy-kubeconfig

in k8scc the insecure port is set by `--insecure_port={{.Cluster.Kubernetes.API.InsecurePort}}`

**question:** is it a problem for us to disable it? and do you prefer to remove insecure port from struct and use `0` directly in k8scc or having `0` as default value in the struct and let the operator overriding it

Rel: https://github.com/giantswarm/giantswarm/issues/2020

### DIFF
--- a/v_0_1_0/master_template.go
+++ b/v_0_1_0/master_template.go
@@ -1553,28 +1553,28 @@ write_files:
           done
       done
       echo "Addons successfully installed"
-  - path: /etc/kubernetes/config/addons-kubeconfig.yml
-    owner: root
-    permissions: 0644
-    content: |
-      apiVersion: v1
-      kind: Config
-      users:
-      - name: proxy
-        user:
-          client-certificate: /etc/kubernetes/ssl/apiserver-crt.pem
-          client-key: /etc/kubernetes/ssl/apiserver-key.pem
-      clusters:
-      - name: local
-        cluster:
-          certificate-authority: /etc/kubernetes/ssl/apiserver-ca.pem
-          server: https://{{.Cluster.Kubernetes.API.Domain}}
-      contexts:
-      - context:
-          cluster: local
-          user: proxy
-        name: service-account-context
-      current-context: service-account-context
+- path: /etc/kubernetes/config/addons-kubeconfig.yml
+  owner: root
+  permissions: 0644
+  content: |
+    apiVersion: v1
+    kind: Config
+    users:
+    - name: proxy
+      user:
+        client-certificate: /etc/kubernetes/ssl/apiserver-crt.pem
+        client-key: /etc/kubernetes/ssl/apiserver-key.pem
+    clusters:
+    - name: local
+      cluster:
+        certificate-authority: /etc/kubernetes/ssl/apiserver-ca.pem
+        server: https://{{.Cluster.Kubernetes.API.Domain}}
+    contexts:
+    - context:
+        cluster: local
+        user: proxy
+      name: service-account-context
+    current-context: service-account-context
 
 - path: /etc/kubernetes/config/proxy-kubeconfig.yml
   owner: root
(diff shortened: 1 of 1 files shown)


xh3b4sd:
I am not sure about the implications of this. Lets wait for an SREs opinion.

calvix:
`k8s-addons` is one of the tools that uses it. I cant recall any other but fore sure this should be a change that is verified by creating a cluster and then running sonobuy k8s conformance test

calvix:
I dont folow this:

> and do you prefer to remove insecure port from struct and use 0 directly in k8scc or having 0 as default value in the struct and let the operator overriding it

 what is the difference

calvix:
I will be personaly sad since i have been using the insecure port for quick access to k8s, but thats just my bad :)

xh3b4sd:
For on prem I am using the insecure port as well to create a SSH tunnel to forward the port to be able to use kubectl locally. How can we do this in case we close the insecure port?

jgsqware:
For on prem, `gsctl` is not working? because I've tested it for AWS and that's working.

@calvix regarding the question, I asking if we disable it at all by default via `k8scloudconfig` or if we let the `x-operator` to set it to 0 by default but use a variable that could be overwritten, eg for testing env...

calvix:
@jgsqware  Thanks for the clarification. 

`gsctl` is working for me for all on-prems and if necessary all k8s apis are accessible from VPN. ( the only problem can be the strange container dance some ppl do on Anubis, but that is not my case as i don't use it).

calvix:
@jgsqware when this is going to be merged can you please make an announcement (`#news-ops`) that the insecure port is disabled on new installations so the ppl are not confused?

jgsqware:
Currently blocked by https://github.com/giantswarm/giantswarm/issues/2062

v_0_1_0/master_template.go
```diff
@@ -1545,7 +1549,7 @@ write_files:
       for manifest in $MANIFESTS
       do
           while
-              /usr/bin/docker run --net=host --rm -v /srv:/srv $KUBECTL apply -f /srv/$manifest
+              /usr/bin/docker run -e KUBECONFIG=${KUBECONFIG} --net=host --rm -v /srv:/srv /etc/kubernetes:/etc/kubernetes $KUBECTL apply -f /srv/$manifest
```
calvix: you missed `-v` before the `/etc/kubernetes` volume on each line

v_0_1_0/master_template.go
```diff
@@ -1457,21 +1457,27 @@ write_files:
...
 
       /usr/bin/docker pull $KUBECTL
 
+##
+-e KUBECONFIG=${KUBECONFIG} --net=host --rm -v /etc/kubernetes:/etc/kubernetes 
+##
+
       # wait for healthy master
-      while [ "$(/usr/bin/docker run --net=host --rm $KUBECTL get cs | grep Healthy | wc -l)" -ne "3" ]; do sleep 1 && echo 'Waiting for healthy k8s'; done
+      while [ "$(/usr/bin/docker run --e KUBECONFIG=${KUBECONFIG} -net=host --rm /etc/kubernetes:/etc/kubernetes $KUBECTL get cs | grep Healthy | wc -l)" -ne "3" ]; do sleep 1 && echo 'Waiting for healthy k8s'; done
```
r7vme: remove one dash `--e`

calvix: also `-net` needs two dashs

v_0_1_0/master_template.go
jgsqware: done

v_0_1_0/master_template.go
jgsqware: done

PR #225, opened 2017-10-25
enable proxy protocol for ingress-nginx

fgimenez:
Towards https://github.com/giantswarm/giantswarm/issues/1952

### DIFF
--- a/v_0_1_0/master_template.go
+++ b/v_0_1_0/master_template.go
@@ -726,6 +726,7 @@ write_files:
     data:
       server-name-hash-bucket-size: "1024"
       server-name-hash-max-size: "1024"
+      use-proxy-protocol: "true"
 - path: /srv/ingress-controller-dep.yml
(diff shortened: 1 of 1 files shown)


fgimenez:
@r7vme thx, about the implementation in AWS indeed, it is explained in the issue https://github.com/giantswarm/giantswarm/issues/1952, we need to add to the ELB a policy with the `ProxyProtocol` attribute set to `true` if it's not already applied, I'm checking it right now.

rossf7:
This will enable proxy protocol on all clusters. Is this also OK on KVM?

calvix:
I have concerns about KVM, this needs to be properly tested on some of our customers before we can merge.

fgimenez:
Thx @calvix, can i perform this testing?

calvix:
sure

fgimenez:
Tested on lycan

PR #227, opened 2017-11-01
Add iscsi support for guest cluster

corest:
Add required mounts for `k8s-kubelet` to make iscsi mounts working in guest cluster


r7vme:
I'm late, but should not we enable iscsid and multipathd services in kvm-operator only?

Like this https://github.com/giantswarm/kvm-operator/blob/master/service/cloudconfig/v_0_1_0_master_template.go#L161

???

PR #233, opened 2017-11-02
run-calico-controller-on-master

calvix:
this PR moves `calico-node-controller` to master and remove the need of working calico network in order to work.

I have encountered a scenario when there was deadlock - calico-node was not starting because of duplicate ip and calico-node-controller can't clean the config because networking was not working.

This PR solve this issue.

### DIFF
--- a/v_0_1_0/master_template.go
+++ b/v_0_1_0/master_template.go
@@ -117,7 +117,18 @@ write_files:
           namespace: kube-system
           labels:
             k8s-app: calico-node-controller
+          annotations:
+            scheduler.alpha.kubernetes.io/critical-pod: ''
         spec:
+          tolerations:
+          - key: node-role.kubernetes.io/master
+            operator: Exists
+            effect: NoSchedule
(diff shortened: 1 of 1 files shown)


v_0_1_0/master_template.go
```diff
@@ -117,7 +117,18 @@ write_files:
...
+            scheduler.alpha.kubernetes.io/critical-pod: ''
         spec:
+          tolerations:
+          - key: node-role.kubernetes.io/master
+            operator: Exists
+            effect: NoSchedule
+          - key: CriticalAddonsOnly
+            operator: Exists
+          nodeSelector:
+            role: master
```
r7vme: Please use label that used in community (e.g. calico, kubeadm, etc.)

https://github.com/giantswarm/etcd-backup/blob/master/helm/etcd-backup-chart/templates/cronjob.yaml#L28-L33

calvix:
@r7vme fixed

PR #234, opened 2017-11-02
Move kube proxy to self-hosted

calvix:
towards: https://github.com/giantswarm/giantswarm/issues/1973

* moved k8s-proxy to daemon set
* added service account for proxy

One of the reason is to have some health check on k8s-proxy


calvix:
ttested

PR #237, opened 2017-11-08
add SecurityContextDeny to AdmissionControl

jgsqware:
Regarding this: 1.1.14 Ensure that the admission control policy is set to SecurityContextDeny

As far as we already have RBAC and PSP, we can disable SecurityContext specific

### DIFF
--- a/v_0_1_0/master_template.go
+++ b/v_0_1_0/master_template.go
@@ -2017,7 +2017,7 @@ coreos:
       --repair-malformed-updates=false \
       --service-account-lookup=true \
       --authorization-mode=RBAC \
-      --admission-control=NamespaceLifecycle,LimitRanger,ServiceAccount,ResourceQuota,DefaultStorageClass,PodSecurityPolicy \
+      --admission-control=NamespaceLifecycle,LimitRanger,ServiceAccount,ResourceQuota,DefaultStorageClass,PodSecurityPolicy,SecurityContextDeny \
       --cloud-provider={{.Cluster.Kubernetes.CloudProvider}} \
       --service-cluster-ip-range={{.Cluster.Kubernetes.API.ClusterIPRange}} \
       --etcd-servers=https://127.0.0.1:2379 \


jgsqware:
@puja108 Regarding PSP

I've created a kube config like this:

```
gsctl --endpoint https://api.g8s.gauss.eu-central-1.aws.gigantic.io create kubeconfig --certificate-organizations='system:masters' -c mz4hc
```

I create a busybox like this:

```yaml
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  annotations:
    deployment.kubernetes.io/revision: "1"
  creationTimestamp: null
  generation: 1
  labels:
    run: busybox
  name: busybox
  selfLink: /apis/extensions/v1beta1/namespaces/default/deployments/busybox
spec:
  replicas: 1
  selector:
    matchLabels:
      run: busybox
  strategy:
    rollingUpdate:
      maxSurge: 1
      maxUnavailable: 1
    type: RollingUpdate
  template:
    metadata:
      creationTimestamp: null
      labels:
        run: busybox
    spec:
      containers:
      - args:
        - sh
        image: busybox
        imagePullPolicy: Always
        name: busybox
        stdin: true
        securityContext:
          privileged: true
```

And I received this in replicatSet:

```
Error creating: pods "busybox-64fccc5b55-" is forbidden: unable to validate ag
ainst any pod security policy: [spec.containers[0].securityContext.privileged: Invalid value: true: Privileged containers are not allowed
```

When I try to validate my link to `system:master` I got this:
```
➔ kubectl auth can-i get pods --as-group=system:masters  
error: requesting groups or user-extra for giantswarm-mz4hc-user without impersonating a user
```

Do you have an idea?

jgsqware:
From where `system:masters` come from? I don't see it anywhere?

jgsqware:
`SecurityContextDeny` cannot be set since it will disallow PSP.

PR #238, opened 2017-11-08
add NodeRestriction to AdmissionControl

jgsqware:
Signed-off-by: Julien Garcia Gonzalez <julien@giantswarm.io>NodeRestriction

Related to: https://github.com/giantswarm/giantswarm/issues/2033

### DIFF
--- a/v_0_1_0/master_template.go
+++ b/v_0_1_0/master_template.go
@@ -2017,7 +2017,7 @@ coreos:
       --repair-malformed-updates=false \
       --service-account-lookup=true \
       --authorization-mode=RBAC \
-      --admission-control=NamespaceLifecycle,LimitRanger,ServiceAccount,ResourceQuota,DefaultStorageClass,PodSecurityPolicy \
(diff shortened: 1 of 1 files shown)


puja108:
I think this only works together with NodeAuthorization if the nodes have certificates with O `system:nodes` and CN `system:node:<nodeName>` like explained here: https://kubernetes.io/docs/admin/admission-controllers/#noderestriction

jgsqware:
Ok so we have to have a different certificate for this?

puja108:
Yes, we need 1 identifying cert per worker, so we cannot reuse between workers then -> bigger issue this!

xh3b4sd:
Note that there are new versions being introduced and your changes here have to be aligned accordingly.

xh3b4sd:
What is with this PR? I assume we can close and delete the branch. Please sort this.

jgsqware:
I'll close this until we work on this https://github.com/giantswarm/giantswarm/issues/3374

PR #239, opened 2017-11-09
Upgrade ingress controller

fgimenez:
Towards https://github.com/giantswarm/giantswarm/issues/2024

Adds the changes suggested in https://github.com/giantswarm/giantswarm/issues/1875

### DIFF
--- a/v_0_1_0/master_template.go
+++ b/v_0_1_0/master_template.go
@@ -1730,7 +1730,7 @@ coreos:
 
       [Service]
       Type=oneshot
-      ExecStartPre=/bin/bash -c "gpasswd -d core rkt; gpasswd -d core docker; gpasswd -d core wheel"
+      ExecStartPre=-/bin/bash -c "gpasswd -d core rkt; gpasswd -d core docker; gpasswd -d core wheel"
       ExecStartPre=/bin/bash -c "until [ -f '/etc/sysctl.d/hardening.conf' ]; do echo Waiting for sysctl file; sleep 1s;done;"
       ExecStart=/usr/sbin/sysctl -p /etc/sysctl.d/hardening.conf
 
--- a/v_0_1_0/master_template.go
+++ b/v_0_1_0/master_template.go
@@ -855,7 +855,7 @@ write_files:
           serviceAccountName: nginx-ingress-controller
           containers:
           - name: nginx-ingress-controller
-            image: quay.io/kubernetes-ingress-controller/nginx-ingress-controller:0.9.0-beta.16
+            image: quay.io/kubernetes-ingress-controller/nginx-ingress-controller:0.9.0-beta.17
             args:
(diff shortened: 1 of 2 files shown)


v_0_1_0/master_template.go
```diff
@@ -833,7 +833,8 @@ write_files:
       strategy:
         type: RollingUpdate
         rollingUpdate:
-          maxUnavailable: 1
+          maxSurge: 1
+          maxUnavailable: 0
```
calvix: what means maxUnavaiable 0? does that work?
we have nginx controller limited to 1 per machine  and there can be created another pod of we dont remove the previous

v_0_1_0/master_template.go
```diff
@@ -880,6 +882,12 @@ write_files:
                 scheme: HTTP
               initialDelaySeconds: 10
               timeoutSeconds: 1
+            lifecycle:
+              preStop:
+                exec:
+                  command:
+                  - sleep
+                  - "15"
```
calvix: why do we wait 15 sec before deletion?

calvix: sorry read the issue and found the reason for this

v_0_1_0/master_template.go
fgimenez: @calvix this comes too from the recommended settings in the linked issues, I've checked it and seems to work, deleted a ic pod, it moves to terminating and a new one is created while the previous one is still there, eventually the old one is gone and then the new one moves from pending to running, is that ok?

fgimenez:
Tested on lycan, normal ingress, tls termination and ssl passthrough working.

v_0_1_0/master_template.go
puja108: as we don't have hard-anti-affitini it should work

fgimenez: yep, also these settings make more sense imo than maxUnavailable=1 (given that only one runs on each host), they try make sure that there will be always (at most) one pending while any of them is terminating.

puja108: ack

calvix: we have hard affinity afaik

fgimenez: @calvix you are right, after trying a rolling update changing the image the new pod was getting stuck on pending with:
```
Events:
  FirstSeen    LastSeen    Count    From            SubObjectPath    Type        Reason            Message
  ---------    --------    -----    ----            -------------    --------    ------            -------
  3m        15s        15    default-scheduler            Warning        FailedScheduling    No nodes are available that match all of the predicates: PodFitsHostPorts (3), PodToleratesNodeTaints (1).
```
and the update didn't progress. With `maxUnavailable` set to 1 the rolling update works fine, already pushed this change.

PR #240, opened 2017-11-14
reserve-resources-for-critical-components

calvix:
towards https://github.com/giantswarm/giantswarm/issues/2112

set reserved resource for:
* os system - 0.15CPU 250MiB ram
* kubelet - 0.15 CPU 250MiB ram
* kube-proxy - 0.075CPU 80MiB ram
* IC controller - 0.5CPU 350MiB ram
* calico kube controllers - 0.03CPU 90MiB ram

components that already have reserved resources:
* calico node - 0.25 CPU
* kube-dns - 0.1CPU 70MiB ram + 0.15CPU 10MiB + 0.01CPU + 20MiB ram
* default-backend - 0.01CPU 20MiB ram

Lets have a discussion here about the limits. Daniel from adidas said that he set reservation for IC to 1CPU and 500 MiB ram, but that seems to be way to muuch if we consider how big default machines are.

Lets do some math:
CPU: on each node: 0.15+0.15+0.075+0.5+0.25 = 1.125CPU (m3.large has 2 CPUs)
RAM: on each node: 250+250+80+350 = 920 MiB (m3.large has 8GB of ram )
+  another 3x(0.1+0.15+0.01) +2xx0.01+0.03= 0.83 CPU spread accross  worker nodes
+ some small extra ram usage.

This seems to be hard to setup by one config. As for big workloads IC limits may need to be bigger but if we set it globally, our reservation will eat like 75% of CPU resources on cloud (if we use m3.large)

cc @r7vme @teemow @othylmann @corest @fgimenez

### DIFF
--- a/v_0_1_0/master_template.go
+++ b/v_0_1_0/master_template.go
@@ -507,6 +507,10 @@ write_files:
                 # kubernetes.default to the correct service clusterIP.
                 - name: CONFIGURE_ETC_HOSTS
                   value: "true"
+              resources:
+                requests:
+                  cpu: 30m
+                  memory: 90Mi
               volumeMounts:
                 # Mount in the etcd TLS secrets.
                 - mountPath: /etc/kubernetes/ssl/etcd
@@ -875,6 +879,10 @@ write_files:
                 path: /healthz
                 port: 10254
                 scheme: HTTP
+            resources:
+              requests:
+                memory: "350Mi"
+                cpu: "500m"
             livenessProbe:
               httpGet:
                 path: /healthz
@@ -983,6 +991,10 @@ write_files:
                   port: 10256
                 initialDelaySeconds: 10
                 periodSeconds: 3
+              resources:
+                requests:
+                  memory: "125Mi"
+                  cpu: "75m"
               securityContext:
                 privileged: true
               volumeMounts:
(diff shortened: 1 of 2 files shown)


teemow:
kube-dns, kube-proxy, calico-node and calico-kube-controller are all set as critical-pod. we should do the same with the ingress controllers imo.

calvix:
added

teemow:
regarding exact numbers of the reservations i'd say we can only find proper defaults if we do some more loadtesting. and then still those numbers might vary depending on the workloads. 

having a minimum as a default and then manually reconfigure production clusters is fine with me for now.

r7vme:
Looks ok for me, hope it won't break under the load :)

othylmann:
The customer thought setting this to burstable instead of guaranteed would be enough. then the cpu limits seem to be fine.

calvix:
By the definition of burstable o k8s docs https://kubernetes.io/docs/tasks/configure-pod-container/quality-service-pod/#create-a-pod-that-gets-assigned-a-qos-class-of-burstable we are good, as we only set requests and not limits.

deichten:
Ok from my side as defaulta for new clusters. Just as a side note: the numbers currently configured (1 core, 512M mem) are being done on what we saw as peak load before.

2nd note: our former default instance on aws was r3.xlarge and going forward should be m4.2xlarge.

PR #243, opened 2017-11-15
backwards-compatibility-for-kube-proxy

calvix:
fix for https://github.com/giantswarm/giantswarm/issues/2122

There is situation where the new worker is spawned but master still doesn't run the new `kube-proxy` as daemon set that will cause that worker is without proxy.

This hack puts `k8s-proxy` unit back but its disabled by default. If after 2 mins after the kubelet start there are no `kube-proxy` pod running on the machine it will start the old k8s-proxy via systemd.

### DIFF
--- a/v_0_1_0/worker_template.go
+++ b/v_0_1_0/worker_template.go
@@ -331,6 +331,7 @@ coreos:
   - name: k8s-proxy-watcher.service
     enable: true
     command: start
+    content: |
       [Unit]
       Description=k8s-proxy-watcher
       Wants=k8s-kubelet.service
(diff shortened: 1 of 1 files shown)


v_0_1_0/worker_template.go
```diff
@@ -295,6 +295,53 @@ coreos:
...
+      --kubeconfig=/etc/kubernetes/config/proxy-kubeconfig.yml \
+      --conntrack-max-per-core 131072 \
+      --v=2"
+      ExecStop=-/usr/bin/docker stop -t 10 $NAME
+      ExecStopPost=-/usr/bin/docker rm -f $NAME
+  - name: k8s-proxy-watcher.service
+    enable: true
+    command: start
+    [Unit]
+      Description=k8s-proxy-watcher
```
r7vme: please remove two indentation whitespaces

calvix:
I tested manually on KVM and works nicely

PR #245, opened 2017-11-21
Removes old node-exporter service

JosephSalisbury:
Towards https://github.com/giantswarm/giantswarm/issues/1482

We'll bring this back eventually running on kubernetes, but I thought i'd tidy it up as I was here anyway.


kopiczko:
For my interest: where it now runs? Were there 2 node-exporters running?

JosephSalisbury:
This is a hold-over from fleet-k8s days, that we never used in G8S.
We will node-exporter (running in guest cluster) in the near-future, I thought I'd just get rid of the old one pre-emptively (i meant to do this months ago :D )

PR #247, opened 2017-11-22
Remove kubeproxy unit

rossf7:
Removes kubeproxy unit and watcher added in #243 as it causes problems on AWS.

Clusters on AWS take longer to start due to nodes booting and ELBs. The timer means a second kube-proxy is created but the ingress ports are already bound so they fail.

Is this change needed for AWS? If not I'll add the units to kvm-operator so they can be embedded in the cloud config.

### DIFF
--- a/v_0_1_0/worker_template.go
+++ b/v_0_1_0/worker_template.go
@@ -295,54 +295,6 @@ coreos:
       --v=2"
       ExecStop=-/usr/bin/docker stop -t 10 $NAME
       ExecStopPost=-/usr/bin/docker rm -f $NAME
-  - name: k8s-proxy.service
-    enable: false
-    command: stop
-    content: |
-      [Unit]
-      Description=k8s-proxy
-      StartLimitIntervalSec=0
-
-      [Service]
-      Restart=always
-      RestartSec=0
-      TimeoutStopSec=10
-      EnvironmentFile=/etc/network-environment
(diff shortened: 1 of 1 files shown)


rossf7:
Ack. Pawel I've added you as a reviewer. The 2 minute timer isn't going to work on AWS but yes needs SRE input on which platforms this is needed on.

r7vme:
I think we are more or less safe to remove this.

- centaur migrated to new clusters (which have daemonset)
- anubis also migrated to new clusters, but this is not 100% what teams using what.

If it breaks new aws clusters this is definitely should be merged.

Next steps:
1) check how many clusters on-prem still use systemd units
2) either implement this in kvm-operator or manually add daemon set to these clusters (which is more reasonable imo)

r7vme:
cc: @calvix @corest

r7vme:
sre ops task https://github.com/giantswarm/giantswarm/issues/2164

rossf7:
Thanks @r7vme and @kopiczko going to merge this. I'll update the issue giantswarm/giantswarm#2122 with the problems we had on AWS.

calvix:
this was mostly issue on some kvm (but that is gone since we migrated to enw clusters) but also for some aws clusters(icc consult for example)

As most of the clusters that were affected are not used anymore we should be fine.

PR #248, opened 2017-11-24
Fix typo for insecure port

jgsqware:


xh3b4sd:
Is there actually a reason to keep the merged branches alive? Our practise so far was to delete them and make sure the branch list stays clean.

puja108:
Agree. I also always delete mine

Puja
________________________________

PR #249, opened 2017-11-27
fix etcd data creation

jgsqware:

Wait for etcd data directory creation before setting permission

### DIFF
--- a/v_0_1_0/master_template.go
+++ b/v_0_1_0/master_template.go
@@ -1796,7 +1796,7 @@ coreos:
       RemainAfterExit=yes
       TimeoutStartSec=0
       ExecStartPre=/usr/bin/mkdir -p /etc/kubernetes/data/etcd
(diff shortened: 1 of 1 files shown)


calvix:
is this tested? in my head it's looks wrong

xh3b4sd:
What exactly looks wrong to you?

calvix:
The logic that is used there.
We need to basically wait for the mount of etcd-data in kvm.
Waiting for the folder is not enough as the folder will be always there (because it was created by the mkdir in command above).

We could add some `drop-in` service unit into kvm-operator which will add dependency for etcd change owner ship unit (for example `After=`) on the mount unit.

jgsqware:
@calvix WDYT? tested and it works. waiting this to merge those https://github.com/giantswarm/kvm-operator/pull/273

calvix:
I don't think this change is really necessary.

for aws it works without this (because the folder is created by the command one line above) and on KVM we wait until the mount is finished so we are fine.

jgsqware:
ok Closing this

PR #253, opened 2017-11-30
Backwards compatibility for kube proxy updated

calvix:
improving my previous PR with waiting for healthy api.
I wanted to copy behavior of k8s-addons and use kubectl to do healthcheck but this needs extra RBAC permission and i dont wanna mess with that.

This should have almost same functionality. if k8s-api is avaiabel then k8s-kubelet is able to spawn kube-proxy pods as its daemon set with host network so no calico is needed for that.


calvix:
tested on both aws and KVM and works good

xh3b4sd:
I saw this got merged. I created a couple of versioned packages where these changes should probably go into as well. How do we streamline this? Do you want to sort it? Can I help?

calvix:
yeah i just asked about that in slack

PR #254, opened 2017-12-01
copied v_0_1_0 package

xh3b4sd:
The CRD migration requires us to use the generated types of our CRD clients. There we have a new common ground which is used to render templates. Thus we need to introduce a new version of the templates in order to be able to use the new types. This here is just copy/paste of the existing templates. No changes are applied. I will provide another PR for the real changes. That way we see better what is going on. 

Note that I chose a simpler versioning since I think the folder naming how we had it was weird. We will not expose our internal versioning so it does not has to be semver. We can simply increment single numbers to distinguish different template versions.

### DIFF
--- a/v2/cloudconfig.go
+++ b/v2/cloudconfig.go
@@ -1,0 +1,64 @@
+package v_0_1_0
+
+import (
+	"bytes"
+	"compress/gzip"
+	"encoding/base64"
+	"text/template"
+
+	"github.com/giantswarm/clustertpr"
+	"github.com/giantswarm/clustertpr/spec"
+)
+
+type Params struct {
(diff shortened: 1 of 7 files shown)


xh3b4sd:
I would say this is a freeze for `v_0_1_0` and further changes should be made in `v2`. Existing PRs should be migrated. I can help with this if needed.

xh3b4sd:
Going with this to provide the PR with the real changes.

xh3b4sd:
See the follow up here: https://github.com/giantswarm/k8scloudconfig/pull/255.

xh3b4sd:
FYI @calvix @r7vme

PR #255, opened 2017-12-01
added changes for v2 package

xh3b4sd:
Follow up of https://github.com/giantswarm/k8scloudconfig/pull/254.


glide.yaml
```diff
@@ -1,5 +1,7 @@
 package: github.com/giantswarm/k8scloudconfig
 import:
+- package: github.com/giantswarm/apiextensions
+  version: clientset
```
xh3b4sd: This is the branch we have to pin for the migration. After we run on CRDs I will switch this to master.

template.go
```diff
@@ -3,13 +3,15 @@ package cloudconfig
 import (
 	"github.com/giantswarm/microerror"
 
+	"github.com/giantswarm/k8scloudconfig/v2"
```
xh3b4sd: There is the new `v2` package now.

template_test.go
```diff
@@ -23,15 +27,15 @@ func Test_Template(t *testing.T) {
 		template, err := NewTemplate(tc.Version)
 		if err != nil {
 			if tc.ErrorMatcher != nil && !tc.ErrorMatcher(err) {
-				t.Fatalf("test %d expected %#v got %#v", i+1, true, false)
+				t.Fatalf("test %d expected %#v got %#v", i, true, false)
```
xh3b4sd: Boyscouting for simplicity.

v2/cloudconfig.go
```diff
@@ -1,19 +1,18 @@
-package v_0_1_0
+package v2
```
xh3b4sd: These are the actual changes of the copied package. Now it is really v2.

v2/cloudconfig.go
```diff
@@ -1,19 +1,18 @@
...
 	"text/template"
 
-	"github.com/giantswarm/clustertpr"
-	"github.com/giantswarm/clustertpr/spec"
+	"github.com/giantswarm/apiextensions/pkg/apis/cluster/v1alpha1"
 )
 
 type Params struct {
-	Cluster   clustertpr.Spec
+	Cluster   v1alpha1.Cluster
```
xh3b4sd: We use the types of our generated clients here. They do not differ in structure so we can simply keep the templates as they are.

glide.yaml
```diff
@@ -1,5 +1,13 @@
...
+- package: k8s.io/apiextensions-apiserver
+  version: release-1.7
+- package: k8s.io/apimachinery
+  version: release-1.7
+- package: k8s.io/client-go
+  version: v4.0.0
```
xh3b4sd: This has to be pinned like we do in operatorkit because of the Kubernetes magic with interfaces and code generation. Once we are running on CRDs I will remove this.

PR #257, opened 2017-12-04
added changes for v1 package

xh3b4sd:
Follow up of https://github.com/giantswarm/k8scloudconfig/pull/256.


template.go
```diff
@@ -3,13 +3,17 @@ package cloudconfig
...
 	"github.com/giantswarm/k8scloudconfig/v_0_1_0"
 )
 
 type version string
 
 const (
 	V_0_1_0 version = "v_0_1_0"
+	// V1 is only used for KVM and has the experimental encryption feature
+	// disabled. See also https://github.com/giantswarm/k8scloudconfig/pull/257.
+	V1 version = "v1"
```
xh3b4sd: We can now use the master branch in the kvm-operator again.

jgsqware:
Basically with tis, we dont touch v1 for new stuff and only applied on v2 ?

xh3b4sd:
I see it like this. 

- There is a version `v_0_1_0` which we currently use everywhere. 
- There is a version `v1` which we can use to get the vendor pinning in KVM straight. 
- There is a version `v2` which is for the general TPR migration. Note this is not usable for KVM because it has the encryption feature enabled. 
- There has to be a version just like `v2` but with encryption disabled. We can call this `v1_1` or whatever.

PR #260, opened 2017-12-04
use semver versioning scheme

xh3b4sd:
I introduced a simplified versioning scheme which we do not want. This PR moves the existing packages to the same semver approach.


v_1_1_0/cloudconfig.go
```diff
@@ -1,4 +1,4 @@
-package v2
+package v_1_1_0
```
xh3b4sd: The diff is weird because I changed all versions in the same commit. I try to do better in the future.

PR #263, opened 2017-12-14
v_3_0_0: unify CloudConfig construction

kopiczko:


v_3_0_0/cloudconfig.go
```diff
@@ -21,14 +28,21 @@ type CloudConfig struct {
...
 		config:   "",
-		params:   params,
-		template: template,
+		params:   config.Params,
+		template: config.Template,
 	}
 
-	return newCloudConfig, nil
+	return c, nil
 }
```
kopiczko: This file changed. Other that that it's moving stuff around.

v_3_0_0/cloudconfig.go
```diff
@@ -21,14 +28,21 @@ type CloudConfig struct {
 	template string
 }
 
-func NewCloudConfig(template string, params Params) (*CloudConfig, error) {
-	newCloudConfig := &CloudConfig{
+func New(config Config) (*CloudConfig, error) {
```
xh3b4sd: I think the naming was right. `v_3_0_0.New()` is not really descriptive. `v_3_0_0.NewCloudConfig()` is though.

kopiczko: I thought it'll be aliased anyway to something like k8scloudconfigv3.

Should Config be CloudConfigConfig in that case?

xh3b4sd: Don't know. As I said somewhere yesterday, I think when we have to alias packages, the package name is just wrong and we should rather that than being "forced" to alias the import.

xh3b4sd: IMO `cloudconfigv3` as package name would work, but then we miss the semver semnatics, wich might be unnecessary anyway.

kopiczko: I think wen package indicates version we're kind of forced to alias them.

kopiczko: I changed that but I still think CloudConfigConfig and DefaultCloudConfigConfig is awkwar.

PR #265, opened 2017-12-14
v_3_0_0: allow setting custom K8s API address for master nodes

kopiczko:
In Azure load balancers backend pool nodes are not allowed to reach the load
balancer. Because of that we have to be able to specify different (loopback)
K8s apiserver address on master nodes.

### DIFF
--- a/v_3_0_0/cloudconfig.go
+++ b/v_3_0_0/cloudconfig.go
@@ -36,6 +36,14 @@ func NewCloudConfig(config CloudConfigConfig) (*CloudConfig, error) {
 		return nil, microerror.Maskf(invalidConfigError, "config.Template must not be empty")
 	}
 
+	// Set default params.
+	if config.Params.MasterAPIDomain == "" {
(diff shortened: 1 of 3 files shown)


v_3_0_0/cloudconfig.go
```diff
@@ -36,6 +36,14 @@ func NewCloudConfig(config CloudConfigConfig) (*CloudConfig, error) {
 		return nil, microerror.Maskf(invalidConfigError, "config.Template must not be empty")
 	}
 
+	// Set default params.
+	if config.Params.MasterAPIDomain == "" {
+		config.Params.MasterAPIDomain = config.Params.Cluster.Kubernetes.API.Domain
+	}
+	if config.Params.Hyperkube.Apiserver.BindAddress == "" {
+		config.Params.Hyperkube.Apiserver.BindAddress = defaultHyperkubeApiserverBindAddress
+	}
```
kopiczko: @calvix I'm also worried about backward compatibility. Those lines fallback to the old values when nothing is set.

PR #267, opened 2017-12-15
v_3_0_0: allow disabling calico

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/1384

This is hopefully the last change in cloud configs. It is also the ugliest.

### DIFF
--- a/v_3_0_0/master_template.go
+++ b/v_3_0_0/master_template.go
@@ -10,6 +10,7 @@ users:
        - "{{ $user.PublicKey }}"
 {{end}}
 write_files:
+{{ if not .DisableCalico -}}
 - path: /srv/calico-ipip-pinger-sa.yaml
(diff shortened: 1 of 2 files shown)


v_3_0_0/master_template.go
```diff
@@ -1544,6 +1548,8 @@ write_files:
           sleep 3s
       done
 
+      {{ end -}}
+
```
kopiczko: How templating works: https://play.golang.org/p/5jX2s35OQi

puja108:
@xh3b4sd agree and we are working on the spec for that, it's a bit more complicated as we have different "classes" of addons that need to be added at certain times during or after the cluster bootstrap (e.g. networking needs to be there before dns, and after dns everything is optional or you might even say dns is optional)

PR #269, opened 2017-12-19
add encryption template

jgsqware:

Related to this https://github.com/giantswarm/kvm-operator/pull/320

### DIFF
--- a/v_2_0_0/cloudconfig_test.go
+++ b/v_2_0_0/cloudconfig_test.go
@@ -80,7 +80,7 @@ func TestCloudConfigTemplating(t *testing.T) {
 						CIDR:   24,
 					},
 				},
-				Extension: &FakeExtension{Keys: FakeKeys{APIServerEncryptionKey: []byte("test")}},
+				Extension: &FakeExtension{},
 			},
 
 			expectedString: `
--- a/v_2_0_0/cloudconfig_test.go
+++ b/v_2_0_0/cloudconfig_test.go
@@ -80,7 +80,7 @@ func TestCloudConfigTemplating(t *testing.T) {
 						CIDR:   24,
 					},
 				},
-				Extension: &FakeExtension{},
+				Extension: &FakeExtension{Keys: FakeKeys{APIServerEncryptionKey: []byte("test")}},
 			},
 
 			expectedString: `
@@ -99,7 +99,6 @@ func TestCloudConfigTemplating(t *testing.T) {
 		if err := cloudConfig.ExecuteTemplate(); err != nil {
 			t.Fatalf("%v: unexpected error templating cloud config: %v", index, err)
 		}
-
 		if !strings.Contains(cloudConfig.String(), test.expectedString) {
 			t.Fatalf("%v: expected string not found in cloud config: %v", index, test.expectedString)
 		}
--- a/v_2_0_0/master_template.go
+++ b/v_2_0_0/master_template.go
@@ -1735,22 +1735,6 @@ write_files:
     [Service]
     ExecStartPre=/bin/bash -c "while [ ! -f /etc/audit/rules.d/10-docker.rules ]; do echo 'Waiting for /etc/audit/rules.d/10-docker.rules to be written' && sleep 1; done"
 
-- path: /etc/kubernetes/encryption/k8s-encryption-config.yaml
-  owner: root
-  permissions: 600
-  content: |
-    kind: EncryptionConfig
-    apiVersion: v1
-    resources:
-      - resources:
-        - secrets
-        providers:
-        - aescbc:
-            keys:
-            - name: key1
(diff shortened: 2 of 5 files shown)


v_2_0_0/master_template.go
```diff
@@ -1735,6 +1735,22 @@ write_files:
...
+    resources:
+      - resources:
+        - secrets
+        providers:
+        - aescbc:
+            keys:
+            - name: key1
+              secret: {{.Extension.Keys.APIServerEncryptionKey | printf "%s" }}
+        - identity: {}
+
```
xh3b4sd: Why does this change here? v2 is already in use in AWS AFAIK. So it means it must not change anymore. IMO v3 is the place to get changes in.

jgsqware: ok lets move it to v3 so

jgsqware: done

v_2_0_0/cloudconfig_test.go
```diff
@@ -99,7 +99,6 @@ func TestCloudConfigTemplating(t *testing.T) {
 		if err := cloudConfig.ExecuteTemplate(); err != nil {
 			t.Fatalf("%v: unexpected error templating cloud config: %v", index, err)
 		}
-
```
xh3b4sd: That does not change anything for the guest clusters but in the future please lets make sure we do not touch "frozen" versions at all.

v_3_0_0/fake_extension.go
```diff
@@ -1,6 +1,11 @@
 package v_3_0_0
 
-type FakeExtension struct{}
+type FakeKeys struct {
+	APIServerEncryptionKey []byte
+}
+type FakeExtension struct {
+	Keys FakeKeys
+}
```
kopiczko: I think it should be added to Params struct.

v_3_0_0/master_template.go
```diff
@@ -1751,6 +1751,22 @@ write_files:
...
+    kind: EncryptionConfig
+    apiVersion: v1
+    resources:
+      - resources:
+        - secrets
+        providers:
+        - aescbc:
+            keys:
+            - name: key1
+              secret: {{.Extension.Keys.APIServerEncryptionKey | printf "%s" }}
```
kopiczko: Where does that come from? I don't see a field added to Params struct.

v_3_0_0/fake_extension.go
jgsqware: It's part of Extension, but the extension will be overwritten

jgsqware: Do you prefer to have it, as new Param field?

kopiczko: FakeExtensions is only for testing purposes. Extension interface is here https://github.com/giantswarm/k8scloudconfig/blob/d75265767b2a885d6cfb004ff00ab700908b8d03/v_3_0_0/types.go#L95 and I don't see a reason why we would add a new field (or rather method - it's an interface) to it.

v_2_0_0/master_template.go
kopiczko: I see this first comment here. And this comment https://github.com/giantswarm/kvm-operator/pull/320#discussion_r157724047

It sounds to me like dev versions and snapshots of them is a good idea. Otherwise we'll have hard times which versions we "can" change.

xh3b4sd: For me the "dev" version is always the latest one. Once it is frozen a new one has to be established so people do not have to think. This is why it does not matter how we name the folder.

kopiczko: Let's say we have version `v_2_0_0` and `v_3_0_0` and we want to add a feature to `v_2_0_0`. This means we need to create a version `v_2_1_0` but we also have `v_3_0_0`. How to recognize that we can work on `v_2_1_0`?

xh3b4sd: Why should we work in both? Doesn't sound right to me. There are multiple versions and one is "dev". That is it. You describe a scenario where there are two dev versions. This does not work IMO. When there is 2 and 3 it means two is frozen and 3 is dev. We never look back. We always only move forward.

kopiczko: What if a customer is using `v_2_0_0` which runs kubernetes 1.8. And our latest (i.e. dev) is `v_3_0_0` which runs kubernetes 1.9. We noticed serious bug in kubernetes 1.8 and we have to update to 1.8.1. But for valid reasons customer can't switch to kubernetes 1.9 ATM. What should we do?

xh3b4sd: Then we introduce a patch and freeze it immediately. 2.1 is then the patch and is not touched anymore after creation. 3 still remains the dev playground.

kopiczko:
Closing in favour of https://github.com/giantswarm/k8scloudconfig/pull/277.

PR #272, opened 2017-12-21
v_3_0_0: make etcd port configurable

rossf7:
Towards giantswarm/giantswarm#2023

Make EtcdPort configurable but default to port 443 for existing platforms.

For AWS we will set this to port 2379. This will let us replace the Etcd internal ELB with a CNAME record to the master node.


rossf7:
Good idea. I've renamed the PR for the commit message.

PR #275, opened 2017-12-21
v3_0_0 and v2_0_0. Upgrade calico to 2.6.4

r7vme:
Upgrades Calico to 2.6.4 and remove unnecessary services.

Tested with kvm-operator. Aws-operator in progress.

Fixes: https://github.com/giantswarm/giantswarm/issues/2185

### DIFF
--- a/v_2_0_0/master_template.go
+++ b/v_2_0_0/master_template.go
@@ -10,167 +10,6 @@ users:
        - "{{ $user.PublicKey }}"
 {{end}}
 write_files:
-- path: /srv/calico-ipip-pinger-sa.yaml
-  owner: root
-  permissions: 644
-  content: |
-    apiVersion: v1
-    kind: ServiceAccount
-    metadata:
-      name: calico-ipip-pinger
-      namespace: kube-system
-- path: /srv/calico-ipip-pinger.yaml
-  owner: root
-  permissions: 644
-  content: |
-    apiVersion: extensions/v1beta1
-    kind: DaemonSet
-    metadata:
-      labels:
-        app: calico-ipip-pinger
-      name: calico-ipip-pinger
-      namespace: kube-system
-    spec:
-      updateStrategy:
-        type: RollingUpdate
-        rollingUpdate:
-          maxUnavailable: 1
-      template:
-        metadata:
-          labels:
-            app: calico-ipip-pinger
-        spec:
-          serviceAccountName: calico-ipip-pinger
-          containers:
-          - name: calico-ipip-pinger
-            image: quay.io/giantswarm/calico-ipip-pinger:0f197abe866c7c811e42534739148345cb3bded5
-            imagePullPolicy: Always
-            securityContext:
-              privileged: true
-            env:
-              # The location of the Calico etcd cluster.
-              - name: ETCD_ENDPOINTS
-                valueFrom:
-                  configMapKeyRef:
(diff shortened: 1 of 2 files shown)


v_2_0_0/master_template.go
```diff
@@ -10,167 +10,6 @@ users:
        - "{{ $user.PublicKey }}"
 {{end}}
 write_files:
-- path: /srv/calico-ipip-pinger-sa.yaml
```
xh3b4sd: We should not modify versions that are already established. I had the hope Joe was the exception because of monitoring. When we want to patch a specific version we should create a new version based of it. I think in this case it might be enough to just put it into v3.

v_2_0_0/master_template.go
```diff
@@ -1508,11 +1300,7 @@ write_files:
        calico-node-sa.yaml\
        calico-kube-controllers-sa.yaml\
        calico-ds.yaml\
-       calico-kube-controllers.yaml\
-       calico-node-controller-sa.yaml\
-       calico-node-controller.yaml\
-       calico-ipip-pinger-sa.yaml\
-       calico-ipip-pinger.yaml"
+       calico-kube-controllers.yaml"
```
calvix: i think we should add commands for deleting the old removed resources

r7vme: This makes sense.

v_2_0_0/master_template.go
r7vme: Agree to bump versions. But in general, this is important bugfix release. We need to ship Calico 2.6.4 to existing clusters.

If i still want to  do changes in v2_x_x. Can you please describe steps i need to do (i think you're the only SME about versions at the moment)?

1) `cp -r v2_0_0 v2_0_1`
2) Apply my changes to v2_0_1
3) ...

xh3b4sd: How do we want to ship this to existing clusters? Are you updating all of them manually somehow? How did we do this before? We never updated anything automatically. Being able to do so is still WIP. AWS is not reconciling and KVM has updates disabled. So even if we modify existing versions we do not really apply them. You would need to roll everything manually somehow when this is so important. 

I am against modifying existing versions that are already in use. Don't know how often I have to say this still. I am wondering why everything is now so important that we have to bypass the process we want to establish. 

You can create a new version but then customers have to create new clusters to get this very important bugfix.

r7vme: I did not know that we disabled automatic updates in kvm-operator (and in aws-operator), so i expected that existing versions will be rolled out once vm recreated.

Can we spec our upgrade/versioning strategy at least in common sentences? It's total miscommunication on such important topic. And looks like it's our common fault.

Can we have a call on that? may be within the sig syncup?

cc: @calvix @teemow @puja108

kopiczko: I'd like to split the PR in to two: one with with v_2_0_1; one with update to v_3_0_0.
I agree with Tim. we should never change existing version. Even if it's blank line removal. If we want to roll patches automatically we can implement that in one of our services like cluster-service or kubernetesd.

r7vme:
Closing in favor  https://github.com/giantswarm/k8scloudconfig/pull/276 (only for v3_0_0)

For v2_0_0 it's not clear what i can do here. I raised my questions in Slack, let's discuss there.

PR #276, opened 2017-12-27
v_3_0_0: Update Calico to 3.0.1

r7vme:
Upgrade Calico to 3.0.1. Tested on geckon.

Partially fixes: https://github.com/giantswarm/giantswarm/issues/2185

### DIFF
--- a/v_3_0_0/master_template.go
+++ b/v_3_0_0/master_template.go
@@ -11,167 +11,6 @@ users:
 {{end}}
 write_files:
 {{ if not .DisableCalico -}}
-- path: /srv/calico-ipip-pinger-sa.yaml
-  owner: root
-  permissions: 644
-  content: |
-    apiVersion: v1
-    kind: ServiceAccount
-    metadata:
-      name: calico-ipip-pinger
-      namespace: kube-system
-- path: /srv/calico-ipip-pinger.yaml
-  owner: root
-  permissions: 644
-  content: |
-    apiVersion: extensions/v1beta1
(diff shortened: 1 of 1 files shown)


v_3_0_0/master_template.go
```diff
@@ -215,31 +54,41 @@ write_files:
...
+            {
+              "type": "portmap",
+              "snat": true,
+              "capabilities": {"portMappings": true}
             }
+          ]
         }
 
+      # If you're using TLS enabled etcd uncomment the following.
+      # You must also populate the Secret below with these files.
```
kopiczko: What to uncomment? Everything is uncommented.

v_3_0_0/master_template.go
```diff
@@ -11,167 +11,6 @@ users:
...
-                  value: "https://kubernetes.default:443"
-              volumeMounts:
-                # Mount in the etcd TLS secrets.
-                - mountPath: /etc/kubernetes/ssl/etcd
-                  name: etcd-certs
-          volumes:
-            # Mount in the etcd TLS secrets.
-            - name: etcd-certs
-              hostPath:
-                path: /etc/kubernetes/ssl/etcd
```
kopiczko: It feels good we can get rid of this stuff.

v_3_0_0/master_template.go
r7vme: These are just default calico manifests (even comments). I'm trying to keep them as default as possible.

PR #281, opened 2018-01-04
v_3_0_0: Replace kube-dns with CoreDNS

r7vme:
tested in `geckon`

Fixes https://github.com/giantswarm/giantswarm/issues/2135

### DIFF
--- a/v_3_0_0/master_template.go
+++ b/v_3_0_0/master_template.go
@@ -360,142 +360,106 @@ write_files:
               hostPath:
                 path: /etc/kubernetes/ssl/etcd
 {{ end -}}
-- path: /srv/kubedns-cm.yaml
-  owner: root
-  permissions: 0644
-  content: |
-    apiVersion: v1
-    kind: ConfigMap
-    metadata:
-      name: kube-dns
-      namespace: kube-system
-      labels:
-        addonmanager.kubernetes.io/mode: EnsureExists
-- path: /srv/kubedns-sa.yaml
+- path: /srv/coredns.yaml
   owner: root
   permissions: 0644
   content: |
     apiVersion: v1
     kind: ServiceAccount
     metadata:
-      name: kube-dns
+      name: coredns
       namespace: kube-system
+    ---
+    apiVersion: rbac.authorization.k8s.io/v1beta1
+    kind: ClusterRole
+    metadata:
       labels:
-        kubernetes.io/cluster-service: "true"
-        addonmanager.kubernetes.io/mode: Reconcile
-- path: /srv/kubedns-dep.yaml
-  owner: root
-  permissions: 0644
-  content: |
-    apiVersion:  extensions/v1beta1
+        kubernetes.io/bootstrapping: rbac-defaults
+      name: system:coredns
+    rules:
+    - apiGroups:
+      - ""
+      resources:
+      - endpoints
+      - services
+      - pods
+      - namespaces
+      verbs:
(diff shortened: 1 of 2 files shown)


v_3_0_0/master_template.go
```diff
@@ -360,210 +360,143 @@ write_files:
...
+      name: coredns
       namespace: kube-system
       labels:
-        k8s-app: kube-dns
-        kubernetes.io/cluster-service: "true"
-        addonmanager.kubernetes.io/mode: Reconcile
+        k8s-app: coredns
+        kubernetes.io/name: "CoreDNS"
     spec:
+      replicas: 2
```
puja108: why only 2 and not 3? or is this how it's done upstream?

v_3_0_0/master_template.go
```diff
@@ -360,210 +360,143 @@ write_files:
...
   content: |
     apiVersion: v1
-    kind: ConfigMap
+    kind: ServiceAccount
     metadata:
-      name: kube-dns
+      name: coredns
       namespace: kube-system
+    ---
+    apiVersion: rbac.authorization.k8s.io/v1beta1
```
puja108: we should start using `v1` instead of `v1beta1`, which will get deprecated soon

v_3_0_0/master_template.go
```diff
@@ -360,210 +360,143 @@ write_files:
...
+      resources:
+      - endpoints
+      - services
+      - pods
+      - namespaces
+      verbs:
+      - list
+      - watch
+    ---
+    apiVersion: rbac.authorization.k8s.io/v1beta1
```
puja108: s.a.

v_3_0_0/master_template.go
r7vme: this it pure upstream manifest https://github.com/coredns/deployment/blob/master/kubernetes/coredns.yaml.sed

v_3_0_0/master_template.go
r7vme: hmm, in upstream they still on v1beta

r7vme: do you know when it became `v1` ? may be they keep beta for compatibility with old versions?

r7vme: anyway will switch to `v1`

puja108: it's v1 since 1.8 and we're on 1.9 already in k8scc.v3 so we're safe to update

v_3_0_0/master_template.go
puja108: nice, all good then, we will need to see how it scales with lots of requests, but I heard CoreDNS is quite performant and nice

r7vme:
@puja108 addressed `v1betav1` comments.

kopiczko:
As always it feels good to remove something from cloud configs.

r7vme:
Tested with kvm-operator.

```
Ran 125 of 709 Specs in 3223.580 seconds
SUCCESS! -- 125 Passed | 0 Failed | 0 Pending | 584 Skipped PASS
```

puja108:
@r7vme do you still have that cluster? can you send me the results (e2e.log and junit_01.xml) and a output of `kubectl version`?

r7vme:
already deleted :/ I can recreate and test one more time if you want? let me know.

puja108:
yeah, that would be great, we need the files and the output of version command to do PR upstream

PR #282, opened 2018-01-10
v_2_0_0: Fix kubernetes audit logging

r7vme:
One more exception to v2_0_0 as this is postmortem.

- Enable legacy audit logging. (It was disabled in 1.8 by default)
- Fix mounts for audit log
- Extend audit log size.

Part of https://github.com/giantswarm/giantswarm/issues/2346

Regarding size:
Audit log size is drowing REALLY FAST (1G of logs is 1.5 days in host cluster). My change limits maximum size to 3G (total we have ~6Gb root on master). Unfortunately we can not use logrotate compression, because of this [upstream bug](https://github.com/kubernetes/kubernetes/issues/52865). Anyway this is temporary solution. I hope central logging will come soon.

Change to v_3_0_0 with proper audit which is beta in 1.9 (not legacy) is coming.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ version directory, and then changes are introduced.
 
 ### Changed
 - Updated calico to 2.6.5.
+- Fix kubernetes audit logging
 
 ### Removed
 - Removed calico-ipip-pinger.
--- a/v_2_0_0/master_template.go
+++ b/v_2_0_0/master_template.go
@@ -1883,6 +1883,7 @@ coreos:
       -v /etc/kubernetes/ssl/:/etc/kubernetes/ssl/ \
       -v /etc/kubernetes/secrets/token_sign_key.pem:/etc/kubernetes/secrets/token_sign_key.pem \
       -v /etc/kubernetes/encryption/:/etc/kubernetes/encryption \
+      -v /var/log:/var/log \
       $IMAGE \
       /hyperkube apiserver \
       --allow_privileged=true \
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -38,6 +38,7 @@ version directory, and then changes are introduced.
 
 ### Changed
 - Updated calico to 2.6.5.
+- Fix kubernetes audit logging
```
kopiczko: Full stop 🚓

xh3b4sd:
Configmaps are reconciled regardless. Deployments are not updated automatically for KVM. Basic updates are only enabled on geckon for testing reasons. When pods restart updates are applied though, which is why we are going to have snowflakes if we do not roll the guest clusters manually.

kopiczko:
@r7vme: according to what @xh3b4sd just said I'm not sure if this won't be less work to make it a proper bugfix version with version bundles and stuff.

General thought: we need more tooling. My thinking is some basic code generation.

xh3b4sd:
> I'm not sure if this won't be less work to make it a proper bugfix

This is what I am saying all the time. You can try to patch it that way but it does not really work. And that it does not really work is not because of version bundles or anything but just the way how our product ever worked and still works. We are working on the automated solution and now what we always had becomes a problem. I don't understand this. This is just hjacking the process for no (for me perceivable) benefit. 

> we need more tooling. My thinking is some basic code generation.

What we need IMO is getting the updates implemented, which is what we are working on. What tooling do you have in mind? How could codegeneration help here?

kopiczko:
@r7vme I think we can close that, correct?

r7vme:
yes, closed.

PR #284, opened 2018-01-11
Copy v2_0_0 to v2_0_1

r7vme:
No real changes included.

Following process as described here:
https://github.com/giantswarm/giantswarm/issues/2335#issuecomment-356927360

### DIFF
--- a/v_2_0_1/cloudconfig.go
+++ b/v_2_0_1/cloudconfig.go
@@ -1,0 +1,63 @@
+package v_2_0_1
+
+import (
+	"bytes"
+	"compress/gzip"
+	"encoding/base64"
+	"text/template"
+
+	"github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1"
+)
+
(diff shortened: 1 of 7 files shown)


v_2_0_1/cloudconfig.go
```diff
@@ -0,0 +1,63 @@
+package v_2_0_1
```
xh3b4sd: This is not just copy/paste. You edited the files and at this point nobody knows what you changed elsewhere. The copy PR is just about `cp src dst`. Nothing else. Then you can just merge right away like vendor PRs. Package name changes and template changes should then be done in the follow up PR.

kopiczko: I'd be ok with package name change only. That declutters changes diff a little bit.

xh3b4sd: Just saying, when you state a PR to be copy paste but it is not only copy paste trust is gone.

PR #285, opened 2018-01-11
V201 fix audit

r7vme:
Enable legacy audit logging. (It was disabled in 1.8 by default)
Fix mounts for audit log
Extend audit log size.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,11 @@ version directory, and then changes are introduced.
 - Remove calico-ipip-pinger.
 - Remove calico-node-controller.
 
+## [v2.0.1]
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -31,6 +31,11 @@ version directory, and then changes are introduced.
 - Remove calico-ipip-pinger.
 - Remove calico-node-controller.
 
+## [v2.0.1]
+
+### Changed
+- Fix audit logging.
```
xh3b4sd: The description you provide in this PR is way nicer than this here. I would like to see the more detailed version here. Fixing the audit log can mean anything.

PR #288, opened 2018-01-23
v_3_0_0: update nginx-ingress-controller container to 0.10.1

corest:
Towards: https://github.com/giantswarm/giantswarm/issues/2374

### DIFF
--- a/v_3_0_0/master_template.go
+++ b/v_3_0_0/master_template.go
@@ -656,7 +656,7 @@ write_files:
               privileged: true
           containers:
           - name: nginx-ingress-controller
-            image: quay.io/giantswarm/nginx-ingress-controller-0.10.2
+            image: quay.io/giantswarm/nginx-ingress-controller:0.10.2
             args:
             - /nginx-ingress-controller
             - --default-backend-service=$(POD_NAMESPACE)/default-http-backend
--- a/v_3_0_0/master_template.go
+++ b/v_3_0_0/master_template.go
@@ -646,7 +646,7 @@ write_files:
           serviceAccountName: nginx-ingress-controller
           containers:
           - name: nginx-ingress-controller
-            image: quay.io/giantswarm/nginx-ingress-controller-0.10.0
+            image: quay.io/giantswarm/nginx-ingress-controller-0.10.1
             args:
             - /nginx-ingress-controller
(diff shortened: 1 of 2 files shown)


puja108:
We need to be careful about breaking changes between our beta and this version. I remember they included a change from the old annotation to a new one, which can be reverted by setting some flags.

We need to decide if we want to tell users to move all their annotations (if they have them) to the new one, or we set the IC to have old annotation style. @othylmann (and customer team, e.g. @pipo02mix) WDYT?

corest:
@puja108 I went through the changelog and don't see any changes except prometheus annotations. Did you mean those?

puja108:
their on the 0.9 final release: https://github.com/kubernetes/ingress-nginx/releases/tag/nginx-0.9.0

puja108:
I talked to @marians and we think that we should go with upstream default and include the new annotation in our changelog and announcement for v3. This way if a customer in the future looks at upstream nginx docs they won't be confused because of different annotation prefix.

/cc @teemow @othylmann

pipo02mix:
Then in case we want to upgrade an existing cluster (IC component to latest version too) we need to encourage the customer to change all annotations, right?

And it would provoke downtime IIRC

marians:
It should be possible to avoid downtime (mostly) by first adding the annotations in the new form before the upgrade, then do the IC upgrade, then remove the old annotations.

v_3_0_0/master_template.go
```diff
@@ -644,14 +644,25 @@ write_files:
                         - nginx-ingress-controller
                   topologyKey: kubernetes.io/hostname
           serviceAccountName: nginx-ingress-controller
+          initContainers:
+          - command:
+            - sh
+            - -c
+            - sysctl -w net.core.somaxconn=32768; sysctl -w net.ipv4.ip_local_port_range="1024 65535"
```
corest: @r7vme I remember you did some manual changes for ingress-controller. Were those somehow related to the queue for connections waiting for acceptance?

corest:
tested last coredns/nginx ingress controller. Works fine, didn't find any issues

corest:
Towards https://github.com/giantswarm/giantswarm/issues/2374

PR #289, opened 2018-01-23
Removes apiserver encryption key validation, as it chokes worker clou…

JosephSalisbury:
…d configs


JosephSalisbury:
This was my thinking.Thanks guys!

PR #295, opened 2018-01-24
v_3_0_0: update coredns from 1.0.1 to 1.0.4

corest:


corest:
Tested on gauss, all good

v_3_0_0/master_template.go
```diff
@@ -452,9 +452,21 @@ write_files:
               effect: NoSchedule
             - key: "CriticalAddonsOnly"
               operator: "Exists"
+          affinity:
+            podAntiAffinity:
+              preferredDuringSchedulingIgnoredDuringExecution:
```
kopiczko: Why not required? Does it make sense to run 2 coredns on the same machine?

corest: Can’t say for now. I tested antiaffinity with required and it fails to run even one pod. 
Soft requirement used also in nginx ic

kopiczko: Ack.

puja108: It does make sense even on the same machine as they cover load no matter on what machine. It would be best to have them on more machines for failover reasons.

PR #296, opened 2018-01-28
v3_1_0: Move k8s components to selfhosted pods

corest:
Towards: https://github.com/giantswarm/giantswarm/issues/2374

### DIFF
--- a/v_3_1_0/master_template.go
+++ b/v_3_1_0/master_template.go
@@ -1782,6 +1782,11 @@ write_files:
       containers:
       - name: k8s-api-server
         image: quay.io/giantswarm/hyperkube:v1.9.2
+        env:
+        - name: HOST_IP
+          valueFrom:
+            fieldRef:
+              fieldPath: status.podIP
         command:
         - /hyperkube
         - apiserver
@@ -1795,7 +1800,7 @@ write_files:
         - --kubelet_https=true
         - --kubelet-preferred-address-types=InternalIP
         - --secure_port={{.Cluster.Kubernetes.API.SecurePort}}
-        - --bind-address={{.Hyperkube.Apiserver.BindAddress}}
+        - --bind-address=$(HOST_IP)
         - --etcd-prefix={{.Cluster.Etcd.Prefix}}
         - --profiling=false
         - --repair-malformed-updates=false
@@ -1808,7 +1813,7 @@ write_files:
         - --etcd-cafile=/etc/kubernetes/ssl/etcd/server-ca.pem
         - --etcd-certfile=/etc/kubernetes/ssl/etcd/server-crt.pem
         - --etcd-keyfile=/etc/kubernetes/ssl/etcd/server-key.pem
-        - --advertise-address={{.Hyperkube.Apiserver.BindAddress}}
+        - --advertise-address=$(HOST_IP)
         - --runtime-config=api/all=true
         - --logtostderr=true
         - --tls-cert-file=/etc/kubernetes/ssl/apiserver-crt.pem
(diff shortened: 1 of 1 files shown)


v_3_1_0/master_template.go
```diff
@@ -1768,6 +1768,211 @@ write_files:
...
+    metadata:
+      name: k8s-api-server
+      namespace: kube-system
+    spec:
+      hostNetwork: true
+      containers:
+      - name: k8s-api-server
+        image: quay.io/giantswarm/hyperkube:v1.9.2
+        env:
+        - name: HOST_IP
```
corest: Difference from systemd unit, where we used `/etc/network-environment` for getting env variable with host ip address.

v_3_1_0/master_template.go
```diff
@@ -1768,6 +1768,211 @@ write_files:
...
+        - --audit-log-maxbackup=30
+        - --audit-log-maxsize=100
+        - --audit-policy-file=/etc/kubernetes/manifests/audit-policy.yml
+        - --experimental-encryption-provider-config=/etc/kubernetes/encryption/k8s-encryption-config.yaml
+        resources:
+          requests:
+            cpu: 300m
+        livenessProbe:
+          tcpSocket:
+            port: {{.Cluster.Kubernetes.API.SecurePort}}
```
corest: Here is the tricky thing. 
I tried httpGet with `HTTPS` scheme, which fails with code `401`, which is actually healthy status code from the health perspective. 
TCP check seems enough for API liveness probe.

corest:
Tested on gauss, got working guest cluster with self-hosted k8s components:
```
calico-kube-controllers-7c57ddf866-6425z                            1/1       Running   0          2m
calico-node-6s8fg                                                   2/2       Running   0          2m
calico-node-qmjnq                                                   2/2       Running   0          2m
coredns-84cd68c555-cnkl6                                            1/1       Running   0          2m
coredns-84cd68c555-jzwkv                                            1/1       Running   0          2m
default-http-backend-6d65799844-5nj8x                               1/1       Running   0          2m
default-http-backend-6d65799844-752zz                               1/1       Running   0          2m
k8s-api-server-ip-10-1-2-64.eu-central-1.compute.internal           1/1       Running   0          2m
k8s-controller-manager-ip-10-1-2-64.eu-central-1.compute.internal   1/1       Running   0          1m
k8s-scheduler-ip-10-1-2-64.eu-central-1.compute.internal            1/1       Running   0          1m
kube-proxy-6s66b                                                    1/1       Running   0          2m
kube-proxy-zm6nx                                                    1/1       Running   0          2m
kube-state-metrics-6b6bbbb948-r7m9t                                 1/1       Running   0          2m
nginx-ingress-controller-6f749fb46d-tps9w                           1/1       Running   0          2m
node-exporter-6rsnf                                                 1/1       Running   0          2m
node-exporter-7pd5c                                                 1/1       Running   0          2m
```

v_3_1_0/master_template.go
```diff
@@ -2114,155 +2321,14 @@ coreos:
...
   - name: k8s-addons.service
     enable: true
     command: start
     content: |
       [Unit]
       Description=Kubernetes Addons
-      Wants=k8s-api-server.service
-      After=k8s-api-server.service
+      Wants=k8s-kubelet.service
+      After=k8s-kubelet.service
```
puja108: I guess this works without race condition even when it takes a while for kubelet to start api server?

corest: `k8s-addons` uses those checks in systemd only for proper order, real dependency is inside the script. So nothing happens before cluster is healthy

v_3_1_0/master_template.go
r7vme: ack.

v_3_1_0/master_template.go
r7vme: even `/healthz` gives you 401?

i see that kubeadm uses https https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/phases/controlplane/manifests.go#L84

corest: I tested that path. Gives me `401`  as well

PR #297, opened 2018-01-30
v_3_1_0: init from v_3_0_0

corest:
This PR is the base for future PR with actual changes, related to migration from systemd units to selfhosted k8s components


corest:
@puja108 is that right, that with selfhosted components we are moving to new minor release?

PR #299, opened 2018-02-02
v3_1_0: fix coredns service name

corest:


r7vme:
JFYI: this was inherited from official manifests.

https://github.com/coredns/deployment/blob/master/kubernetes/coredns.yaml.sed#L133-L141

i hope there are no dependencies on kube-dns service name.

PR #300, opened 2018-02-08
Add WIP version 3.1.0 to changelog

corest:


corest:
OIDC is implemented on operator side and uses k8s-api-server extra arguments, which were introduced in 3.0.0.

PR #303, opened 2018-02-13
New WIP patch version v_3_1_1

r7vme:
I see potential in doing patch changes in batches.

This change just adds "dev"/WIP version for future patches.

First candidate is new Calico 3.0.2 cc: @calvix

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ The latest version is considered WIP and it is a subject of change. All other
 versions are frozen. To freeze current version all files are copied to a new
 version directory, and then changes are introduced.
 
+## [v3.1.1] WIP
+
+TBD
+
 ## [v3.1.0]
 
 ### Changed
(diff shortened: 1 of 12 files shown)


corest:
Not sure about this approach. We have this copy magic as one step just to make PRs cleaner. Now this one, which actually doesn't bring any change.

kopiczko:
Doesn't bring any change, but makes futher PRs way cleaner.

corest:
We can just put bumping version into the first commit and then bring some changes with next commits to see why this PR was created

kopiczko:
To make the actual changes PR cleaner. This isn't maybe a representative example. Look at https://github.com/giantswarm/kvm-operator/commit/96d7bc329c4752e6d2d044fd4ab17058d418c1d9 & https://github.com/giantswarm/kvm-operator/commit/783d200305854779024c1b7d4ef73f0844dd96e2

I prefer having this changes separated. Otherwise it's very difficult to review.

calvix:
i think we should automatize this in script or something,   In few months there will be 10 steps that need to be done and so much thinking around this.

xh3b4sd:
There is already a script for the operators and really decent documentation. I suggest you use and read it. Then we can talk about further improvements.

PR #305, opened 2018-02-15
Fix api-server bind address

corest:
Previous change broke possibility for templating api-server bind address. This feature is required for `azure-operator` and blocks release of the new bundle.

### DIFF
--- a/v_3_1_1/types.go
+++ b/v_3_1_1/types.go
@@ -5,6 +5,8 @@ import (
 )
 
 const (
+	// Default value references environment variable,
+	// defined inside api-server pod definition.
 	defaultHyperkubeApiserverBindAddress = "$(HOST_IP)"
 )
 
--- a/v_3_1_1/types.go
+++ b/v_3_1_1/types.go
@@ -5,7 +5,7 @@ import (
 )
 
 const (
-	defaultHyperkubeApiserverBindAddress = "${DEFAULT_IPV4}"
+	defaultHyperkubeApiserverBindAddress = "$(HOST_IP)"
 )
 
 type Params struct {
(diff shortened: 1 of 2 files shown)


v_3_1_1/types.go
```diff
@@ -5,7 +5,7 @@ import (
 )
 
 const (
-	defaultHyperkubeApiserverBindAddress = "${DEFAULT_IPV4}"
+	defaultHyperkubeApiserverBindAddress = "$(HOST_IP)"
```
kopiczko: Unrelated: What is `$(HOST_IP)`? What are parenthesis doing?

corest: It is env variable, which comes from pod.IP https://github.com/giantswarm/k8scloudconfig/blob/e0bb4328626571fb87adfa9ddb6380b9eb1fc63e/v_3_1_1/master_template.go#L1792
Parenthesis is the way to access env vars as arguments for container in pod

kopiczko: Why it isn't `${HOST_IP}`?

corest: Because it is the way to distinguish way of accessing env vars. You can't pass host env vars with that simple reference. If you could - all those changes would be useless. 
Here is the example: https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#use-environment-variables-to-define-arguments

kopiczko: Thanks for explaining that.

r7vme: please add the same explanation to this constant. very specific things.

I had exactly the same questions as Pawel :)

PR #308, opened 2018-02-15
Add HostMount extension

corest:
With selfhosted components, we need another way to add custom mounts. This is how I see we can do it. Naming is not that god, but for now I'm just testing stuff here: https://github.com/giantswarm/azure-operator/pull/122 . With this configuration I'm able to run k8s components as selfhosted. There are still some other issues, so this is only a part of a change

### DIFF
--- a/v_3_1_1/types.go
+++ b/v_3_1_1/types.go
@@ -83,11 +83,11 @@ type HyperkubeDocker struct {
 }
 
 type HyperkubePod struct {
-	PodExtraMounts   []PodMount
-	CommandExtraArgs []string
+	HyperkubePodHostExtraMounts []HyperkubePodHostMount
+	CommandExtraArgs            []string
 }
 
-type PodMount struct {
+type HyperkubePodHostMount struct {
 	Name     string
 	Path     string
 	ReadOnly bool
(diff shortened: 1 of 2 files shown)


v_3_1_1/types.go
```diff
@@ -66,11 +66,11 @@ type HyperkubeApiserver struct {
 	//
 	// azure-operator sets that to 0.0.0.0. Other operators leave it empty.
 	BindAddress string
-	Docker      HyperkubeDocker
```
corest: This one become useless for ApiServe rand ControllerManager as they don't need docker extra arguments, but instead - extra mounts

v_3_1_1/types.go
```diff
@@ -82,6 +82,17 @@ type HyperkubeDocker struct {
 	CommandExtraArgs []string
 }
 
+type HyperkubePod struct {
+	HostExtraMounts  []HostMount
+	CommandExtraArgs []string
+}
+
+type HostMount struct {
```
kopiczko: I would like to have types prefixed with their "parent" types. Something like `HyperkubePodHostMount` in this case. I know the name is ugly. But when you remove "parent" type you know what "child" types should removed so we don't have garbage left.

PR #310, opened 2018-02-21
update-etcd-3.3.1

calvix:
* update etcd to latest version


calvix:
1. WIP k8s cloud-config should not be vendored into `not` WIP version in operators. If its WIP in azure then its fine, if not then someone made a mistake.  

2.  nope, no changes needed, we have this tested on host cluster aready

PR #314, opened 2018-02-21
Kube proxy fix

rossf7:
Towards giantswarm/giantswarm#2598

Adds v_2_0_2 with the fix to start kube-proxy if it is missing.


v_2_0_2/worker_template.go
```diff
@@ -295,6 +295,57 @@ coreos:
...
+        
+        [Service]
+        Environment="KUBECONFIG=/etc/kubernetes/config/kubelet-kubeconfig.yml"
+        Environment="KUBECTL=quay.io/giantswarm/docker-kubectl:1dc536ec6dc4597ba46769b3d5d6ce53a7e62038"
+        Type=oneshot
+        ExecStartPre=/bin/sh -c "while ! /usr/bin/docker run -e KUBECONFIG=${KUBECONFIG} --net=host --rm -v /etc/kubernetes:/etc/kubernetes $KUBECTL get no 2>/dev/null 1>/dev/null; do sleep 1 && echo 'Waiting for healthy k8s'; done"
+        ExecStartPre=/bin/sh -c "sleep 2m"
+        ExecStart=/bin/sh -c "proxyCount=$(docker ps | grep 'kube-proxy' | wc -l);if [ "$proxyCount" == "0" ]; then systemctl start k8s-proxy; fi;"
+        ExecStartPost=/bin/sh -c "echo Done."
+        RemainAfterExit=yes
```
rossf7: This is from v_0_1_0 I just updated the version to k8s 1.8.4.

v_2_0_2/worker_template.go
```diff
@@ -295,6 +295,57 @@ coreos:
       --v=2"
       ExecStop=-/usr/bin/docker stop -t 10 $NAME
       ExecStopPost=-/usr/bin/docker rm -f $NAME
+      - name: k8s-proxy.service
```
calvix: this looks like bad indentation

rossf7: Good catch, thx! Fixed.

v_2_0_2/worker_template.go
```diff
@@ -295,6 +295,57 @@ coreos:
       --v=2"
       ExecStop=-/usr/bin/docker stop -t 10 $NAME
       ExecStopPost=-/usr/bin/docker rm -f $NAME
+    - name: k8s-proxy.service
```
calvix: i still think its wrong as kubelet that is above has only 2 spaces before the `- name:` 
so the whole block might be shifted by 2 spaces

rossf7: @calvix Should be good now. PTAL.

calvix: I thnik we still need to remove 2 space from `    - name: k8s-proxy.service` and `    - name: k8s-proxy-watcher.service` and we should be good :)

rossf7: Damnit, sorry this took a few goes :( Done.

calvix:
:+1:

PR #316, opened 2018-02-22
Adapt kubelet flags to protect kube resources

pipo02mix:
Towards https://github.com/giantswarm/giantswarm/issues/2523

### DIFF
--- a/v_3_1_1/master_template.go
+++ b/v_3_1_1/master_template.go
@@ -2333,9 +2333,8 @@ coreos:
       --pod-manifest-path=/etc/kubernetes/manifests \
       --kubeconfig=/etc/kubernetes/config/kubelet-kubeconfig.yml \
       --node-labels="node-role.kubernetes.io/master,role=master,kubernetes.io/hostname=${HOSTNAME},ip=${DEFAULT_IPV4},{{.Cluster.Kubernetes.Kubelet.Labels}}" \
-      --kube-reserved="cpu=500m,memory=250Mi" \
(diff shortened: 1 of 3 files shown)


corest:
Can you also delete WIP from changelog after this is done

v_3_1_1/master_template.go
```diff
@@ -2402,8 +2402,9 @@ coreos:
       --pod-manifest-path=/etc/kubernetes/manifests \
       --kubeconfig=/etc/kubernetes/config/kubelet-kubeconfig.yml \
       --node-labels="node-role.kubernetes.io/master,role=master,kubernetes.io/hostname=${HOSTNAME},ip=${DEFAULT_IPV4},{{.Cluster.Kubernetes.Kubelet.Labels}}" \
-      --kube-reserved="cpu=150m,memory=250Mi" \
+      --kube-reserved="cpu=200m,memory=250Mi" \
       --system-reserved="cpu=150m,memory=250Mi" \
```
pipo02mix: Finally, I leave 200m for container runtime and kubelet after checking how other open tools (kops, rancher, openshift) set this value.

v_3_1_1/master_template.go
```diff
@@ -2402,8 +2402,9 @@ coreos:
...
+      --enforce-node-allocatable=pods \
       --v=2"
```
pipo02mix: With this value no kubelet will consider `Allocatable` as capacity for the node and it will evict pods in case it overpass it

PR #322, opened 2018-03-02
Enable aggregation layer for k8s api server

pipo02mix:
Towards https://github.com/giantswarm/giantswarm/issues/2656


pipo02mix:
Tested PTAL

PR #323, opened 2018-03-05
v_3_2_2: copy v_3_2_1

kopiczko:


kopiczko:
I wait for approval here, because I don't know if 3.2.2 makes sense or this should be other version

xh3b4sd:
No strong feelings here. I still think the templates could simply be versioned by single integers. Don't know if the semver format was ever useful here.

kopiczko:
OK. Merging as it is. This will be WIP so we can rename or replace with another. This is just an indication 3_2_1 is now frozen.

PR #324, opened 2018-03-05
v_3_2_2: update imports

kopiczko:


kopiczko:
Next PR. It is broken in a few places.

PR #330, opened 2018-03-15
v_3_2_3: remove hard limits from core cluster components

corest:
Towards https://github.com/giantswarm/giantswarm/issues/2751


corest:
This will arrive no earlier than Monday because I also want to test rate limiting for api

PR #334, opened 2018-03-22
v_3_2_4: Workaround for systemd-networkd behavior

r7vme:
Fixes: https://github.com/giantswarm/giantswarm/issues/2862

See bug: https://github.com/systemd/systemd/issues/8524

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ The latest version is considered WIP and it is a subject of change. All other
 versions are frozen. To freeze current version all files are copied to a new
 version directory, and then changes are introduced.
 
+## [v3.2.4] WIP
+
+### Changed
+- Masked systemd-networkd-wait-online unit.
+
 ## [v3.2.3]
 
 ### Changed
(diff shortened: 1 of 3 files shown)


v_3_2_4/worker_template.go
```diff
@@ -200,8 +200,7 @@ coreos:
     command: stop
     mask: true
   - name: systemd-networkd-wait-online.service
-    enable: true
-    command: start
+    mask: true
```
kopiczko: Doesn't it mean it will be basically disabled?
What does this unit do?

r7vme: this unit checks that all interfaces in `networkctl` are in status `configured` in some random cases, when interface comes up w/o ipv6 link-local interface, it will be in `configuring` state and this systemd service will fail after timeout.

See here: https://github.com/giantswarm/giantswarm/issues/2553
I've also created an upstream bug: https://github.com/systemd/systemd/issues/8524

CHANGELOG.md
```diff
@@ -9,6 +9,11 @@ The latest version is considered WIP and it is a subject of change. All other
 versions are frozen. To freeze current version all files are copied to a new
 version directory, and then changes are introduced.
 
+## [v3.2.4] WIP
```
corest: I think we agreed to not mark cloudconfig as WIP because it should be ready for release in any time.

corest: There is no way. Everything should be released every time.

PR #335, opened 2018-03-22
removes injection of unencrypted Kubernetes encryption key

xh3b4sd:
This is properly tested on ginger via https://github.com/giantswarm/aws-operator/pull/870.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ version directory, and then changes are introduced.
 
 ### Changed
 - Masked systemd-networkd-wait-online unit.
+- Removed injection of unencrypted Kubernetes encryption key.
 
 ## [v3.2.3]
 
@@ -119,6 +120,10 @@ version directory, and then changes are introduced.
 
 ## [v0.1.0]
 
+
+
+[v3.2.4]: https://github.com/giantswarm/k8scloudconfig/commits/master/v_3_2_4
+[v3.2.3]: https://github.com/giantswarm/k8scloudconfig/commits/master/v_3_2_3
 [v3.2.2]: https://github.com/giantswarm/k8scloudconfig/commits/master/v_3_2_2
 [v3.2.1]: https://github.com/giantswarm/k8scloudconfig/commits/master/v_3_2_1
 [v3.1.1]: https://github.com/giantswarm/k8scloudconfig/commits/master/v_3_1_1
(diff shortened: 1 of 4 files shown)


v_3_2_4/master_template.go
```diff
@@ -1837,22 +1837,6 @@ write_files:
...
-    apiVersion: v1
-    resources:
-      - resources:
-        - secrets
-        providers:
-        - aescbc:
-            keys:
-            - name: key1
-              secret: {{ .ApiserverEncryptionKey }}
-        - identity: {}
```
kopiczko: We want that for azure and kvm, don't we?

xh3b4sd: Oh shit. I have to check the other operators how this is supposed to be. My thinking is that they have to inject it on their own. The template change here is fine since this should not be there because the management is different across operators. I need to get into KVM and Azure though. Maybe you can point out some things for Azure?

kopiczko: In general we have `ExtraManifests` which you can set in both azure and kvm operators https://github.com/giantswarm/k8scloudconfig/blob/392865b96d922615f7b4b946962ffd13e18a9983/v_3_2_4/types.go#L30

But I'd be more keen on adding `DisableEnscryptionAtREST` like we do with calico to keep manifests in sync https://github.com/giantswarm/k8scloudconfig/blob/392865b96d922615f7b4b946962ffd13e18a9983/v_3_2_4/types.go#L14

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ version directory, and then changes are introduced.
 
 ### Changed
 - Masked systemd-networkd-wait-online unit.
+- Removed injection of unencrypted Kubernetes encryption key.
```
kopiczko: Changelog entry is stale.

v_3_2_4/master_template.go
xh3b4sd: I did a thing. Please have another look. Currently working on the operators. Will ping there after testing confirmation.

CHANGELOG.md
xh3b4sd: Good catch. Done.

xh3b4sd:
Sorted operators for KVM, AWS and Azure. Tested on all installations. Rolling now.

PR #342, opened 2018-04-03
Changes required for new docker

r7vme:
fixes: https://github.com/giantswarm/giantswarm/issues/2913

In new docker `17.12.1` (in latest stable CoreOS):
- was completely removed flag `--disable-legacy-registry=true`
- removed support for `shared` mount mode, instead `rshared` should be used (details wtf it means [here](https://github.com/giantswarm/giantnetes-terraform/pull/56)).


r7vme:
This change is compatible with older docker so all good, but yeah we need new CoreOS in general

PR #346, opened 2018-04-13
v_3_2_5: switch to hyperkube v1.10.1

kopiczko:


v_3_2_5/master_template.go
```diff
@@ -1827,13 +1827,13 @@ write_files:
...
-        - --insecure_bind_address=0.0.0.0
+        - --allow-privileged=true
+        - --insecure-bind-address=0.0.0.0
         - --anonymous-auth=false
         - --insecure-port=0
-        - --kubelet_https=true
+        - --kubelet-https=true
         - --kubelet-preferred-address-types=InternalIP
-        - --secure_port={{.Cluster.Kubernetes.API.SecurePort}}
+        - --secure-port={{.Cluster.Kubernetes.API.SecurePort}}
```
kopiczko: Those are deprecated we should replace them with some sort of config.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ version directory, and then changes are introduced.
 ## [v3.2.5] WIP
 
 ### Changed
+- Update hyperkube 1.10.1.
```
MarcelMue: Should this be `Updated` for consistency? Everything else is `Changed`, `Disabled`

kopiczko: Actually past tense is inconsistent.

puja108: Do we say hyperkube or do we just default to kubernetes, cause hyperkube might be confusing (it's an implementation detail) or we mention both?

v_3_2_5/master_template.go
puja108: right kubelet is now dynamically configurable (you can even hotswap configuration IIRC) but it needs to be a config file I think

kopiczko:
Tested on everything in the world.

CHANGELOG.md
kopiczko: This is k8scloudconfig. Not exposed to the customer. I think implementation details are ok here

PR #348, opened 2018-04-13
v_3_2_5: Add /lib/modules mount for kubelet

r7vme:
This allows to use iscsi volumes. Kubelet needs iscsi_tcp module.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ version directory, and then changes are introduced.
 - Update hyperkube to version 1.10.1.
 - Changed kubelet bind mount mode from "shared" to "rshared".
 - Disabled etcd3-defragmentation service in favor systemd timer.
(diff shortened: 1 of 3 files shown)


calvix:
@r7vme don't you also need `/dev` and `/usr/sbin/iscsid` mounted? AS this was something i need for dynamic provisioning of iscsi for openebs and i would think its kinda same

r7vme:
I'm not 100% confident, but afaik kubelet only uses `iscsiadm` to manage iscsi (e.g. [here](https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/iscsi/iscsi_util.go#L248)). I also did not see any issues on freshly recreated nodes. Only this change was needed.

calvix:
I had this issue on openebs https://github.com/openebs/openebs/issues/160

PR #353, opened 2018-04-19
v_3_2_5: Disable SSL passthrough in nginx-ingress-controller

r7vme:
This change disables unnecessary built-in SSL proxying.

Tested in `gauss`. SSL passthru still works fine.

With this change we still can use SSL passthru, because we have AWS ELB proxy protocol enabled. In case, of KVM/Azure we also need to have proxy protocol enabled in external load balancer. Enabling this flag in nginx-ingress-controller only makes sense if no external load balancer if used.

See my comment for details: https://github.com/kubernetes/ingress-nginx/issues/2354#issuecomment-382757307

Fixes: https://github.com/giantswarm/giantswarm/issues/2948

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ version directory, and then changes are introduced.
 - Updated Calico to 3.0.5.
 - Updated Etcd to 3.3.3.
 - Added trusted certificate CNs to aggregation API allowed names.
+- Disabled SSL passthrough in nginx-ingress-controller.
 
 ### Removed
 - Removed docker flag "--disable-legacy-registry".
--- a/v_3_2_5/master_template.go
+++ b/v_3_2_5/master_template.go
@@ -654,7 +654,6 @@ write_files:
             - /nginx-ingress-controller
             - --default-backend-service=$(POD_NAMESPACE)/default-http-backend
             - --configmap=$(POD_NAMESPACE)/ingress-nginx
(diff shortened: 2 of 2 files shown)


calvix:
>  With this change we still can use SSL passthrough, because we have AWS ELB proxy protocol enabled

IMHO this si false, as if you enable proxy protocol in ELB than ELB is sending http packet encapsulated proxy protocol. If you forward this to simple http endpoint, it will fail.

IN KVM/Azure we don't enable proxy protocol because nothing is sending proxy encapsulated packets.

I think you can easily test this and see.

puja108:
Yeah, I think @calvix has a point there. Maybe @teemow also knows some things in this context?

r7vme:
@calvix i've tested that we still can use ssl pass through in `gauss` and can show you how it works. See my comment for details https://github.com/kubernetes/ingress-nginx/issues/2354#issuecomment-382757307

> If you forward this to simple http endpoint, it will fail.

You're right, that's why we still need `nginx.ingress.kubernetes.io/ssl-passthrough: "true"` in ingress manifest. But `--enable-ssl-passthrough` is nothing to do with nginx. Flag name is just confusing. It should have been called smth like `--enable-internal-ssl-proxy`.

r7vme:
@calvix Do you want me to demonstrate things for confidence? need your approval.

PR #354, opened 2018-04-19
add master/worker CNs to aggregator allowed names

puja108:
towards https://github.com/giantswarm/giantswarm/issues/2973

TLDR:
We reuse our CA and certs for aggregation API thus we need to allow at least for the known CNs (master and worker certs) which are the two templated domains I added so that aggregation API and services that use it do not log errors all the time (see PM)


xh3b4sd:
Please ping @giantswarm/sig-operator as a whole to get the knowledge share and awareness. Other people can help as well and we can maybe distribute efforts more evenly.

PR #356, opened 2018-04-24
Add trusted CA into sshd config

corest:
Towards: https://github.com/giantswarm/giantswarm/issues/2867

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ version directory, and then changes are introduced.
 
 ## [v3.4.0] WIP
 
+### Added
+- Added SSO public key into ssh trusted CA.
+
 ### Changed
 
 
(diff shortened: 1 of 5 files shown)


corest:
@calvix Nope, it will not break anything
@r7vme I'm fine with this

corest:
@r7vme ping again

calvix:
I see some changes to v3_4_0 files and some to v3_4_4. This wont probably work :)

calvix:
Also I am just curious how many ppl are actually using the CA-based ssh auth :p

corest:
Well, I can't force anyone to use that

calvix:
I think we need to add this to `v3_4_4` as if you add new k8s version are always based on the latest and this would mean that the latest doesn't have the changes

PR #358, opened 2018-04-25
Add RBAC missing rules

pipo02mix:
Towards https://github.com/giantswarm/giantswarm/issues/2985


pipo02mix:
I have tested copying the file contents and adding in as cluster role.

PR #359, opened 2018-04-25
v_3_2_5: Remove explicit hostname labeling

r7vme:
This is exception for already freezed 3.2.5 release.

Problem that it was broken for a while (from beginning?). We either can remove brackets `{}` to let systemd handle environment variable (still not clear why brackets break this) or remove explicit label setup completely. I've tested in kvm/azure/aws and kubelet properly detects hostname.

Fixes: https://github.com/giantswarm/giantswarm/issues/2999


r7vme:
Merging.

PR #360, opened 2018-04-26
Name ports for node exporter

pipo02mix:
Towards https://github.com/giantswarm/vodafone/issues/324


JosephSalisbury:
team batman is working on some chart for node-exporter that replaces this in future, iirc - if they do, this should probably be there too

PR #366, opened 2018-05-11
Remove kube-state-metrics

fgimenez:
Towards https://github.com/giantswarm/giantswarm/issues/1902

The component is hosted in https://github.com/giantswarm/kubernetes-kube-state-metrics/. It will be managed by chart-operator as part of the managed services story.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ version directory, and then changes are introduced.
 - Updated hyperkube to version 1.10.2.
 
 ### Removed
+- Removed kube-state-metrics related components (will be managed by
+chart-operator).
 
 
 ## [v3.2.6]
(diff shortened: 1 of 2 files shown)


r7vme:
@giantswarm/sre see my comment above

puja108:
@r7vme there's a migration script in that can be triggered pre-install of the chart itself. @rossf7 and @fgimenez can tell you more about the details

/cc @giantswarm/sig-operator (FYI, as we are starting to move yamls out of the k8scc and with that also out of the operators)

fgimenez:
Yes, each component ships a migration chart as a dependency. The migration chart is installed before the target chart and its `Job` does the defined cleanup using https://github.com/giantswarm/k8s-migrator/. 

See https://github.com/giantswarm/kubernetes-node-exporter/tree/master/helm/kubernetes-node-exporter-chart/charts/migration for an example of migration chart and https://github.com/giantswarm/kubernetes-node-exporter/pull/14#issuecomment-387001161 for the motivation about the approach taken (local migration chart, resources as pre-install hooks, hooks weight, etc.).

calvix:
so nice removing the yaml nagic :p Go team!

PR #368, opened 2018-05-13
Freeze v_3_3_0 and make v_3_3_1 WIP

r7vme:
I'm preparing new release of azure-operator with [German Cloud support](https://github.com/giantswarm/azure-operator/pull/203) and i need k8scc with 1.10.2.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,14 @@ version directory, and then changes are introduced.
 
 
 
-## [v3.3.0] WIP
+## [v3.3.1] WIP
+
+### Changed
+
+### Removed
+
+
+## [v3.3.0]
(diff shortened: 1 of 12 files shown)


rossf7:
`v_3_3_0` includes removing kube-state-metrics so it can be managed by chart-operator. So we need to release cluster-operator at the same time or move this change to a later release.

The migration is looking OK in testing so I think we should deploy the cluster-operator change. @puja108 @fgimenez WDYT?

puja108:
ack, let's deploy the change then.

r7vme:
No rush, looks like we don't need this change ASAP (we can not release azure-operator, because of blocker.) Closing this PR.

PR #369, opened 2018-05-17
move to retagged images

puja108:
towards https://github.com/giantswarm/adidas/issues/320 goes with https://github.com/giantswarm/retagger/pull/85

### DIFF
--- a/v_3_3_1/master_template.go
+++ b/v_3_3_1/master_template.go
@@ -138,7 +138,7 @@ write_files:
             # container programs network policy and routes on each
             # host.
             - name: calico-node
-              image: quay.io/calico/node:v3.0.5
+              image: quay.io/giantswarm/node:v3.0.5
               env:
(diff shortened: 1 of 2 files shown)


v_3_3_1/master_template.go
```diff
@@ -642,7 +642,7 @@ write_files:
             - sh
             - -c
             - sysctl -w net.core.somaxconn=32768; sysctl -w net.ipv4.ip_local_port_range="1024 65535"
-            image: alpine:3.6
+            image: quay.io/giantswarm/alpine:3.6
```
fgimenez: fwiw 3.7 is out, do we stick to 3.6 for any reason?

puja108: I guess we could use this to upgrade here, also for security, WDYT?

fgimenez: :+1: fine with me

PR #373, opened 2018-05-22
 v_3_3_1: switch hyperkube to v1.10.3

rossf7:
Updates hyperkube version to v1.10.3 which is latest and has fixes for kubelet we want.

> kubelet: fix hangs in updating Node status after network interruptions/changes between the kubelet and API server (#63492, @liggitt)

Full [changelog](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.10.md#changelog-since-v1102).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ version directory, and then changes are introduced.
 - Changed some remaining images to be pulled from Giant Swarm's registry
 - Updated Alpine sidecar for Ingress Controller to version 3.7
 - Fixed mkfs.xfs for containerized kubelet.
+- Updated hyperkube to version 1.10.3.
 
 ## [v3.3.0]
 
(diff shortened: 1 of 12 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ version directory, and then changes are introduced.
 - Changed some remaining images to be pulled from Giant Swarm's registry
 - Updated Alpine sidecar for Ingress Controller to version 3.7
 - Fixed mkfs.xfs for containerized kubelet.
+- Updated hyperkube to version 1.10.3.
```
MarcelMue: Do we finish a change with a `.` or not? Half of them have it and half of them don't.

rossf7: Good catch. IMO they should end with a `.`. There were a couple missing, now done.

MarcelMue: Thanks <3

rossf7:
Tested on gauss and in giantswarm/aws-operator/pull/949.

PR #375, opened 2018-05-24
copy v_3_3_1 to v_3_3_2

rossf7:


v_3_3_2/master_template.go
```diff
@@ -0,0 +1,2316 @@
+package v_3_3_1
```
indyfree: I don't know much about the process, but the package seems to be `v_3_3_1` everywhere in the `v_3_3_2` folder. Or is that intentional?

rossf7: Yes, the copy PR is just to create the new files. This makes the PR easier to review. Next PR will make v_3_3_2 the new WIP and I'll fix the imports then.

Like vendor PRs the reviews on copy PRs are for information. So you can merge without a review. The process is basically the same as for new version bundles.

indyfree: Ah alright. Thanks for the enlightment! :bulb:

PR #376, opened 2018-05-24
Freezes v3.3.1 and makes v3.3.2 WIP

rossf7:


CHANGELOG.md
```diff
@@ -179,6 +189,7 @@ chart-operator).
 
 
 
+[v3.3.1]: https://github.com/giantswarm/k8scloudconfig/commits/master/v_3_3_1
```
xh3b4sd: `v3.3.2` is missing, which you just added.

rossf7: Good catch. Done.

CHANGELOG.md
```diff
@@ -9,14 +9,24 @@ The latest version is considered WIP and it is a subject of change. All other
...
+### Removed
+
+
+## [v3.3.1]
 
 ### Changed
 - Changed some remaining images to be pulled from Giant Swarm's registry.
 - Updated Alpine sidecar for Ingress Controller to version 3.7.
 - Fixed mkfs.xfs for containerized kubelet.
 - Updated hyperkube to version 1.10.3.
```
puja108: I still think this should say `Kubernetes` or at least `Kubernetes (hyperkube)` as hyperkube is just the name of the container and the component is actually Kubernetes.

rossf7: Yes, that makes sense to me as k8s is the component. I went with `Kubernetes (hyperkube)`.

PR #378, opened 2018-05-31
Add parameter to disable Ingress Controller resources

rossf7:
Towards giantswarm/giantswarm#1902

Makes Ingress Controller resources configurable. This is so we can migrate providers to chart-operator independently.


rossf7:
Closing in favour of #387.

PR #379, opened 2018-06-05
v_3_3_2: downgrade Kubernetes to 1.10.2 and freeze

rossf7:
Towards giantswarm/giantswarm/issues/3332

Downgrades K8s to 1.10.2 due to a regression with config maps. Also freezes v_3_3_2 so it can be used in AWS release 4.2.0.


rossf7:
I missed some hyperkube images but done now. Going with this.

PR #380, opened 2018-06-05
copy v_3_3_2 to v_3_3_3

rossf7:
Adding the new version.


rossf7:
Copy PR. Going with this.

PR #387, opened 2018-06-11
Add parameter for disabling ingress controller

rossf7:
Towards giantswarm/giantswarm#1902

Allows disabling of Ingress Controller. This is so we can migrate it to chart-operator on Azure first.

### DIFF
--- a/v_3_3_4/master_template.go
+++ b/v_3_3_4/master_template.go
@@ -531,6 +531,7 @@ write_files:
         }
       ]
     }
+{{ if not .DisableIngressController -}}    
 - path: /srv/default-backend-dep.yml
   owner: root
   permissions: 0644
@@ -718,6 +719,7 @@ write_files:
         targetPort: 443
       selector:
         k8s-app: nginx-ingress-controller
+{{ end -}}
 - path: /srv/kube-proxy-sa.yaml
   owner: root
   permissions: 0644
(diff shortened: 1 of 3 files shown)


v_3_3_4/types.go
```diff
@@ -15,6 +15,10 @@ type Params struct {
 	// DisableEncryptionAtREST flag when set removes all manifests from the cloud
 	// config related to Kubernetes encryption at REST.
 	DisableEncryptionAtREST bool
+	// DisableIngressController flag when set removes all manifests from the
```
indyfree: Nitpick: `DisableIngressController flag when set` doesn't read very nicely for me. But maybe thats just me. Or is there a comma missing? `when set, removes...`?

rossf7: Yes, I see what you mean. I added a full stop after flag so the new sentence describes what it does.

v_3_3_4/master_template.go
```diff
@@ -1283,7 +1289,7 @@ write_files:
               /usr/bin/docker run -e KUBECONFIG=${KUBECONFIG} --net=host --rm -v /srv:/srv -v /etc/kubernetes:/etc/kubernetes $KUBECTL apply -f /srv/$manifest
               [ "$?" -ne "0" ]
           do
-              echo "failed to apply /src/$manifest, retrying in 5 sec"
+              echo "failed to apply /srv/$manifest, retrying in 5 sec"
```
rossf7: I'm sneaking this fix in. As these paths were wrong.

PR #389, opened 2018-06-13
increase kubelet start timeout

calvix:
towards https://github.com/giantswarm/giantswarm/issues/3386

Pulling hyperkube image for kubelet can take longer, and if it hits timeout systemd will terminate it and the process starts again so its take much more time to spin up.


calvix:
Maybe for china we should have even bigger limit as that pull can take auite long. It would make sense to make it configurable via operators flags and have default of `240`

rossf7:
@calvix Yes I think that can be useful for China. I'll create an issue.

PR #390, opened 2018-06-13
Set eviction threshold for  memory

calvix:
As lots of our customer often overloads machines and don't set proper request/limits this is very helpful.

This configuration will make sure that we always have some resources on machines and if the free memory gets too low kubelet will start evicting pod and machine should survive.

Soft eviction is waiting for graceful pod termination
Hard eviction doesn't wait for graceful pod termination

Needs to be properly tested.

WDYT guys?


r7vme:
Looking thru following issue, i feel that we need to be really careful with this as kubelet can do "false positive" evictions, when it thinks no memory, but it was just a page cache.

https://github.com/kubernetes/kubernetes/issues/43916

calvix:
Good point, that could be dangerous

calvix:
So I did some testing here. 
Page cache is counted as free memory. With that in mind, no eviction is executed unless there is real low memory on the system.

It can be tested nicely as when you start node there is always like 2-4GB in page cache and when you start memory eating pods page chance is freed.

Kubelet seems to pretty nice work similar to `free`  but the line where cache/buffers are considered.


Anyway, I would really appreciate if also another @giantswarm/sre can test this to be sure we are on the right track here.

You can simply test this change when you deploy azure operator with branch `test-guest-cluster-eviction`.
Then create a cluster with release `0.5.0` (latest). I did testing via creating single pod (`calvix/dev` image) and running `stress -m 1 --vm-bytes 6810M --vm-keep` which will eat some memory. You can change the memory value to play with it. 

In kubelet logs you are looking for `eviction manager` entries.

v_3_3_4/master_template.go
```diff
@@ -2020,7 +2020,7 @@ coreos:
       [Service]
       TimeoutStartSec=300
       Restart=always
-      RestartSec=0
+      RestartSec=2
```
r7vme: Is that change related?

calvix: Change is not related, it was just something Timo suggested, anyway I will move this to another PR. It was there just for some tests.

calvix:
ping @giantswarm/sig-host-cluster @giantswarm/sig-customer @giantswarm/sig-operator 
Hello I would like to get some bigger feedback as it might be not that small change.

This PR introduced kubelet eviction policy based on free memory in the OS. When the free memory on the machine gets under threshold kubelet will start evicting ( that mean killing pod and rescheduling somewhere else) to protect the OS. 

This solution was created because we often have issues when customer overload machines with pods without a proper set of requests/limits  (especially on memory) which causes the node to die.

Having eviction policies *should* prevent nodes from dying which should be beneficial for both us and customer. But the customer could be surprised that his pods were killed.

As we use priority classes for pods like calico or ingress, we should be safe and infra pods should not be evicted.

calvix:
Please put any feedback concerns about this.

corest:
Maybe as additional step we can setup default `RangeLimit`  (https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/memory-default-namespace/) ?

calvix:
`RangeLimit` is interesting, but from the quick look, you need to create that config in each namespace which is hard for a customer.

puja108:
Yes, I recently also looked into adding Limits and Quotas by default, but they are namespace specific and we cannot predict namespaces.
We should enable range limit admission controller though if we have not yet.

We can include that in a best practice guide then that you should add rangelimits and quotas to your namespaces.

Maybe related: I was recently thinking of a "multi-tenancy controller/operator". I will write my thoughts for that down and ping you guys on it.

pipo02mix:
I am thinking it we should communicate it to the customers or not. I am inclined to write a note in slack

puja108:
Yes, definitely should be communicated

calvix:
> I am thinking it we should communicate it to the customers or not. I am inclined to write a note in slack

Probably yes, but not to scare everyone, this will come in a certain release version.

v_3_4_0/master_template.go
```diff
@@ -2090,6 +2090,10 @@ coreos:
       --node-labels="node-role.kubernetes.io/master,role=master,ip=${DEFAULT_IPV4},{{.Cluster.Kubernetes.Kubelet.Labels}}" \
       --kube-reserved="cpu=200m,memory=250Mi" \
       --system-reserved="cpu=150m,memory=250Mi" \
+      --eviction-soft='memory.available<500Mi' \
+      --eviction-hard='memory.available<350Mi' \
+      --eviction-soft-grace-period='memory.available=15s' \
```
tuommaki: Is `500Mi` with `15s` a good combination for soft thresholds when some workloads can hog available resources very quickly? I can imagine the time taken from `500Mi` -> `350Mi` is very short and `eviction-soft-grace-period` states following:
> eviction-soft-grace-period describes a set of eviction grace periods (e.g. memory.available=1m30s) that correspond to how long a soft eviction threshold must hold before triggering a Pod eviction.

This makes me think if `soft-grace-period` of `5s` or even `3s` would be better? I don't have hard evidence, but I'd claim that most of the cases that will hit this are from use of various runtimes with GC and in those cases it is totally exceptional to release allocated memory at least in any meaningful period of time.

To summarize: I'm slightly afraid that time from `500Mi` -> `350Mi` is very short and therefore in most cases kubelet starts hard eviction practically immediately and given conditions, it depends on workload if kubelet gets to a point of selecting the right POD to terminate. If there are processes that use a lot of memory and I/O but which have even remotely reasonable requests values (ElasticSearch or Nexus indexing), it could be that those processes slow down the whole machine so that we would still get occasional flapping NodeStatus / KubeletDown alerts.

Nevertheless, disclaimer: These are just thoughts that came from the values & k8s docs and I don't have any hard evidence backing what I'm saying and therefore I can be totally wrong here. Just 2 cents :)

jgsqware:
Direct customer will be happy with this. Customers's customer will probably
have fear about this !
But I vote for it as we had too many problem with this.

We, end of the month, argue we don't have less node failure and reboot

Le mar. 19 juin 2018 à 16:07, Tuomas Mäkinen <notifications@github.com> a
écrit :

othylmann:
The only question I have @calvix what happens when the entire cluster is overloaded. Will it start evicting stuff left and right just to find there is no space or will it stop evicting nodes?

v_3_4_0/master_template.go
calvix: sure, this period is something to consider, I started with the example value of 90s but that felt way too long so I decreased it, but I wasn't sure where I should stop.

5 sec might not be such bad idea.

r7vme: Please don't go this low. Normal application need a bunch of time to properly shut down after receiving SIGTERM. By default afaik kubernetes has graceful termination timeout 30 seconds. May be let's just use 30sec too.

r7vme:
> Will it start evicting stuff left and right just to find there is no space or will it stop evicting nodes?

this is actually very reasonable question. Yes, it will start evicting them back and forth.

v_3_4_0/master_template.go
calvix: We are talking about the grace period for soft threshold, that means that the memory has to be under the limit for this period.

shutdown pod grace period is still same and set to 60s for now

calvix: https://github.com/giantswarm/k8scloudconfig/pull/390/files#diff-c968cf84270a51fc1f48f59c2f86f053R2096

puja108:
> > Will it start evicting stuff left and right just to find there is no space or will it stop evicting nodes?

> this is actually very reasonable question. Yes, it will start evicting them back and forth.

Still, the nodes and our components will stay healthy, and for overloaded clusters there needs to be alerting early on and communication with the customer.

calvix:
> The only question I have @calvix what happens when the entire cluster is overloaded. Will it start evicting stuff left and right just to find there is no space or will it stop evicting nodes? 

If the cluster is very overloaded then it would not be able to do work anyway and this would cause only a few pods that would be evicting instead of whole nodes being dead.

But having alerts for this would be quite useful.

calvix:
I made the same change to host cluster and leave it running for a week or so to see if there are any side effect, after that this will be merged.

PR #391, opened 2018-06-13
increase-timeoutot-300

calvix:


calvix:
just pushing the value to 300 to align with host cluster

PR #393, opened 2018-06-15
v_3_3_4: Update VerbatimSections to return an error

rossf7:
For China I need to disable encryption for the K8s storage class we create. As it uses KMS which is not yet available in AWS China regions.

Since I need to render a template I've updated VerbatimSections to return an error.

@corest you added v_3_4_0. Should I be using this for dev?

### DIFF
--- a/v_3_3_4/common_template_test.go
+++ b/v_3_3_4/common_template_test.go
@@ -6,6 +6,6 @@ func (nopWriter) Write(b []byte) (int, error) { return len(b), nil }
 
 type nopExtension struct{}
 
-func (nopExtension) Files() ([]FileAsset, error)         { return nil, nil }
-func (nopExtension) Units() ([]UnitAsset, error)         { return nil, nil }
-func (nopExtension) VerbatimSections() []VerbatimSection { return nil }
+func (nopExtension) Files() ([]FileAsset, error)                  { return nil, nil }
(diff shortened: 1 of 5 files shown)


v_3_3_4/common_template_test.go
```diff
@@ -6,6 +6,6 @@ func (nopWriter) Write(b []byte) (int, error) { return len(b), nil }
 
 type nopExtension struct{}
 
-func (nopExtension) Files() ([]FileAsset, error)         { return nil, nil }
-func (nopExtension) Units() ([]UnitAsset, error)         { return nil, nil }
-func (nopExtension) VerbatimSections() []VerbatimSection { return nil }
+func (nopExtension) Files() ([]FileAsset, error)                  { return nil, nil }
+func (nopExtension) Units() ([]UnitAsset, error)                  { return nil, nil }
+func (nopExtension) VerbatimSections() ([]VerbatimSection, error) { return nil, nil }
```
rossf7: The other methods already return an error.

corest:
Well, as `3.3.4` already in use in Azure - how can you update it now again?

rossf7:
There is no changelog for 3.4.0 so I wasn't sure which version is WIP.

Anyway closing as this isn't needed. It was to disable KMS for China but Fede has done this as part of his Vault changes.

PR #395, opened 2018-06-19
make-3_4_0_WIP

calvix:
set 3.3.4 as stable
set 3.4.0 as WIP


calvix:
first time doing this :) thanks for hints

All good now @r7vme ?

PR #396, opened 2018-06-21
add-rbac-for-node-operator

calvix:
add rbac rules for node operator


calvix:
{{.Cluster.Kubernetes.Domain}} containts `cluster.local` :( i thought it will have the cluster DNS domain like `508iz.k8s.gauss.eu-central-1.aws.gigantic.io`  

Not sure how to get this now

xh3b4sd:
@kopiczko recently worked with the domain magics maybe he has an idea.

kopiczko:
The long term goal is to have `baseDomain: k8s.gauss.eu-central-1.aws.gigantic.io` in the CR. Then you could achieve that by combining clusterID and baseDomain. For now you don't have that option.

For now I recommend taking `{{.Cluster.Kubernetes.API.Domain}}` and removing `api.` prefix and setting that in the Params struct like we set etcd port (maybe w/o surrounding `if` statement):

https://github.com/giantswarm/k8scloudconfig/blob/db8017556986e8783e398d8a460aed39fde7aff5/v_3_4_0/cloudconfig.go#L39-L41

It will got eventually sorted in scope of https://github.com/giantswarm/giantswarm/issues/2383

xh3b4sd:
Yeah for sure. When this just works we can revert the domain change in the `cluster-operator` soon.

PR #398, opened 2018-06-22
Fix templating whitespaces and new rows

corest:
Previous fix was not tested properly on Azure as there was no issue before the fix. So this fix fixes fix :trollface:


corest:
Tested both on azure and kvm

PR #400, opened 2018-06-27
enable-monitoring-for-IC

calvix:
enable scraping of IC
towards https://github.com/giantswarm/giantswarm/issues/3537

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ version directory, and then changes are introduced.
 ### Added
 - Added SSO public key into ssh trusted CA.
 - Added RBAC rules for node-operator.
+- Enabled monitoring for ingress controller metrics.
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


JosephSalisbury:
Can we remove the worker scraping now? IIRC that was a hacky way we wanted to monitor ingress in guest clusters. 

I mean this https://github.com/giantswarm/kvm-operator/pull/412

calvix:
@JosephSalisbury  no, as they are not same.
1) this monitor/scrape Ic controllers in guest clusters  (it  scrapes IC pod ip)
2) the service in kvm operator is monitoring if each node has running and working Ingress connection ( it connects to NodePort which forwards to IC controller.)

There can be cases when 1 is working but 2 is not. For example, one node is online but not joined K8s cluster.

JosephSalisbury:
Legit. Thanks!

PR #401, opened 2018-07-02
Parameterize image registry

fgimenez:
Adds parameter for registry domain, defaults to `quay.io`.

### DIFF
--- a/v_3_4_0/types.go
+++ b/v_3_4_0/types.go
@@ -37,7 +37,7 @@ type Params struct {
 	// then apply the manifest by adding it to ExtraManifests.
 	ExtraManifests []string
 	Node           v1alpha1.ClusterNode
-	// Domain of the docker image registry to use.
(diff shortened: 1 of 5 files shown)


v_3_4_0/types.go
```diff
@@ -37,6 +37,8 @@ type Params struct {
 	// then apply the manifest by adding it to ExtraManifests.
 	ExtraManifests []string
 	Node           v1alpha1.ClusterNode
+	// Domain of the docker image registry to use.
```
rossf7: Should be RegistryDomain.

fgimenez:
> lgtm if you are sure that you found all the use cases of hardcoded quay :)

At least here I think so! :) We still need to update the already extracted components (those managed by chart-operator) to be able to configure the registry to use, that will come once this part is done.

v_3_4_0/types.go
fgimenez: Good point, done.

PR #405, opened 2018-07-03
add rbac rules for prometheus

calvix:
add RBAC rules for the host cluster prometheus ( will be needed when we switch to specific certs per service, now we mostly use `api.xxx` certs but we will use `prometheus.xxx`)

RBAC rules copied from https://github.com/prometheus/prometheus/blob/master/documentation/examples/rbac-setup.yml

### DIFF
--- a/v_3_4_0/master_template.go
+++ b/v_3_4_0/master_template.go
@@ -908,6 +908,20 @@ write_files:
       name: node-operator
       apiGroup: rbac.authorization.k8s.io
     ---
+    ## prometheus-external is prometheus from host cluster
+    kind: ClusterRoleBinding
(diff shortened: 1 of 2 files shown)


v_3_4_0/master_template.go
```diff
@@ -908,6 +908,20 @@ write_files:
       name: node-operator
       apiGroup: rbac.authorization.k8s.io
     ---
+    ## prometheus-external is prometheus from host cluster
+    kind: ClusterRoleBinding
+    apiVersion: rbac.authorization.k8s.io/v1
+    metadata:
+      name: prometheus-external
```
calvix: `-external` so that we dont collied for possible RBAC rules for prometheus that run in-cluster

PR #409, opened 2018-07-19
v_3_5_0: Use separate parameter for disabling the ingress controller service

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/3710

Adds a new parameter so the Ingress Controller service can be enabled or disabled separately from the other resources.

This is so on AWS and KVM we can migrate to chart-operator but continue using the existing controller service. The service will be replaced in a later release to avoid causing downtime.


rossf7:
Tested on ginger and ghost.

v_3_5_0/master_template.go
```diff
@@ -1440,6 +1442,8 @@ write_files:
       MANIFESTS="${MANIFESTS} default-backend-svc.yml"
       MANIFESTS="${MANIFESTS} ingress-controller-cm.yml"
       MANIFESTS="${MANIFESTS} ingress-controller-dep.yml"
+      {{ end -}}
+      {{ if not .DisableIngressControllerService -}}
```
kopiczko: Those indentations look weird. Maybe just put the `if`s at the very beggining of the line. This will give extra visibility what can be disabled.

rossf7: I see what you mean about the indents but these are the same as for the current IC param. I'll test without them and see how it looks.

rossf7:
I tested without indentation but e2e failed in azure-operator. Going with the tested fix.

PR #410, opened 2018-07-27
Disable HSTS headers for Ingress Controller

rossf7:
See https://github.com/giantswarm/kubernetes-nginx-ingress-controller/pull/23 from @marians. I'd like to keep the chart and the manifests here in sync until the AWS and KVM migration is complete.


marians:
Thanks!

PR #412, opened 2018-08-02
copy v_3_5_0 to v_3_6_0

rossf7:


rossf7:
Copy PR. Going with this.

PR #413, opened 2018-08-02
Make v_3_6_0 WIP

rossf7:
Freezes v_3_5_0 and makes v_3_6_0 WIP.


rossf7:
I want to merge the azure-operator change in this PR https://github.com/giantswarm/azure-operator/pull/298

I've tested in aws-operator but it isn't ready to merge until the cluster-operator change is ready. AIUI k8scloudconfig should be frozen before merging in any of the operators.

kopiczko:
> Do we need to release the provider operators now as well?

I don't think we need to do that. But we need to do it other way around. We need to release k8scloudconfig before creating provider active release.

PR #414, opened 2018-08-15
Switch to k8s 1.11.1

corest:


v_3_6_0/worker_template.go
```diff
@@ -300,32 +356,17 @@ coreos:
...
       --kubeconfig=/etc/kubernetes/config/kubelet-kubeconfig.yml \
       --node-labels="ip=${DEFAULT_IPV4},{{.Cluster.Kubernetes.Kubelet.Labels}}" \
-      --kube-reserved="cpu=200m,memory=250Mi" \
-      --system-reserved="cpu=150m,memory=250Mi" \
-      --eviction-soft='memory.available<500Mi' \
-      --eviction-hard='memory.available<350Mi' \
-      --eviction-soft-grace-period='memory.available=5s' \
-      --eviction-max-pod-grace-period=60 \
-      --enforce-node-allocatable=pods \
       --v=2"
```
kopiczko: So not all settings are moved to config files? Wow...

v_3_6_0/master_template.go
```diff
@@ -1503,6 +1476,37 @@ write_files:
...
+    apiVersion: kubelet.config.k8s.io/v1beta1
+    address: ${DEFAULT_IPV4}
+    port: 10250
+    healthzBindAddress: ${DEFAULT_IPV4}
+    healthzPort: 10248
+    clusterDNS:
+      - {{.Cluster.Kubernetes.DNS.IP}}
+    clusterDomain: {{.Cluster.Kubernetes.Domain}}
+    featureGates:
+      ExpandPersistentVolumes: true
```
tuommaki: `CustomResourceSubresources: true` is required and missing. What about `PodPriority=true`?

v_3_6_0/worker_template.go
corest: Not all, but component scoped

v_3_6_0/worker_template.go
```diff
@@ -209,6 +250,21 @@ coreos:
...
+    command: start
+    content: |
+      [Unit]
+      Description=k8s-setup-kubelet-config Service
+      After=k8s-setup-network-env.service docker.service
+      Requires=k8s-setup-network-env.service docker.service
+
+      [Service]
+      EnvironmentFile=/etc/network-environment
+      ExecStart=/bin/bash -c '/usr/bin/envsubst </etc/kubernetes/config/kubelet-config.yaml.tmpl >/etc/kubernetes/config/kubelet-config.yaml'
```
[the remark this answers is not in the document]
corest: True. Don't have better ideas

corest:
I actually removed all the feature flags because found in docs they are enabled by default. But looking again in docs turned out that in some places it has one value, on others - another value. So for now I'll bring all the feature-flags back and add issue to removed them with 1.12

corest:
Nah, I decided to find out what are default values. So, here is from help output in kubelet:
```
ExpandPersistentVolumes=true|false (BETA - default=true)
CustomResourceSubresources=true|false (BETA - default=true)
PodPriority=true|false (BETA - default=true)
```

Scheduler:
```
PodPriority=true|false (BETA - default=true)
ExpandPersistentVolumes=true|false (BETA - default=true)
CustomResourceSubresources=true|false (BETA - default=true)
```

Controller
```
ExpandPersistentVolumes=true|false (BETA - default=true)
PodPriority=true|false (BETA - default=true)
CustomResourceSubresources=true|false (BETA - default=true)
```

So, all those are enabled by default. Removing

PR #415, opened 2018-08-22
Upgrade calico to 3.2.0

r7vme:
Part of: https://github.com/giantswarm/giantswarm/issues/3922

also move to single manifest like we do in "control plane" cluster. this allows to copy-paste official upstream manifests.

### DIFF
--- a/v_3_6_0/master_template.go
+++ b/v_3_6_0/master_template.go
@@ -48,7 +48,7 @@ write_files:
       calico_backend: "bird"
 
       # Configure the MTU to use
-      veth_mtu: "1440"
+      veth_mtu: "{{.Cluster.Calico.MTU}}"
 
       # The CNI network configuration to install on each node.  The special
       # values in this config will be automatically populated.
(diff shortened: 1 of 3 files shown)


v_3_6_0/master_template.go
```diff
@@ -18,34 +18,16 @@ write_files:
...
-    metadata:
-      name: calico-node
-      namespace: kube-system
-- path: /srv/calico-configmap.yaml
-  owner: root
-  permissions: 644
-  content: |
-    # Calico Version v3.0.5
-    # https://docs.projectcalico.org/v3.0/releases#v3.0.5
+    # Calico Version v3.2.0
```
r7vme: Just upstream manifest with few changes for parametrized values.

v_3_6_0/master_template.go
```diff
@@ -2153,6 +2189,7 @@ coreos:
       -v /run/docker.sock:/run/docker.sock:rw \
       -v /usr/lib/os-release:/etc/os-release \
       -v /usr/share/ca-certificates/:/etc/ssl/certs \
+      -v /var/lib/calico/:/var/lib/calico \
```
r7vme: From 3.1 they store node name in `/var/lib/calico`, which afaik is used by CNI. So we need to mount it for kubelet

PR #416, opened 2018-09-14
remove coreDNS components

fgimenez:
Towards https://github.com/giantswarm/giantswarm/issues/1902

Removes coreDNS components that will be managed by chart-operator.


fgimenez:
> LGTM but tests are failing

Yep, there was an issue with on CircleCI side with VM provisioning, green after rerunning.

fgimenez:
@rossf7 good point! I've added a `DisableCoreDNS` param and updated the changelog message, PTAL.

PR #418, opened 2018-09-21
Do not remove network policy with coredns manifests

fgimenez:
Thx @puja108 for catching this!

### DIFF
--- a/v_3_6_0/master_template.go
+++ b/v_3_6_0/master_template.go
@@ -646,6 +646,7 @@ write_files:
       - name: dns-tcp
         port: 53
         protocol: TCP
+{{- end }}
 - path: /srv/network-policy.json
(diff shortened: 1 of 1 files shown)


fgimenez:
@corest Damn, i took into account that it is the highest version in the repo, also the entry in the changelog shows WIP https://github.com/giantswarm/k8scloudconfig/blob/master/CHANGELOG.md and saw that it was used in aws-operator WIP. But you are right, I see that it has been used since v15 https://github.com/giantswarm/aws-operator/blob/master/service/controller/v15/cloudconfig/master_template.go#L8 How can we proceed, should we create `v_3_7_0`?

corest:
I don't think it is even in any use. I think it can be removed in next major version I'm working on now

PR #423, opened 2018-09-24
Add filemanager for rendering content from gotemplates

corest:
Towards https://github.com/giantswarm/giantswarm/issues/2816

As the current ignition spec uses data schema uri for files, all the files will be moved into separate templates and then rendered into base64 representation. 

E.g. `v_4_0_0` will have folder `files` with subfolders `config`, `kubeconfig`, `k8s-resources` etc. Main master cloudconfig then will use next construction for rendering file in operating system

```
    - path: /srv/psp_policies.yaml
      filesystem: root
      mode: 0644
      contents:
        source: "data:text/plain;charset=utf-8;base64,{{  index .Files "k8s-resource/psp_policies.yaml" }}"
```

### DIFF
--- a/v_4_0_0/filemanager.go
+++ b/v_4_0_0/filemanager.go
@@ -3,7 +3,6 @@ package v_4_0_0
 import (
 	"bytes"
 	"encoding/base64"
-	"fmt"
 	"os"
 	"path/filepath"
 	"text/template"
--- a/v_4_0_0/filemanager.go
+++ b/v_4_0_0/filemanager.go
@@ -11,9 +11,16 @@ import (
 	"github.com/giantswarm/microerror"
 )
 
-// Files is map[string]string for files that we fetched from disk and then filled with data.
+// Files is map[string]string (k: filename, v: contents) for files that are fetched from disk
+// and then filled with data.
 type Files map[string]string
 
+// RenderFiles walks over filesdir and parses all regular files with
+// text/template. Parsed templates are then rendered with ctx, base64 encoded
+// and added to returned Files.
+//
+// filesdir must not contain any other files than templates that can be parsed
+// with text/template.
 func RenderFiles(filesdir string, ctx interface{}) (Files, error) {
 	files := Files{}
 
@@ -27,7 +34,6 @@ func RenderFiles(filesdir string, ctx interface{}) (Files, error) {
 			tmpl.Execute(&data, ctx)
 
 			relativePath, err := filepath.Rel(filesdir, path)
-			fmt.Printf("path: %s", relativePath)
 			if err != nil {
 				return microerror.Mask(err)
 			}
--- a/v_4_0_0/filemanager.go
+++ b/v_4_0_0/filemanager.go
@@ -1,0 +1,42 @@
+package v_4_0_0
+
+import (
+	"bytes"
+	"encoding/base64"
+	"io/ioutil"
+	"path"
+	"text/template"
+
+	"github.com/giantswarm/microerror"
+)
+
+// Files is map[string]string for files that we fetched from disk and then filled with data.
+type Files map[string]string
+
+func RenderFiles(filespath string, ctx interface{}) (*Files, error) {
+	files := Files{}
+	dirList, err := ioutil.ReadDir(filespath)
+	if err != nil {
+		return nil, microerror.Maskf(err, "Failed to read files dir: %s, error: %#v", filespath, err)
+	}
+
+	for _, dir := range dirList {
+		fileList, err := ioutil.ReadDir(path.Join(filespath, dir.Name()))
+		if err != nil {
+			return nil, microerror.Maskf(err, "Failed to read dir: %s, error: %#v", path.Join(filespath, dir.Name()), err)
+		}
+
+		for _, file := range fileList {
+			tmpl, err := template.ParseFiles(path.Join(filespath, dir.Name(), file.Name()))
+			if err != nil {
+				return nil, microerror.Maskf(err, "Failed to file: %s, error: %#v", path.Join(filespath, dir.Name(), file.Name()), err)
+			}
+			var data bytes.Buffer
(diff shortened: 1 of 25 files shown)


v_4_0_0/filemanager.go
```diff
@@ -0,0 +1,42 @@
...
+	"path"
+	"text/template"
+
+	"github.com/giantswarm/microerror"
+)
+
+// Files is map[string]string for files that we fetched from disk and then filled with data.
+type Files map[string]string
+
+func RenderFiles(filespath string, ctx interface{}) (*Files, error) {
```
kopiczko: Why there is a pointer returned? In general map should be returned as a value.

v_4_0_0/filemanager.go
```diff
@@ -0,0 +1,42 @@
...
+	files := Files{}
+	dirList, err := ioutil.ReadDir(filespath)
+	if err != nil {
+		return nil, microerror.Maskf(err, "Failed to read files dir: %s, error: %#v", filespath, err)
+	}
+
+	for _, dir := range dirList {
+		fileList, err := ioutil.ReadDir(path.Join(filespath, dir.Name()))
+		if err != nil {
+			return nil, microerror.Maskf(err, "Failed to read dir: %s, error: %#v", path.Join(filespath, dir.Name()), err)
```
kopiczko: Error annotations should be lower-cased, i.e. `failed to read...`.

v_4_0_0/filemanager.go
kopiczko: `filespath` is supposed to be a directory. I would make that clear in the name.

v_4_0_0/filemanager.go
```diff
@@ -0,0 +1,42 @@
...
+			tmpl, err := template.ParseFiles(path.Join(filespath, dir.Name(), file.Name()))
+			if err != nil {
+				return nil, microerror.Maskf(err, "Failed to file: %s, error: %#v", path.Join(filespath, dir.Name(), file.Name()), err)
+			}
+			var data bytes.Buffer
+			tmpl.Execute(&data, ctx)
+
+			files[dir.Name()+"/"+file.Name()] = base64.StdEncoding.EncodeToString(data.Bytes())
+		}
+	}
```
kopiczko: I'm not sure what do you intend to use this function for but [filepath.Walk](https://golang.org/pkg/path/filepath/#Walk) may be a better candidate.

v_4_0_0/filemanager.go
```diff
@@ -0,0 +1,42 @@
...
+
+		for _, file := range fileList {
+			tmpl, err := template.ParseFiles(path.Join(filespath, dir.Name(), file.Name()))
+			if err != nil {
+				return nil, microerror.Maskf(err, "Failed to file: %s, error: %#v", path.Join(filespath, dir.Name(), file.Name()), err)
+			}
+			var data bytes.Buffer
+			tmpl.Execute(&data, ctx)
+
+			files[dir.Name()+"/"+file.Name()] = base64.StdEncoding.EncodeToString(data.Bytes())
```
tuommaki: Maybe use `path.Join()` instead of `dir.Name()+"/"+file.Name()`?

corest:
refactored `RenderFiles`, ptal

v_4_0_0/filemanager.go
```diff
@@ -0,0 +1,43 @@
...
+	"path/filepath"
+	"text/template"
+
+	"github.com/giantswarm/microerror"
+)
+
+// Files is map[string]string for files that we fetched from disk and then filled with data.
+type Files map[string]string
+
+func RenderFiles(filesdir string, ctx interface{}) (Files, error) {
```
tuommaki: Maybe add godoc here? For example:
```go
// RenderFiles walks over filesdir and parses all regular files with
// text/template. Parsed templates are then rendered with ctx, base64 encoded
// and added to returned Files.                                                 
//
// filesdir must not contain any other files than templates that can be parsed
// with text/template.
```

v_4_0_0/filemanager.go
```diff
@@ -0,0 +1,43 @@
...
+	"encoding/base64"
+	"fmt"
+	"os"
+	"path/filepath"
+	"text/template"
+
+	"github.com/giantswarm/microerror"
+)
+
+// Files is map[string]string for files that we fetched from disk and then filled with data.
```
tuommaki: Would it be worth to add mention about key/val meaning:
```go
// Files is map[string]string (k: filename, v: contents) for files taht we fetched from disk and then filled with data.
```

v_4_0_0/filemanager.go
```diff
@@ -0,0 +1,43 @@
...
+		if f.Mode().IsRegular() {
+			tmpl, err := template.ParseFiles(path)
+			if err != nil {
+				return microerror.Maskf(err, "failed to parse file #%q", path)
+			}
+			var data bytes.Buffer
+			tmpl.Execute(&data, ctx)
+
+			relativePath, err := filepath.Rel(filesdir, path)
+			fmt.Printf("path: %s", relativePath)
```
tuommaki: Is this print intentional for final use or possible leftover from debugging?

corest: leftover

v_4_0_0/filemanager.go
```diff
@@ -0,0 +1,48 @@
...
+// filesdir must not contain any other files than templates that can be parsed
+// with text/template.
+func RenderFiles(filesdir string, ctx interface{}) (Files, error) {
+	files := Files{}
+
+	err := filepath.Walk(filesdir, func(path string, f os.FileInfo, err error) error {
+		if f.Mode().IsRegular() {
+			tmpl, err := template.ParseFiles(path)
+			if err != nil {
+				return microerror.Maskf(err, "failed to parse file #%q", path)
```
kopiczko: It's `%#q` :)

PR #424, opened 2018-09-24
Migrate cloudinit to ignition

corest:
Towards https://github.com/giantswarm/giantswarm/issues/2816

It is far from final result, but there are things I wanted to discuss here

### DIFF
--- a/v_4_0_0/cloudconfig_test.go
+++ b/v_4_0_0/cloudconfig_test.go
@@ -2,10 +2,11 @@ package v_4_0_0
 
 import (
 	"encoding/base64"
-	"strings"
 	"testing"
+)
 
-	"github.com/giantswarm/apiextensions/pkg/apis/provider/v1alpha1"
+const (
+	FilesDir = "files"
 )
 
 func TestCloudConfig(t *testing.T) {
@@ -41,6 +42,17 @@ func TestCloudConfig(t *testing.T) {
 	for _, tc := range tests {
 		c := DefaultCloudConfigConfig()
 
+		filesPath, err := GetFilesPath()
+		if err != nil {
+			t.Error(err)
+		}
+
+		files, err := RenderFiles(filesPath, tc.params)
+		if err != nil {
+			t.Errorf("failed to render ignition files, %v:", err)
+		}
+		tc.params.Files = files
+
 		c.Params = tc.params
 		c.Template = tc.template
 
@@ -60,73 +72,6 @@ func TestCloudConfig(t *testing.T) {
 		cloudConfigBase64 := cloudConfig.Base64()
 		if _, err := base64.StdEncoding.DecodeString(cloudConfigBase64); err != nil {
 			t.Errorf("The string isn't Base64 encoded: %v", err)
-		}
-	}
-}
-
-func TestCloudConfigTemplating(t *testing.T) {
-	tests := []struct {
-		template string
-		params   Params
-
-		expectedString string
-	}{
-		// Test that calico range is properly templated.
-		{
-			template: MasterTemplate,
-			params: Params{
-				Cluster: v1alpha1.Cluster{
-					Calico: v1alpha1.ClusterCalico{
-						Subnet: "127.0.0.1",
-						CIDR:   16,
-					},
-				},
-				Extension: nopExtension{},
-			},
-
-			expectedString: `
-                      - name: CALICO_IPV4POOL_CIDR
-                        value: "127.0.0.1/16"
-`,
-		},
-
-		// A different test for calico range templating.
-		{
-			template: MasterTemplate,
-			params: Params{
-				Cluster: v1alpha1.Cluster{
-					Calico: v1alpha1.ClusterCalico{
-						Subnet: "192.168.0.0",
-						CIDR:   24,
-					},
-				},
-				Extension: nopExtension{},
-			},
-
-			expectedString: `
-                      - name: CALICO_IPV4POOL_CIDR
-                        value: "192.168.0.0/24"
-`,
-		},
-	}
-
-	for index, test := range tests {
-		c := DefaultCloudConfigConfig()
-
-		c.Params = test.params
-		c.Template = test.template
-
-		cloudConfig, err := NewCloudConfig(c)
-		if err != nil {
-			t.Fatalf("%v: unexpected error creating cloud config: %v", index, err)
-		}
-
-		if err := cloudConfig.ExecuteTemplate(); err != nil {
-			t.Fatalf("%v: unexpected error templating cloud config: %v", index, err)
-		}
-
-		if !strings.Contains(cloudConfig.String(), test.expectedString) {
-			t.Fatalf("%v: expected string not found in cloud config: %v", index, test.expectedString)
 		}
 	}
 }
--- a/v_4_0_0/cloudconfig_test.go
+++ b/v_4_0_0/cloudconfig_test.go
@@ -85,8 +85,8 @@ func TestCloudConfigTemplating(t *testing.T) {
 			},
 
 			expectedString: `
-                - name: CALICO_IPV4POOL_CIDR
-                  value: "127.0.0.1/16"
+                      - name: CALICO_IPV4POOL_CIDR
+                        value: "127.0.0.1/16"
 `,
 		},
 
@@ -104,8 +104,8 @@ func TestCloudConfigTemplating(t *testing.T) {
 			},
 
 			expectedString: `
-                - name: CALICO_IPV4POOL_CIDR
-                  value: "192.168.0.0/24"
+                      - name: CALICO_IPV4POOL_CIDR
+                        value: "192.168.0.0/24"
 `,
 		},
 	}
--- a/v_4_0_0/filemanager.go
+++ b/v_4_0_0/filemanager.go
@@ -5,6 +5,7 @@ import (
 	"encoding/base64"
 	"io/ioutil"
 	"path"
+	"runtime"
 	"text/template"
 
 	"github.com/giantswarm/microerror"
@@ -39,4 +40,15 @@ func RenderFiles(filespath string, ctx interface{}) (*Files, error) {
 	}
 
 	return &files, nil
+}
+
+// GetFilesPath retrieves runtime path for the ignition templates
+func GetFilesPath() (string, error) {
+	_, filename, _, ok := runtime.Caller(0)
+	if !ok {
+		return "", microerror.New("failed to retrieve runtime information")
+	}
+	filesPath := path.Join(path.Dir(filename), FilesDir)
+
+	return filesPath, nil
 }
--- a/v_4_0_0/filemanager.go
+++ b/v_4_0_0/filemanager.go
@@ -1,0 +1,42 @@
+package v_4_0_0
+
+import (
+	"bytes"
+	"encoding/base64"
+	"io/ioutil"
+	"path"
+	"text/template"
+
+	"github.com/giantswarm/microerror"
+)
+
+// Files is map[string]string for files that we fetched from disk and then filled with data.
+type Files map[string]string
+
+func RenderFiles(filespath string, ctx interface{}) (*Files, error) {
+	files := Files{}
+	dirList, err := ioutil.ReadDir(filespath)
+	if err != nil {
+		return nil, microerror.Maskf(err, "Failed to read files dir: %s, error: %#v", filespath, err)
+	}
+
+	for _, dir := range dirList {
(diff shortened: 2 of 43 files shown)


v_4_0_0/cloudconfig_test.go
```diff
@@ -63,70 +75,3 @@ func TestCloudConfig(t *testing.T) {
 		}
 	}
 }
-
-func TestCloudConfigTemplating(t *testing.T) {
```
corest: I removed this test completely as files are rendered as base64 strings. Rendering itself is verified by other tests, so the question is do we want to check that those exact templates are still rendered into proper configuration?

v_4_0_0/filemanager.go
```diff
@@ -0,0 +1,54 @@
...
+			files[dir.Name()+"/"+file.Name()] = base64.StdEncoding.EncodeToString(data.Bytes())
+		}
+	}
+
+	return &files, nil
+}
+
+// GetFilesPath retrieves runtime path for the ignition templates
+func GetFilesPath() (string, error) {
+	_, filename, _, ok := runtime.Caller(0)
```
corest: I'm not sure about this approach. But as we are referencing real master template from unit tests, we also need real files, used in this template.

corest:
I'm aware this change isn't that small, but it mostly about moving parts of cloudinit into files, which are injected then into ignition

corest:
> We adding some files and units in operators.

Sure, added note about that here https://github.com/giantswarm/giantswarm/issues/2816#issuecomment-423580429

> We also need to convert to 'ignition" format (json) at some point.

Sure, but this PR is about valid templating, ignition converting with validation based on the official `2.2.0` spec will be in separate PR

> Seems nothing really change here, right?

All the cloudinit definitions are replaced with ignition definitions. So I would say this change is the biggest from all other changes related to this issue. Also this PR based on the `filemanager` PR, which is responsible for generating base64 representations of injected files

v_4_0_0/filemanager.go
```diff
@@ -0,0 +1,54 @@
+package v_4_0_0
```
calvix: I wonder if this code is really cloud config version dependent, maybe this should go one folder up so we don't copy the same code several times

calvix:
The problem we now face is that need need somehow ship the files to operators. By default `dep` will prune any file that is not used in the project and as this is not code it will probabbly remove it too. 

Then as the operator is  builded it will again be left over as we only copy the binary.

We might need to impelement gobindata or something similar

Otherwise LGTM

Fun thing: this repo will be called k8scloudconfig but it will be ignition and not cloudconfig anymore :p

r7vme:
With operator-per-version approach, which team spirit is aiming now, i think we will have to create a new repo that will use "git based" versioning and will have a proper naming e.g. k8signition or so

MarcelMue:
Ping @giantswarm/team-spirit 
We should sit together and think about this. I was not really aware of this problem, let's make sure we  understand the consequences / challenges we face here while we are still implementing the `release-operator` stuff.

v_4_0_0/filemanager.go
corest: Probably, we can move it upper into some pkg/tooling package. But as the discussion goes on, we probably will make a new repo for this, so I would proceed with this. And when there will be a decision on structuring - we can change it. 
For now implementation can be unbounded from structure imo

teemow:
My 2 cents:

 * ignition support should be introduced in a new release and all older releases should stay untouched. 
 * it should not affect the release-operator "stuff" - from the release-operator point of view ignition should just be an implementation detail.

v_4_0_0/master_template.go
```diff
@@ -2057,21 +97,15 @@ coreos:
       ExecStart=/usr/bin/docker run --rm --net=host -v /etc:/etc --name $NAME $IMAGE
       ExecStop=-/usr/bin/docker stop -t 10 $NAME
       ExecStopPost=-/usr/bin/docker rm -f $NAME
```
calvix: lots of units dont have 
```
      [Install]
      WantedBy=multi-user.target
```
Which is not needed with cloudconfig but will be necessary for ignition

tuommaki:
> By default dep will prune any file that is not used in the project and as this is not code it will probabbly remove it too.

AFAIUT: https://golang.github.io/dep/docs/Gopkg.toml.html#prune you should be able to configure pruning exceptions per-project so setting `non-go = false` for `k8scloudconfig` would probably do the job?

corest:
Tried this `non-go` option and it works in some weird way.
Only if I add `unused-packages=false` then it imports non-go files from cloudconfig. But with this option it also imports all unused packages. 
```
[[prune.project]]
  name = "github.com/giantswarm/k8scloudconfig"
  go-tests = true
  unused-packages = false
  non-go = false
```

corest:
All this stuff is useless as all assets should be moved into const strings :/

kopiczko:
Works pretty well for me:

```
$ cat main.go
package main

import _ "github.com/giantswarm/k8scloudconfig/v_4_0_0"
```

```
$ cat Gopkg.toml
# Gopkg.toml example
#
# Refer to https://golang.github.io/dep/docs/Gopkg.toml.html
# for detailed Gopkg.toml documentation.
#
# required = ["github.com/user/thing/cmd/thing"]
# ignored = ["github.com/user/project/pkgX", "bitbucket.org/user/project/pkgA/pkgY"]
#
# [[constraint]]
#   name = "github.com/user/project"
#   version = "1.0.0"
#
# [[constraint]]
#   name = "github.com/user/project2"
#   branch = "dev"
#   source = "github.com/myfork/project2"
#
# [[override]]
#   name = "github.com/x/y"
#   version = "2.4.0"
#
# [prune]
#   non-go = false
#   go-tests = true
#   unused-packages = true

[[constraint]]
  name = "github.com/giantswarm/k8scloudconfig"
  branch = "ignition"

[[prune.project]]
  name = "github.com/giantswarm/k8scloudconfig"
  unused-packages = false
  non-go = false

[prune]
  go-tests = true
  unused-packages = true
```

```
$ ls -l ./vendor/github.com/giantswarm/k8scloudconfig/v_4_0_0/files/k8s-resource/
total 72
-rw-r--r-- 1 pawel staff 14025 Sep 26 15:31 calico-all.yaml
-rw-r--r-- 1 pawel staff  3508 Sep 26 15:31 coredns.yaml
-rw-r--r-- 1 pawel staff   692 Sep 26 15:31 default-backend-dep.yaml
-rw-r--r-- 1 pawel staff   248 Sep 26 15:31 default-backend-svc.yaml
-rw-r--r-- 1 pawel staff   481 Sep 26 15:31 ingress-controller-cm.yaml
-rw-r--r-- 1 pawel staff  2646 Sep 26 15:31 ingress-controller-dep.yaml
-rw-r--r-- 1 pawel staff   479 Sep 26 15:31 ingress-controller-svc.yaml
-rw-r--r-- 1 pawel staff   208 Sep 26 15:31 k8s-encryption-config.yaml
-rw-r--r-- 1 pawel staff  2160 Sep 26 15:31 kube-proxy-ds.yaml
-rw-r--r-- 1 pawel staff    89 Sep 26 15:31 kube-proxy-sa.yaml
-rw-r--r-- 1 pawel staff   881 Sep 26 15:31 psp_bindings.yaml
-rw-r--r-- 1 pawel staff   915 Sep 26 15:31 psp_policies.yaml
-rw-r--r-- 1 pawel staff   567 Sep 26 15:31 psp_roles.yaml
-rw-r--r-- 1 pawel staff  2901 Sep 26 15:31 rbac_bindings.yaml
-rw-r--r-- 1 pawel staff  2451 Sep 26 15:31 rbac_roles.yaml
```

v_4_0_0/filemanager.go
corest: This function now exported and used in k8scloudconfig tests as well as in operator tests. It returns absolute path to the k8scloudconfig location so in tests it can find files in vendor folder. So its result is used as base dir in `GetIgnitionPath` 
On real usage `GetIgnitionPath` gets path to the ignition assets, located in Docker filesystem.

v_4_0_0/render_asset_content.go
```diff
@@ -21,3 +22,19 @@ func RenderAssetContent(assetContent string, params interface{}) ([]string, erro
 	content := strings.Split(buf.String(), "\n")
 	return content, nil
 }
+
+func RenderFileAssetContent(assetContent string, params interface{}) (string, error) {
```
corest: This function returns base64 representation of the assetContent and used in operators for generating files with ignition spec.

v_4_0_0/types.go
```diff
@@ -90,21 +91,25 @@ type HyperkubePodHostMount struct {
 type FileMetadata struct {
 	AssetContent string
 	Path         string
-	Owner        string
+	Owner        Owner
```
corest: Owner is represented with seprate user/group structures in ignition

v_4_0_0/types.go
```diff
@@ -90,21 +91,25 @@ type HyperkubePodHostMount struct {
...
 
+type Owner struct {
+	User  string
+	Group string
+}
+
 type FileAsset struct {
 	Metadata FileMetadata
-	Content  []string
+	Content  string
```
corest: Content for files is base64 string, not a list of strings

v_4_0_0/types.go
```diff
@@ -90,21 +91,25 @@ type HyperkubePodHostMount struct {
...
 }
 
 type UnitMetadata struct {
 	AssetContent string
 	Name         string
-	Enable       bool
-	Command      string
+	Enabled      bool
```
corest: alligning with ignition spec

corest:
I've added few more comments regarding changes, required for proper usage of this package in operators. 
PTAL

v_4_0_0/cloudconfig_test.go
kopiczko: No. Those tests weren't worth much IMO.

v_4_0_0/render_asset_content.go
kopiczko: Does `RenderIngnitionFileContent` name make sense here? Also you can write your comment a godoc.

v_4_0_0/render_asset_content_test.go
```diff
@@ -36,3 +36,27 @@ func TestRenderAssetContent(t *testing.T) {
...
+
+	for _, tc := range tests {
+		content, err := RenderFileAssetContent(tc.assetContent, tc.params)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		assert.Equal(t, tc.expectedContent, content, "content should be equal")
+	}
+}
```
kopiczko: It would be good to align newly written tests with what we have here https://github.com/giantswarm/fmt/blob/master/go/unit_tests.md.

v_4_0_0/filemanager.go
```diff
@@ -5,11 +5,14 @@ import (
 	"encoding/base64"
 	"os"
 	"path/filepath"
+	"runtime"
 	"text/template"
 
 	"github.com/giantswarm/microerror"
 )
 
+const FilesDir = "v_4_0_0/files"
```
kopiczko: I'd rather call it version and set the value to `v_4_0_0`, then `GetIgnitionPath` becomes clearer. Also it doesn't need to be exported.

v_4_0_0/filemanager.go
kopiczko: How is it going to be used in operator tests?

v_4_0_0/filemanager.go
```diff
@@ -46,3 +49,22 @@ func RenderFiles(filesdir string, ctx interface{}) (Files, error) {
...
+// returns /go/src/k8scloudconfig.
+// This function used only in tests for retrieving ignition assets in runtime.
+func GetPackagePath() (string, error) {
+	_, filename, _, ok := runtime.Caller(0)
+	if !ok {
+		return "", microerror.New("failed to retrieve runtime information")
+	}
+
+	return filepath.Dir(filepath.Dir(filename)), nil
+}
```
kopiczko: In general I'm don't get how the whole ignition template is going to be generated. How is the `Files` type used and where?

v_4_0_0/render_asset_content.go
kopiczko: I still think the name makes sense then, but I leave it to you.

v_4_0_0/filemanager.go
kopiczko: From our chat: they are just in the main struct and iterated and referenced in the template.

PR #428, opened 2018-09-25
Use patched k8s 1.11.1 for v3.6.1 and 4.0.0

tuommaki:
Towards https://github.com/giantswarm/giantswarm/issues/4187


tuommaki:
@r7vme @corest It probably makes sense to fix `v4.0.0` image as well (as I did here)?

PR #438, opened 2018-10-12
Bring back lost change in 4.0.0 and backport to 3.6.2

r7vme:
in 4.0.0 change https://github.com/giantswarm/k8scloudconfig/pull/420 was lost also backport to 3.6.2 (which will be current WIP in operators)


r7vme:
No it's not. I'm vendoring it now to WIP releases of operators.

PR #440, opened 2018-10-17
Remove staticPodPath for workers

corest:
Towards https://github.com/giantswarm/giantswarm/issues/4447

### DIFF
--- a/v_4_0_0/files/config/kubelet-config.yaml.tmpl
+++ b/v_4_0_0/files/config/kubelet-config.yaml.tmpl
@@ -7,7 +7,6 @@ healthzPort: 10248
 clusterDNS:
   - {{.Cluster.Kubernetes.DNS.IP}}
 clusterDomain: {{.Cluster.Kubernetes.Domain}}
-staticPodPath: /etc/kubernetes/manifests
 evictionSoft:
   memory.available: "500Mi"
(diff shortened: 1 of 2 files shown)


v_4_0_0/files/config/kubelet-config-worker.yaml.tmpl
```diff
@@ -0,0 +1,24 @@
...
+evictionSoft:
+  memory.available: "500Mi"
+evictionHard:
+  memory.available: "200Mi"
+evictionSoftGracePeriod:
+  memory.available: "5s"
+evictionMaxPodGracePeriod: 60
+authentication:
+  anonymous:
+    enabled: true # Defaults to false as of 1.10
```
xh3b4sd: I see this is copy/paste but can we maybe have comments above as a usual block? This here makes it harder to read the actual YAML structure.

kopiczko:
I'd also prefer to have an `{{ if }}`.

corest:
Well, yes, I would also love to have. There is just no data for statement. Anyway, I moved it to the flag option from config and add it only in master template

kopiczko:
Isn't the flag deprecated?

corest:
Nope, it was moved into the config just to make template clearer

PR #443, opened 2018-10-26
k8s 1.12

teemow:
to make replacing the k8s version a bit easier i've factored the images
that are used multiple times into constants. the etcd image was already
different for the server and for backups.

### DIFF
--- a/v_3_7_0/cloudconfig.go
+++ b/v_3_7_0/cloudconfig.go
@@ -13,6 +13,8 @@ import (
 
 const (
 	defaultRegistryDomain = "quay.io"
+	kubernetesImage       = "giantswarm/hyperkube:v1.12.1"
+	etcdImage             = "giantswarm/etcd:v3.3.9"
 )
 
 type CloudConfigConfig struct {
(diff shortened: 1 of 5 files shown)


v_3_7_0/cloudconfig.go
```diff
@@ -13,6 +13,8 @@ import (
 
 const (
 	defaultRegistryDomain = "quay.io"
+	kubernetesImage       = "giantswarm/hyperkube:v1.12.1"
```
xh3b4sd: Where did `v1.11.1-cec4fb8023db783fbf26fb056bf6c76abfcd96cf-giantswarm` come from and why is the image here different?

teemow: Did you see the title of the PR? :smirk: 

I guess your question is not why this image is different but why the image name format looks different. The `-giantswarm` image was used because we patched the api-server memory leak in `v1.11.1`.

xh3b4sd: The second part answers my question.

PR #446, opened 2018-11-01
remove anonymous access to kubelet

teemow:
removed specific configuration for kubelet access and auth to go with
the upstream defaults.

See:
https://godoc.org/k8s.io/kubelet/config/v1beta1#KubeletConfiguration.Authentication

Also removed the readonly port that was used by heapster.


teemow:
Needs to wait for https://github.com/giantswarm/giantswarm/pull/2223

PR #450, opened 2018-11-02
remove readonly port from kubelet

teemow:
this port was used by heapster before. heapster got replaced so there is
no need to keep the port.

### DIFF
--- a/v_3_7_0/master_template.go
+++ b/v_3_7_0/master_template.go
@@ -1585,7 +1585,6 @@ write_files:
         enabled: false # Deafults to true as of 1.10
     authorization:
       mode: AlwaysAllow # Deafults to webhook as of 1.10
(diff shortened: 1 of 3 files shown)


v_3_7_0/master_template.go
```diff
@@ -1585,7 +1585,6 @@ write_files:
         enabled: false # Deafults to true as of 1.10
     authorization:
       mode: AlwaysAllow # Deafults to webhook as of 1.10
-    readOnlyPort: 10255 # Used by heapster. Defaults to 0 (disabled) as of 1.10. Needed for metrics.
```
kopiczko: Shouldn't those changes be also ported to `v_4_0_0`?

teemow: 4.0.0 is ignition. I haven't looked into that so far.

corest: I’ll port all the changes (which happened after ignition base commit) after operators will be ready. So no worries about that

PR #452, opened 2018-11-02
proxy fixes for 1.12

teemow:
* add clusterCIDR to proxy config for off-cluster bridging
* add dbus socket to proxy
* update addons kubelet to 1.12

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ version directory, and then changes are introduced.
 - Updated rbac resources to v1
 - Remove static pod path from worker nodes
 - Remove readonly port from kubelet
+- Add DBUS socket and ClusterCIDR to kube-proxy
 
 ## [v3.6.2]
 
--- a/v_3_7_0/master_template.go
+++ b/v_3_7_0/master_template.go
@@ -962,6 +962,8 @@ write_files:
               - mountPath: /etc/kubernetes/config/
                 name: config-kubernetes
                 readOnly: true
+              - mountPath: /var/run/dbus/system_bus_socket
+                name: dbus
               - mountPath: /etc/kubernetes/ssl
                 name: ssl-certs-kubernetes
                 readOnly: true
@@ -975,6 +977,9 @@ write_files:
           - hostPath:
               path: /etc/kubernetes/ssl
             name: ssl-certs-kubernetes
+          - hostPath:
+              path: /var/run/dbus/system_bus_socket
+            name: dbus
           - hostPath:
               path: /usr/share/ca-certificates
             name: ssl-certs-host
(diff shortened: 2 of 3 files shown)


teemow:
@r7vme i just stumbled upon this: https://github.com/giantswarm/installations/blob/master/godsmack/draughtsman-configmap-values.yaml#L4

`Calico.Subnet` is used a couple of times in our cloudconfig configuration. How are those values filled or replaced on Azure?

* calico-node: https://github.com/giantswarm/k8scloudconfig/blob/master/v_3_7_0/master_template.go#L220
* coredns: https://github.com/giantswarm/k8scloudconfig/blob/master/v_3_7_0/master_template.go#L569

And now I also needed it for the kube-proxy. Can we just have the correct subnet configuration in the Azure installations too or is this happen dynamically?

kopiczko:
In azure we disable calico: https://github.com/giantswarm/k8scloudconfig/blob/15bca1ca35a703f528c7dfc4ded2e1f0a5c6421c/v_3_7_0/master_template.go#L20.
And use extensions to pass custom manifests: https://github.com/giantswarm/azure-operator/blob/61b7cc933775693f7d506b2731eee611320cecbe/service/controller/v5/cloudconfig/template.go#L4.

teemow:
Yes I understood this already. 

Still this is a bit weird. 

To understand this a bit better. Do we use the private/public subnets in Azure? Or do we use them as VMs and pods in Azure.

So CoreDNS was always misconfigured in Azure? How can I configure the proxy in Azure? I haven't understood the `.Hyperkube.ControllerManager.Pod.CommandExtraArgs` concept.

teemow:
I'll merge this now to unblock the aws/kvm k8s 1.12 releases. But the azure operator will be blocked until this is fixed.

teemow:
Azure k8s 1.12 is here: https://github.com/giantswarm/azure-operator/pull/393

r7vme:
> To understand this a bit better. Do we use the private/public subnets in Azure? Or do we use them as VMs and pods in Azure.

No we don't use public/private subnets. afaik we take CIDR that was assigned for cluster and split it between master, worker and pods - https://github.com/giantswarm/azure-operator/blob/master/service/network/network.go#L21-L50

> So CoreDNS was always misconfigured in Azure? How can I configure the proxy in Azure? I haven't understood the .Hyperkube.ControllerManager.Pod.CommandExtraArgs concept.

CoreDNS uses services CIDR (not pods cidr) as far as i see in our web frontend cluster (2.0.0)

```
kubectl get cm coredns -n kube-system -o yaml
apiVersion: v1
data:
  Corefile: |
    .:53 {
        errors
        health
        kubernetes cluster.local 172.31.0.0/16  {
          pods insecure
          upstream
          fallthrough in-addr.arpa ip6.arpa
        }
        prometheus :9153
        proxy . /etc/resolv.conf
        cache 30
    }
```

Regarding `.Hyperkube.ControllerManager.Pod.CommandExtraArgs`, yes if we have a chance to get rid of it would be great, but it's more to @kopiczko as he in context i think.

teemow:
I've provided the link already above. CoreDNS is configured with the Calico CIDR: https://github.com/giantswarm/k8scloudconfig/blob/master/v_3_7_0/master_template.go#L569

> Regarding .Hyperkube.ControllerManager.Pod.CommandExtraArgs, yes if we have a chance to get rid of it would be great, but it's more to @kopiczko as he in context i think.

What is `CommandExtraArgs`? I don't want to get rid of it. I'd just like to understand how I can configure kube-proxy with the right CIDR.

PR #454, opened 2018-11-07
install kube-proxy in master before calico

teemow:
during an upgrade calico was not coming up. so the proxy stayed at 1.10
although the cluster was already 1.12. so since the proxy is not
dependent on calico we can just install it before we are installing and
waiting for calico.


calvix:
cool his might caused more issues in past :p

PR #455, opened 2018-11-07
revert-removing-pod-priority-calico

calvix:


v_3_6_2/master_template.go
```diff
@@ -153,6 +153,7 @@ write_files:
             - effect: NoExecute
               operator: Exists
           serviceAccountName: calico-node
+          priorityClassName: system-cluster-critical
```
r7vme: Please add this to calico manifest changes, otherwise it will be probably gone on nxt update.

calvix: Where is calico manifest changes?

calvix:
> Can you please provide more info where it comes?

all our critical components like coredns, kubeproxy  and others  have this to ensure that they have priority in scheduling and also that they are not evicted when there is pressure on node

v_3_6_2/master_template.go
kopiczko: Should we also add this in azure-operator manifests?

teemow: Yes definitely. We should make sure to use the same source for both imo. Do we have an issue for that already?

calvix: probably yes, I am kinda shady on where is that

calvix: using same source fro both can be tricky as azure uses completely different calico ( AFAIK)

teemow: then we maybe should put a big fat warning in here. and add a link to the one in the azure operator.

calvix: Yeah make sense

r7vme: Changes are tracked in comments here https://github.com/giantswarm/k8scloudconfig/blob/master/v_3_7_1/master_template.go#L25

We have calico in many places, so we should be careful, i found most reliable way is to track deviations from upstream manifests in comments. Better ideas are welcome

r7vme: https://github.com/giantswarm/k8scloudconfig/pull/457

PR #458, opened 2018-11-17
add 3.7.2 to fix calico startup during upgrade

teemow:
calico is a bit picky about the etcd connection and the number of
(healthy) nodes in the cluster. so this makes sure that dns to etcd is
wired properly and that the old master is removed before calico is
(re-)started.

### DIFF
--- a/v_3_7_2/master_template.go
+++ b/v_3_7_2/master_template.go
@@ -1442,9 +1442,9 @@ write_files:
           done
       done
 
-	  # wait for etcd dns (return code 35 is bad certificate which is good enough here)
+      # wait for etcd dns (return code 35 is bad certificate which is good enough here)
       while
-	      curl "https://{{ .Cluster.Etcd.Domain }}:{{ .EtcdPort }}" -k 2>/dev/null >/dev/null
+          curl "https://{{ .Cluster.Etcd.Domain }}:{{ .EtcdPort }}" -k 2>/dev/null >/dev/null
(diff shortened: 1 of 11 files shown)


corest:
We are usually copy new version first and then provide changes in next PRs. Otherwise it is impossible to find what was the actual change

teemow:
thats true :D

v_3_7_2/master_template.go
```diff
@@ -0,0 +1,2368 @@
...
+          while
+              /usr/bin/docker run -e KUBECONFIG=${KUBECONFIG} --net=host --rm -v /srv:/srv -v /etc/kubernetes:/etc/kubernetes $KUBECTL apply -f /srv/$manifest
+              [ "$?" -ne "0" ]
+          do
+              echo "failed to apply /srv/$manifest, retrying in 5 sec"
+              sleep 5s
+          done
+      done
+
+      # wait for etcd dns (return code 35 is bad certificate which is good enough here)
```
teemow: This while loop and the next for loop are the only changes I did in here.

kopiczko: We normally do a copy PR which doesn't have to be approved and then other PR with the changes. Then it's clear what was changed in the version. It's a poor man's changes history. This is important during the review but also if we want to see what was changed in a version at some point in the future.

teemow: this was already mentioned above.

PR #459, opened 2018-11-22
(addons) check etcd dns 10 times

teemow:
to make sure that etcd dns is really working we need to check more
often. we have switched back to a loadbalancer and the dns propagation
takes some time. sometimes it resolves correctly but the next one will
fail again.

### DIFF
--- a/v_3_7_2/master_template.go
+++ b/v_3_7_2/master_template.go
@@ -1443,7 +1443,7 @@ write_files:
       done
 
       # wait for etcd dns (return code 35 is bad certificate which is good enough here)
-      # since dns is flapping while we switch back to an elb for etcd we do this 10 times in a row
+      # to avoid issues with flapping dns once it is changed on an upgrade we better check 10 times in a row.
       n=0
       until [ $n -ge 10 ]
       do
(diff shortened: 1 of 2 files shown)


v_3_7_2/master_template.go
```diff
@@ -1443,13 +1443,20 @@ write_files:
       done
 
       # wait for etcd dns (return code 35 is bad certificate which is good enough here)
-      while
-          curl "https://{{ .Cluster.Etcd.Domain }}:{{ .EtcdPort }}" -k 2>/dev/null >/dev/null
-          RET_CODE=$?
-          [ "$RET_CODE" -ne "35" ]
+      # since dns is flapping while we switch back to an elb for etcd we do this 10 times in a row
```
kopiczko: I think it would be good to create a TODO issue and track removal of this code https://github.com/giantswarm/fmt/blob/master/go/todo_comments.md

teemow: I am not sure we should remove it. Any other similar change would cause the same problems again. Also it covers the time until a new master would be registered within the loadbalancer.

kopiczko: So the comment should be updated, because we won't be switching back to elb in future versions.

teemow:
```suggestion
      # to avoid issues with flapping dns once it is changed on an upgrade we better check 10 times in a row.
```

teemow:
:+1:

PR #462, opened 2018-11-30
double fs.inotify.max_user_watches

jgsqware:
Related to this PM
https://github.com/giantswarm/giantswarm/issues/4777

This PR double the inotify watches

### DIFF
--- a/v_3_7_2/master_template.go
+++ b/v_3_7_2/master_template.go
@@ -2018,6 +2018,7 @@ write_files:
   owner: root
   permissions: 0600
   content: |
+    fs.inotify.max_user_watches = 16384
     kernel.kptr_restrict = 2
     kernel.sysrq = 0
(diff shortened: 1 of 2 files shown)


jgsqware:
Is this the right version upgraded? Do I have to vendor all the operator or can it be done later as it's not a critical change?

xh3b4sd:
I am not too much into the PM but when this is a fix we want to have in all further versions it has to at least land in `v_4_0_0` because this is the folder that will be copied to make up new versions. Right now your fix is not in there.

rossf7:
@jgsqware I want to clean up the YAMLs in 3.7.2 for https://github.com/giantswarm/giantswarm/issues/4833. So I can take the copying foo and add 3.7.3.

jgsqware:
Closing in favor of https://github.com/giantswarm/k8scloudconfig/pull/478

PR #466, opened 2018-12-03
v_3_7_3: Remove params and manifests for CoreDNS and Ingress Controller

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/4833

We used params for disabling the CoreDNS and Ingress Controller manifests so we could migrate the provider operators separately. This is cleanup now all providers are migrated.

@corest I also need to remove from 4.0.0 but I'll do as a separate PR. But let me know if that causes hassle for the ignition migration.


rossf7:
Sorry I put this in for review too soon. Closing as its outdated.

PR #476, opened 2018-12-07
Allign v_4_0_0 with v3_7_4

corest:
Ignition update


v_4_0_0/cloudconfig.go
```diff
@@ -28,6 +30,16 @@ func DefaultCloudConfigConfig() CloudConfigConfig {
 	}
 }
 
+func DefaultParams() Params {
```
corest: I moved images to the DefaultParams as we need that data on the rendering, which happens earlier, than `NewCloudConfig` is called (https://github.com/giantswarm/aws-operator/pull/1199/files#diff-02eb7e1387f09a13ec4b2cca29c1ae1fR44)

PR #479, opened 2018-12-17
Restart master kubelet after running core components

corest:
The ugliest workaround you'll find today on the internet. 
But this thing just blocks me from moving with ignition and k8s 1.13. 
Upstream issue: https://github.com/kubernetes/kubernetes/issues/71078
Shortly - kubelet loses connection in the middle of booting and can't restore it. Happens only to master. Restarting container helps.


calvix:
Wow :P, just be sure that  this wont create any restart loop

PR #480, opened 2018-12-19
Remove CoreDNS and Ingress Controller manifests

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/4833

Clean up now both are migrated to chart-operator for all providers.


rossf7:
Tested on ginger.

PR #481, opened 2018-12-27
Enable IPVS in kube-proxy

corest:
Towards https://github.com/giantswarm/giantswarm/issues/4975


corest:
Tested migration with aws-operator - works fine as now config is stored in configmap

PR #482, opened 2019-01-03
Move etcdPort into default params

corest:
It is used in the rendering of the template so should be available as a default parameter

### DIFF
--- a/v_4_0_0/cloudconfig.go
+++ b/v_4_0_0/cloudconfig.go
@@ -6,8 +6,6 @@ import (
 	"encoding/base64"
 	"text/template"
 
-	"strings"
-
 	ignition "github.com/giantswarm/k8scloudconfig/ignition/v_2_2_0"
 	"github.com/giantswarm/microerror"
 )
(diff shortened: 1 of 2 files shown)


v_4_0_0/cloudconfig.go
```diff
@@ -54,14 +54,6 @@ func NewCloudConfig(config CloudConfigConfig) (*CloudConfig, error) {
 		return nil, microerror.Maskf(invalidConfigError, "config.Template must not be empty")
 	}
 
-	// Default to 443 for non AWS providers.
-	if config.Params.EtcdPort == 0 {
-		config.Params.EtcdPort = 443
-	}
-
-	// extract cluster base domain
-	config.Params.BaseDomain = strings.TrimPrefix(config.Params.Cluster.Kubernetes.API.Domain, "api.")
```
corest: This part is moved into operator (e.g. https://github.com/giantswarm/kvm-operator/blob/ignition/service/controller/v18/cloudconfig/master_template.go#L23) as it is required for rending of the templates

PR #484, opened 2019-01-10
Change calico-node log level to warn

jgsqware:


jgsqware:
Do I need to replicate this change to the other version?
Do I need to do anything else in operator, as this is not critical?

PR #485, opened 2019-01-10
Bring back missing apiserver config

corest:
If we don't specify this - api-server will listen on 8080 without auth. It was there before ignition. Just somehow deleted it by mistake during migration


corest:
@rossf7 yes, just bringing it back as was deleted by mistake

rossf7:
👍 All good just wanted to check.

PR #486, opened 2019-01-15
Update to kubernetes 1.13.2

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/4709


LICENSE
```diff
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright {yyyy} {name of copyright owner}
+   Copyright 2016 - 2019 GiantSwarm GmbH
```
rossf7: 🤣  Yes CI needs to catch this.

Gopkg.toml
```diff
@@ -26,7 +26,7 @@
 
 
 [[constraint]]
-  branch = "master"
+  branch = "kubernetes-1.13.2"
```
rossf7: TODO

LICENSE
TheoBrigitte: good catch

PR #487, opened 2019-01-28
adds worker role labels

xh3b4sd:
We were wondering why there are no worker role labels. The change here is pretty simple. What could possibly go wrong? I checked some upstream issues and only found concerns of master role labels being applied in an uncontrolled way. So we should be safe. Further I am not sure why we should have two different labels for that.

### DIFF
--- a/v_4_0_0/worker_template.go
+++ b/v_4_0_0/worker_template.go
@@ -164,7 +164,7 @@ systemd:
       --network-plugin=cni \
       --register-node=true \
       --kubeconfig=/etc/kubernetes/kubeconfig/kubelet.yaml \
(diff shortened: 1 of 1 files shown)


xh3b4sd:
What I understood is that the labels we provide here make the kubelet register itself with the labels. So this is fine, right?

tuommaki:
> What I understood is that the labels we provide here make the kubelet register itself with the labels. So this is fine, right?

Yes, exactly. The thing is just that we can't rely on this before all clusters have these and that will take little bit of time, but the direction is right here.

xh3b4sd:
Got you. I wanted to get this out so we are fine in summer or fall or something.

PR #491, opened 2019-02-01
Make render file template more robust

mazzy89:
Make `RenderFileAssetContent` function more robust.

In case an invalid template is passed in, the function panics.

This has been addressed to avoid issues like https://github.com/giantswarm/giantswarm/issues/5179

In that case an invalid template was passed in and the function was not crashing causing the cloudconfig template issue where the encrypted cert was going in.

### DIFF
--- a/v_4_0_0/cloudconfig.go
+++ b/v_4_0_0/cloudconfig.go
@@ -72,7 +72,7 @@ func (c *CloudConfig) ExecuteTemplate() error {
 	buf := new(bytes.Buffer)
 	err = tmpl.Execute(buf, c.params)
 	if err != nil {
-		return err
+		return microerror.Maskf(invalidConfigError, "cloudconfig template is not valid")
 	}
 
 	ignitionJSON, err := ignition.ConvertTemplatetoJSON(buf.Bytes())
--- a/v_4_0_0/render_asset_content.go
+++ b/v_4_0_0/render_asset_content.go
@@ -25,10 +25,7 @@ func RenderAssetContent(assetContent string, params interface{}) ([]string, erro
 
 // RenderFileAssetContent returns base64 representation of the rendered assetContent.
 func RenderFileAssetContent(assetContent string, params interface{}) (string, error) {
-	tmpl, err := template.New("").Parse(assetContent)
-	if err != nil {
-		return "", err
-	}
+	tmpl := template.Must(template.New("").Parse(assetContent))
 
 	buf := new(bytes.Buffer)
 
--- a/v_4_0_0/render_asset_content.go
+++ b/v_4_0_0/render_asset_content.go
@@ -25,7 +25,10 @@ func RenderAssetContent(assetContent string, params interface{}) ([]string, erro
 
 // RenderFileAssetContent returns base64 representation of the rendered assetContent.
 func RenderFileAssetContent(assetContent string, params interface{}) (string, error) {
-	tmpl := template.Must(template.New("").Parse(assetContent))
+	tmpl, err := template.New("").Parse(assetContent)
+	if err != nil {
+		return "", err
+	}
 
 	buf := new(bytes.Buffer)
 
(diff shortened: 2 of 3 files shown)


mazzy89:
By thinking to it twice, the unit test here does not address that specific case. 

Let's try to represent the problem:

**before the fix**
encrypted cert -> `RenderFileAssetContent` (template creation & base64 encoding) -> asset in base64

**after the fix**
encrypted cert -> ~~`RenderFileAssetContent` (template creation and parsing & base64 encoding)~~ base64 encode -> asset in base64

The problem was in the fact that the encoding cert was passed to the template function which corrupted it so the templating broke the parse. Indeed once removed from the equation the template and have only left the base64 encoding the fix worked. 

Realistically here we need to pass to the `RenderFileAssetContent` the encrypted cert and see how it behaves and make it more robust writing an unit test which covers this particular case.

mazzy89:
> We should rather find some sequence that proofs the broken parsing and find a way to detect and mitigate it in the first place.

The broken parsing here is caused by the fact that the encoded certificate has been passed in to the `RenderFileAssetContent` func which create using it a template. So the best thing would be to validate if the input passed to the function is **valid** to be used as input of the template because as in here https://github.com/giantswarm/k8scloudconfig/pull/491#issuecomment-459788586 the template function does not error in case of an invalid template so we do need to do a validation.

Any input here how to validate the input are very welcomed

xh3b4sd:
Let's maybe hang and talk about it.

mazzy89:
After a bit of debugging the function to error is this https://github.com/giantswarm/aws-operator/blob/d9b9116fad89b23700fc925e61b2a6a6007e3cc3/vendor/github.com/giantswarm/k8scloudconfig/v_4_0_0/cloudconfig.go#L66 at the line https://github.com/giantswarm/aws-operator/blob/d9b9116fad89b23700fc925e61b2a6a6007e3cc3/vendor/github.com/giantswarm/k8scloudconfig/v_4_0_0/cloudconfig.go#L75 with the following error `"template: cloudconfig:261:23: executing "cloudconfig" at <.Extension.Files>: error calling Files: template: :2: unexpected unrecognized character in action: U+05FF in command"` so we are able to reproduce the error successfully.

mazzy89:
Tested this and I was able to reproduce the error from aws-operator with the change above

```
➜ stern "aws-operator-*" -ngiantswarm | grep "cloudconfig template is not valid"
+ aws-operator-6df649b898-lcp8m › aws-operator
aws-operator-6df649b898-lcp8m aws-operator {"caller":"github.com/giantswarm/aws-operator/vendor/github.com/giantswarm/operatorkit/controller/resource/retryresource/crud_resource_ops_wrapper.go:96","controller":"aws-operator","event":"update","function":"GetDesiredState","level":"warning","loop":"6","message":"retrying due to error","object":"/apis/provider.giantswarm.io/v1alpha1/namespaces/default/awsconfigs/cxw5i","resource":"s3objectv22","stack":"[{/Users/salvo.giantswarm/go/src/github.com/giantswarm/aws-operator/vendor/github.com/giantswarm/operatorkit/controller/resource/retryresource/crud_resource_ops_wrapper.go:89: } {/Users/salvo.giantswarm/go/src/github.com/giantswarm/aws-operator/service/controller/v22/resource/s3object/desired.go:68: } {/Users/salvo.giantswarm/go/src/github.com/giantswarm/aws-operator/service/controller/v22/cloudconfig/worker_template.go:73: } {/Users/salvo.giantswarm/go/src/github.com/giantswarm/aws-operator/vendor/github.com/giantswarm/k8scloudconfig/v_4_0_0/cloudconfig.go:75: cloudconfig template is not valid} {invalid config error}]","time":"2019-02-11T17:11:56.030523+00:00","underlyingResource":"s3objectv22","version":"100420564"}
```

mazzy89:
We don't need to test any function in here because the fix https://github.com/giantswarm/aws-operator/commit/42926469601db2baafbc071b344bbe2ace8beb7f has been addressed during the creation of the cloudconfig template in the `Extension.Files` function. Closing this and addressing in aws-operator

PR #492, opened 2019-02-10
Bring resource reservation back

corest:
There was a question from shutterstock here https://gigantic.slack.com/archives/CD9A0CKB9/p1549722298204300
And I was going to say - yes, sure, we have all that reservation
But didn't find it in  our latest cloudinit. And in previous. Last version, which had resources reserved for systemd/container runtime was `3.5.2`. But during move to `3.6.0` we lost those changes https://github.com/giantswarm/k8scloudconfig/blob/master/v_3_5_2/master_template.go#L2172-L2178


corest:
Whole ignition is important. Therefore it doesn’t make sense to put some parts into tests

xh3b4sd:
So, and because everything is important you do not want to ensure that this very bug does not happen again in the future? That sounds pretty wrong to me personally. Usually it is like, something goes wrong, and we try to ensure that the thing that went wrong does not happen again. One tool for this are tests. How do you plan to improve the situation here?

corest:
No, I didn't mean this bug will not happen again. Because it is not a bug. It is a human mistake. The part we lost is static data. There are no variables. No functions called to render that piece. There is nothing to test. Yes, you can duplicate static data in test, but what are you testing then? I don't agree this particular mistake could be fixed by any test. 

Plan to improve - better tooling for managing ignition versions in the future and more care on changes.

corest:
Moved into WIP versions for cloud-init and ignition

PR #493, opened 2019-02-11
Add calicoctl and crictl tenant clusters

ferrandinand:
* Add calicoctl and crictl and aliases etcdctl
* Moved to release v_4_1_0
* Changed CHANGELOG


corest:
Can you also add etcdctl alias as in original CP ignition?

ferrandinand:
> Can you also add etcdctl alias as in original CP ignition?

Should be solved now.

calvix:
Did you test this? I think etcd LB uses port 2379

calvix:
It might even differ across platforms so we should sue some kind of variable that might be already used in k8scloudconfig

ferrandinand:
> Did you test this? I think etcd LB uses port 2379
Yes, it worked in gorgoth tenant master.

I saw in a previous PR you are removing LB in terraform, maybe we can focus on using same config so I will use local installation with 2379 port as it is done in other resources.

I need to rebase changes to put it in other vendor version, I will include this change also

v_4_0_0/master_template.go
```diff
@@ -306,6 +306,18 @@ systemd:
       ExecStartPost=/usr/bin/systemctl restart k8s-kubelet.service
       [Install]
       WantedBy=multi-user.target
+  
+  - name: debug-tools.service
+    enabled: true
+    contents: |
+      [Unit]
+      Description=Calicoctl and crictl tools
```
kopiczko: Should it be following?

```suggestion
      Description=Install calicoctl and crictl tools
```

kopiczko:
Also would be good to add an entry to `CHANGELOG.md`.

calvix:
> Yes, it worked in gorgoth tenant master.

KVM and cloud tenants can be different, the port may be different, that's why its better to use variables that are filled by operators rather than hard-coding ports.

PR #498, opened 2019-02-13
Enrich template error with stacktraces

mazzy89:
This could help to debug the still annoying issue with cloudconfig template appeared yesterday in e2e test

### DIFF
--- a/v_4_0_0/cloudconfig.go
+++ b/v_4_0_0/cloudconfig.go
@@ -66,7 +66,7 @@ func NewCloudConfig(config CloudConfigConfig) (*CloudConfig, error) {
 func (c *CloudConfig) ExecuteTemplate() error {
 	tmpl, err := template.New("cloudconfig").Parse(c.template)
 	if err != nil {
-		return err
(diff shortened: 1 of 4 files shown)


mazzy89:
I put it under `v4_0_0`. I think I should also port to `4_1_0` right?

xh3b4sd:
Towards the version, We should at least do it in the latest version. When you want to backport it further, go for it.

mazzy89:
Regarding the version, `v4_1_0` is still not used.

xh3b4sd:
The rational is basically that when you have changes in the latest version, based on the current versioning approach, somebody will copy the latest version and make a new one. When you change the penultimate version, that changes will not make it into the new versions. So what you want to have propagated should go to the latest version. Beyond that, we can decide if something should be back ported to other versions as well.

xh3b4sd:
Note that there is also `render_asset_content.go` missing. Then we have ALL errors I guess.

mazzy89:
Let's go 🚀

PR #499, opened 2019-02-15
Fix install-debug-tools script

corest:
Script fails in azure without this


corest:
Merging as I need to move with azure-operator

PR #501, opened 2019-02-16
Fix mounts in controller-manager

corest:
Add missing controller-manager mount


corest:
Adding reviewers just for information. It blocks me so merging

PR #503, opened 2019-02-21
Add a new limit of map count

pipo02mix:
Towards https://github.com/giantswarm/giantswarm/issues/5418


pipo02mix:
Tested on ginger

PR #504, opened 2019-02-25
switch kube-proxy back to iptables

teemow:
ipvs is still experimental and with 1.12.x clusters we are still having
issues with it. so for now we go back to 'iptables'


corest:
That makes me sad :(

teemow:
@corest yes but we are having real issues with it and there is no time to solve this right now. also `ipvs` is still declared as `experimental` and we have no current urge to switch it on.

PR #510, opened 2019-03-11
Added 4_0_1 for kubernetes 1.12.6

ferrandinand:
We need to create a new version for aws-operator(4.6.1) which must be to run on kubernetes version 1.12.6


ferrandinand:
I see, now PR only contents changes between package v_4_0_0  and package v_4_0_1

PR #512, opened 2019-03-22
Pin calico kube controllers to master

tuommaki:
Towards giantswarm/giantswarm#5584

Also see discussion here: https://gigantic.slack.com/archives/C3S482Z98/p1553159399026800


tuommaki:
Ok, need to create a new version and move this there. Hang on a bit.

tuommaki:
Can I ask for another look here as I moved this to another version & changed changelog entry a bit. :)

tuommaki:
Tested on `geckon`, seems to work as expected.

PR #517, opened 2019-03-29
Remove staticpodpath from worker kubelet configuration

corest:
Towards https://github.com/giantswarm/giantswarm/issues/5559

PR #518, opened 2019-03-29
v_4_2_0

calvix:
v_4_2_0


corest:
Is that synced with latest 4.1.1? It feels like we can lose some updates again

calvix:
I copied 4.1.2, so   if anyone forgets to add stuff to 4.1.2 than  i think its their mistake

corest:
Can you cherry-pick this https://github.com/giantswarm/k8scloudconfig/pull/517

calvix:
thanks

PR #521, opened 2019-03-29
Fix kubelet labels

corest:
Towads https://github.com/giantswarm/giantswarm/issues/5560

blocked by https://github.com/giantswarm/giantswarm/issues/5712

### DIFF
--- a/v_4_2_0/master_template.go
+++ b/v_4_2_0/master_template.go
@@ -264,7 +264,7 @@ systemd:
       --register-node=true \
       --register-with-taints=node-role.kubernetes.io/master=:NoSchedule \
       --kubeconfig=/etc/kubernetes/kubeconfig/kubelet.yaml \
-      --node-labels="node-role.kubernetes.io/master,role=master,ip=${DEFAULT_IPV4},{{.Cluster.Kubernetes.Kubelet.Labels}}" \
+      --node-labels="kubernetes.io/role=master,ip=${DEFAULT_IPV4},{{.Cluster.Kubernetes.Kubelet.Labels}}" \
       --v=2"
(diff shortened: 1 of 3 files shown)


v_4_2_0/master_template.go
```diff
@@ -264,7 +264,7 @@ systemd:
       --register-node=true \
       --register-with-taints=node-role.kubernetes.io/master=:NoSchedule \
       --kubeconfig=/etc/kubernetes/kubeconfig/kubelet.yaml \
-      --node-labels="node-role.kubernetes.io/master,role=master,ip=${DEFAULT_IPV4},{{.Cluster.Kubernetes.Kubelet.Labels}}" \
```
tuommaki: Same as elsewhere: We rely on these old labels here and there so it makes sense to aling those as well. In `4.1.2` I added node selection for `role=master` to `calico-kube-controllers` in here. There might be other similar pinning as well.

MarcelMue:
Can this be merged or what is missing?

corest:
@MarcelMue a lot of things are missing. In general - this issue should be solved first https://github.com/giantswarm/giantswarm/issues/5712

MarcelMue:
Ack so this is also blocking 1.14 ? 
I am trying to figure out if we need to merge this first or if I can create a new k8scc WIP for 1.14 and see what else we need to do

corest:
It is not blocking 1.14 if we don't change labels. But 1.14 is the latest version where those labels are deprecated. With 1.15 kubelet will not be able to start. So I think it is better to fix this now. 
If 4.2.0 is not used within active releases - you can use that one for 1.14

PR #522, opened 2019-03-30
Fix systemd race issue

calvix:
fix race condition issues with systemd dependency in units see https://github.com/giantswarm/giantswarm/issues/5692

This  mainly to help fix flaky e2e test where the race condition appeared more often

also backporting to older version to have all e2e fixed ( both WIP and CUR tests)

### DIFF
--- a/v_4_1_1/master_template.go
+++ b/v_4_1_1/master_template.go
@@ -84,15 +84,13 @@ systemd:
     contents: |
       [Unit]
       Description=k8s-setup-network-env Service
-      Wants=network.target docker.service
-      After=network.target docker.service
+      Wants=network.target docker.service wait-for-domains.service
+      After=network.target docker.service wait-for-domains.service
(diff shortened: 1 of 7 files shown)


v_4_1_1/master_template.go
```diff
@@ -84,15 +84,13 @@ systemd:
...
-      After=network.target docker.service
+      Wants=network.target docker.service wait-for-domains.service
+      After=network.target docker.service wait-for-domains.service
       [Service]
       Type=oneshot
-      RemainAfterExit=yes
       TimeoutStartSec=0
       Environment="IMAGE={{.Cluster.Kubernetes.NetworkSetup.Docker.Image}}"
       Environment="NAME=%p.service"
-      Environment="NETWORK_CONFIG_CONTAINER="
```
calvix: not really needed, legacy

PR #523, opened 2019-04-08
Add new version 4.3.0

MarcelMue:
Adding a new WIP for the upcoming k8s 1.14

Towards https://github.com/giantswarm/giantswarm/issues/5688

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,15 @@ The latest version is considered WIP and it is a subject of change. All other
 versions are frozen. To freeze current version all files are copied to a new
 version directory, and  then changes are introduced.
 
-## [v4.2.0] WIP
+## [v4.3.0] WIP
+
+## [v4.2.0]
+
+### Changed
 - Fix race condition issue with systemd units.
+
+### Removed
+
 - Remove `UsePrivilegeSeparation` option from sshd configuration.
 
 ## [v4.1.2]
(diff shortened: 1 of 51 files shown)


CHANGELOG.md
```diff
@@ -9,8 +9,15 @@ The latest version is considered WIP and it is a subject of change. All other
 versions are frozen. To freeze current version all files are copied to a new
 version directory, and  then changes are introduced.
 
-## [v4.2.0] WIP
+## [v4.3.0] WIP
```
MarcelMue: Is it correct to change this information here? 
Or is this usually done in a separate PR?

v_4_3_0/cloudconfig.go
```diff
@@ -0,0 +1,101 @@
+package v_4_3_0
```
MarcelMue: Copied v_4_2_0 and fixed imports.

xh3b4sd:
So far we did the copy PRs separate as well as the import changes.

MarcelMue:
@xh3b4sd I agree with operators I just looked at the last copy PR and it was done at the same time ( probably because there is no wiring here, just a replace)
I can provide separate PRs if it's important to you.

mazzy89:
We usually do the process in two different PRs:

i.e.

first PR
copy version 4.1.0-> 4.2.0
second PR
fix all the imports

and for both of these PRs you don't need any approvals

mazzy89:
At the end I quote @xh3b4sd and for that PR now that it is already in place could work. The most important thing is to not forget the WIP in the README and in this case is there so all good for me

PR #524, opened 2019-04-09
k8s 1.14 release

MarcelMue:
Adding new k8s 1.14.1 

Should we also update calico? There is 3.6.1 for calico.

### DIFF
--- a/v_4_3_0/files/k8s-resource/calico-all.yaml
+++ b/v_4_3_0/files/k8s-resource/calico-all.yaml
@@ -1,0 +1,514 @@
+# CALICO HAS SEPARATE MANIFEST FOR AZURE
+# the azure manifest can be found in: https://github.com/giantswarm/azure-operator/blob/master/service/controller/vX/cloudconfig/template.go
+# where X is the version of azure operator
+#
+# Extra changes:
+#  - Added resource limits to calico-node and calico-kube-controller.
+#  - Added resource limits to install-cni.
+#  - Added 'priorityClassName: system-cluster-critical' to calico daemonset.
+#
+# Calico Version v3.5.1
+# https://docs.projectcalico.org/v3.2/releases#v3.5.1
+# This manifest includes the following component versions:
+#   calico/node:v3.5.1
+#   calico/cni:v3.5.1
+#   calico/kube-controllers:v3.5.1
+
+# This ConfigMap is used to configure a self-hosted Calico installation.
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: calico-config
+  namespace: kube-system
+data:
+  # Configure this with the location of your etcd cluster.
+  etcd_endpoints: "https://{{ .Cluster.Etcd.Domain }}:{{ .EtcdPort }}"
+
+  # If you're using TLS enabled etcd uncomment the following.
+  # You must also populate the Secret below with these files.
+  etcd_ca: "/calico-secrets/client-ca.pem"
+  etcd_cert: "/calico-secrets/client-crt.pem"
+  etcd_key: "/calico-secrets/client-key.pem"
+  # Configure the Calico backend to use.
+  calico_backend: "bird"
+
+  # Configure the MTU to use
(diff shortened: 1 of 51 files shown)


corest:
Did you test it already? Any deprecations, flag updates, default feature gates, resource api changes (moving from alpha/beta etc)? I never saw that small change related to new k8s version

MarcelMue:
No testing next, I wanted to get some reviews if this is the correct approach to this / if we want to update more like calico.

Probably shouldn't have pinged this many people in hindsight, sorry.

corest:
An approach is correct. Also, as a suggestion for testing - I'm usually updating k8scloudconfig directly in operator for latest WIP and test it in that way to skip vendoring on every update. When change is properly tested - then cherry-picking it to k8scloudconfig repo and request review. And yes, calico should be updated as well and that requires careful testing

v_4_3_0/files/k8s-resource/calico-all.yaml
```diff
@@ -351,7 +351,7 @@ spec:
         k8s-app: calico-kube-controllers
     spec:
       nodeSelector:
-        beta.kubernetes.io/os: linux
+        kubernetes.io/os: linux
```
mazzy89: Just for transparency, could you put here the source?

is it needed for calico upgrade?

mazzy89: I've found it in the deprecation list 

> OS and Arch information is now recorded in kubernetes.io/os and kubernetes.io/arch labels on Node objects. The previous labels (beta.kubernetes.io/os and beta.kubernetes.io/arch) are still recorded, but are deprecated and targeted for removal in v1.18. (#73333, @yujuhong)

mazzy89:
> NetworkPolicy resources will no longer be served from extensions/v1beta1 in v1.16. Migrate use to the networking.k8s.io/v1 API, available since v1.8. Existing persisted data can be retrieved via the networking.k8s.io/v1 API.

Finally, something moves around `Ingress`. We might consider changing it. I need to look if the TC creates any ingress resources. I think so because we spin up there nginx. If so I'd vote to change it.

mazzy89:
> PriorityClass resources have been promoted to scheduling.k8s.io/v1 with no changes. The scheduling.k8s.io/v1beta1 and scheduling.k8s.io/v1alpha1 versions are now deprecated and will stop being served by default in v1.17. (#73555, #74465, @bsalamat)

Do we create any priority class for our TC?

v_4_3_0/files/k8s-resource/calico-all.yaml
calvix: we use https://docs.projectcalico.org/v3.2/releases#v3.6.1 as source for calico manifests and than just update when they do

mazzy89: I'm going to review the labels here because in KVM `kube-calico-controllers` does not go up due to some labels misalignment 

```
Events:
  Type     Reason            Age                  From               Message
  ----     ------            ----                 ----               -------
  Warning  FailedScheduling  83s (x16 over 6m6s)  default-scheduler  0/4 nodes are available: 4 node(s) didn't match node selector.
```

calvix: yeah, i thnik Tuomas recently introduced label node selector so it sits on master

mazzy89: I've personally verified the upstream calico manifests for the version 3.6.1 and no major changes here.

tuommaki: > yeah, i thnik Tuomas recently introduced label node selector so it sits on master

Yeah and that was one of the things that had to be refined as we dropped the old `role=master` label from node. That might cause the problem here?

MarcelMue:
@mazzy89 How confident are we to merge this now?

MarcelMue:
RFR Please take a look here, this should be fine now.

mazzy89:
Rever taint for master.

Conformance tests failed and could not pass.

PR #525, opened 2019-04-19
Allow kube-proxy to tolerate all NoSchedule taint

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/5744

We encountered an issue where `kube-proxy` was not scheduled on some node due to them having the following taint:
```
 taints:
 - effect: NoSchedule
   key: NodeWithImpairedVolumes
   value: "true"
```

Therefore we need to have `kube-proxy` tolerating every `NoSchedule` taint and not only the ones with `key: node-role.kubernetes.io/master`.

How about aligning toleration with calico and adding toleration for `CriticalAddonsOnly` and `NoExecute` also ?
https://github.com/giantswarm/k8scloudconfig/blob/80b22e313bd60f651c8f24456e366931ef2a347d/v_4_3_0/files/k8s-resource/calico-all.yaml#L120-L128

### DIFF
--- a/v_4_3_0/files/k8s-resource/calico-all.yaml
+++ b/v_4_3_0/files/k8s-resource/calico-all.yaml
@@ -118,14 +118,11 @@ spec:
         beta.kubernetes.io/os: linux
       hostNetwork: true
       tolerations:
-        # Make sure calico-node gets scheduled on all nodes.
-        - effect: NoSchedule
-          operator: Exists
         # Mark the pod as a critical add-on for rescheduling.
         - key: CriticalAddonsOnly
           operator: Exists
-        - effect: NoExecute
-          operator: Exists
+        # Make sure the pod gets scheduled on all nodes.
+        - operator: Exists
       serviceAccountName: calico-node
       priorityClassName: system-cluster-critical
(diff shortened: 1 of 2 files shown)


TheoBrigitte:
Found out that kube-proxy manifest from kubernetes repo has even more aggressive toleration in order to tolerate all taints. 
> As a essential core component, kube-proxy should generally run on all
nodes even if the cluster operator taints nodes for special purposes.
> [source](https://github.com/kubernetes/kubernetes/commit/bb28449e31ee17cd183b1606fef3f3e3aa40de58)
```
      tolerations:
      - key: CriticalAddonsOnly
        operator: Exists
      - operator: Exists
```

JosephSalisbury:
@TheoBrigitte we want this for calico too, no?

TheoBrigitte:
> @TheoBrigitte we want this for calico too, no?

Good point. I updated calico to have the same taint as kube-proxy.

So they now both tolerate every taint effect and taint with `key:CriticalAddonsOnly`. PTALA

MarcelMue:
You merged this without a changelog entry :(
We are currently using this code branch for 1.14.1 and are about to release it. Can you add something to the changelog?
https://github.com/giantswarm/k8scloudconfig/blob/master/CHANGELOG.md

PR #535, opened 2019-05-06
Add felix metric server for expose calico metrics

pipo02mix:
Expose Felix metrics in calico, so we or customers can use Prometheus to scrape Felix state.


mazzy89:
Wrong version. 4.3.0 is already active

PR #539, opened 2019-06-05
Add labels for core components

corest:


JosephSalisbury:
Possibly also towards https://github.com/giantswarm/giantswarm/issues/5982

corest:
Added also `tier=control-plane` label. Not sure about key, but that what is used in GKE.

PR #540, opened 2019-06-11
k8s 1.14.3 release

MarcelMue:
towards https://github.com/giantswarm/giantswarm/issues/6109

### DIFF
--- a/v_4_4_0/files/k8s-resource/calico-all.yaml
+++ b/v_4_4_0/files/k8s-resource/calico-all.yaml
@@ -619,7 +619,6 @@ spec:
         - name: etcd-certs
           hostPath:
             path: /etc/kubernetes/ssl/etcd
-            mode: 0400
 
 ---
 
--- a/v_4_4_0/files/k8s-resource/calico-all.yaml
+++ b/v_4_4_0/files/k8s-resource/calico-all.yaml
@@ -7,12 +7,12 @@
 #  - Added resource limits to install-cni.
 #  - Added 'priorityClassName: system-cluster-critical' to calico daemonset.
 #
-# Calico Version v3.6.1
-# https://docs.projectcalico.org/v3.2/releases#v3.6.1
+# Calico Version v3.7.2
+# https://docs.projectcalico.org/v3.7/release-notes/
 # This manifest includes the following component versions:
-#   calico/node:v3.6.1
-#   calico/cni:v3.6.1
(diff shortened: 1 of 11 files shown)


mazzy89:
@MarcelMue have you checked that no changes have been introduced in the new manifests for calico `3.7.x`?

MarcelMue:
Not yet.

mazzy89:
Ok then please check that the manifests we have don't diverge with the new manifests because we run those ones. I've checked the changelong.

v_4_4_0/files/k8s-resource/calico-all.yaml
```diff
@@ -120,11 +120,14 @@ spec:
         kubernetes.io/os: linux
       hostNetwork: true
       tolerations:
+        # Make sure calico-node gets scheduled on all nodes.
```
MarcelMue: These tolerations were changed upstream, not sure if it makes a big difference.

v_4_4_0/files/k8s-resource/calico-all.yaml
```diff
@@ -501,6 +505,8 @@ rules:
     verbs:
       # Needed for clearing NodeNetworkUnavailable flag.
       - patch
+      # Calico stores some configuration information in node annotations.
```
MarcelMue: There are a few more RBAC changes but it seems related to TYPHA which we don't use IIRC

v_4_4_0/files/k8s-resource/calico-all.yaml
```diff
@@ -420,6 +423,7 @@ spec:
         - name: etcd-certs
           hostPath:
             path: /etc/kubernetes/ssl/etcd
+            defaultMode: 0400
```
MarcelMue: Upstream uses a secret for this, trying to change the mode with our mounted file as well :man_shrugging:

mazzy89:
I've noticed that `ConfigMap` has this property:

```
  # Typha is disabled.
  typha_service_name: "none"
```

I would add it to the config. WDYT?

mazzy89:
Tested and seems all good. @MarcelMue we can go ahead

mazzy89:
I'm going to merge it and then in a different PR update the changelog and create a new WIP @MarcelMue

MarcelMue:
Ack - let's go @mazzy89  :+1:

PR #541, opened 2019-06-11
Use new owner structure

corest:
This PR aligns Owner structure with ignition spec and allows to use `id` of the group instead of name. Required for this https://github.com/giantswarm/azure-operator/pull/519

### DIFF
--- a/v_4_4_0/master_template.go
+++ b/v_4_4_0/master_template.go
@@ -544,9 +544,17 @@ storage:
     - path: {{ .Metadata.Path }}
       filesystem: root
       user:
-        name: {{ .Metadata.Owner.User }}
+      {{- if .Metadata.Owner.User.ID }}
+        id: {{ .Metadata.Owner.User.ID }}
+      {{- else }}
+        name: {{ .Metadata.Owner.User.Name }}
+      {{- end }}
       group:
-        name: {{ .Metadata.Owner.Group }}
+      {{- if .Metadata.Owner.Group.ID }}
+        id: {{ .Metadata.Owner.Group.ID }}
+      {{- else }}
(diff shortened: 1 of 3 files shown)


v_4_4_0/types.go
```diff
@@ -97,8 +97,18 @@ type FileMetadata struct {
...
+	Group Group
+	User  User
+}
+
+type Group struct {
+	ID   int
+	Name string
+}
+
+type User struct {
```
kopiczko: I'd like to have a godoc that ID and Name are mutually exclusive and if both are set then ID is preferred.

v_4_4_0/master_template.go
```diff
@@ -544,9 +544,17 @@ storage:
     - path: {{ .Metadata.Path }}
       filesystem: root
       user:
-        name: {{ .Metadata.Owner.User }}
+      {{- if .Metadata.Owner.User.ID }}
+        id: {{ .Metadata.Owner.User.ID }}
+      {{- else }}
+        name: {{ .Metadata.Owner.User.Name }}
```
kopiczko: Do you have a plan to get rid of Name at some point or we need to keep both?

corest: No, both of those params are supported in ignition spec, so I’d like to keep them (specially for cases like the one we have now in external-dns - same names in coreos on host system and alpine container have different id, so you can’t reference ownership with a name as then container user will not be able to get that data. So you need to use id of non-existing user on host

MarcelMue:
Vendoring this in `aws-operator` breaks it. What needs to be done to get it working again? 

I can take a look tomorrow if it's not a lot of work.

PR #543, opened 2019-06-13
Harden SSH

mazzy89:
Towards https://github.com/giantswarm/giantswarm/issues/6070

Same thing applied for CP.

Notice here that the forwarding is disabled both for workers and masters.

It is not needed


mazzy89:
tested on godsmack. it works. merging here.

PR #549, opened 2019-06-28
add ceph/rbd storage support config

calvix:
for https://github.com/giantswarm/dvag/issues/21


v_4_5_0/master_template.go
```diff
@@ -208,7 +208,7 @@ systemd:
       EnvironmentFile=/etc/network-environment
       Environment="IMAGE={{ .RegistryDomain }}/{{ .Images.Kubernetes }}"
       Environment="NAME=%p.service"
-      Environment="NETWORK_CONFIG_CONTAINER="
```
calvix: this env was useless a super legacy

tuommaki:
Is this PR complete? I see only few env changes for `$PATH` - nothing `ceph/rbd` specific besides changelog. Did I miss something?

calvix:
Yes, this is just necessary changes and the RBD itself is installed via `ceph-install` pod when needed and it installs files in `/opt/bin`.

 If we decide we want to include all files in all installations we can do it later but this is hard requirement.

PR #551, opened 2019-06-28
Add namespaces labels

corest:

### DIFF
--- a/v_4_5_0/files/conf/k8s-addons
+++ b/v_4_5_0/files/conf/k8s-addons
@@ -10,11 +10,11 @@ KUBECTL={{ .RegistryDomain }}/giantswarm/docker-kubectl:f5cae44c480bd797dc770dd5
 while [ "$(/usr/bin/docker run -e KUBECONFIG=${KUBECONFIG} --net=host --rm -v /etc/kubernetes:/etc/kubernetes $KUBECTL get cs | grep Healthy | wc -l)" -ne "3" ]; do sleep 1 && echo 'Waiting for healthy k8s'; done
 
 # label namespaces (required for network egress policies)
-NAMESPACES="kube-system giantswarm"
+NAMESPACES="default giantswarm kube-system" 
 for namespace in ${NAMESPACES}
 do
     while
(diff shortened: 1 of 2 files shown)


v_4_5_0/files/conf/k8s-addons
```diff
@@ -9,6 +9,19 @@ KUBECTL={{ .RegistryDomain }}/giantswarm/docker-kubectl:f5cae44c480bd797dc770dd5
 # wait for healthy master
 while [ "$(/usr/bin/docker run -e KUBECONFIG=${KUBECONFIG} --net=host --rm -v /etc/kubernetes:/etc/kubernetes $KUBECTL get cs | grep Healthy | wc -l)" -ne "3" ]; do sleep 1 && echo 'Waiting for healthy k8s'; done
 
+# label namespaces (required for network egress policies)
+NAMESPACES="kube-system giantswarm"
```
pipo02mix: Why not label `default` too? (in case customers can add netpol for that)

v_4_5_0/files/conf/k8s-addons
```diff
@@ -9,6 +9,19 @@ KUBECTL={{ .RegistryDomain }}/giantswarm/docker-kubectl:f5cae44c480bd797dc770dd5
...
+for namespace in ${NAMESPACES}
+do
+    while
+        /usr/bin/docker run -e KUBECONFIG=${KUBECONFIG} --net=host --rm -v /etc/kubernetes:/etc/kubernetes $KUBECTL label namespace ${namespace} name=${namespace}
```
pipo02mix: Out of curiosity, why do you need `host` network to run kubectl?

corest: I don't remember the reason for this. Not needed here for sure. Just copy-pasted from commands below

v_4_5_0/files/conf/k8s-addons
corest: I don't care about customer's needs :trollface: 
updating...

PR #552, opened 2019-07-01
Drop giantswarm namespace from labeling

corest:
Namespace is not there for a time of k8s-addons execution


corest:
Yep, I can add labeling there

PR #553, opened 2019-07-05
Fix labeling loop

corest:
Another round :/

### DIFF
--- a/v_4_5_0/files/conf/k8s-addons
+++ b/v_4_5_0/files/conf/k8s-addons
@@ -14,11 +14,16 @@ NAMESPACES="default kube-system"
 for namespace in ${NAMESPACES}
 do
     while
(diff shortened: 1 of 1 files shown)


v_4_5_0/files/conf/k8s-addons
```diff
@@ -14,11 +14,16 @@ NAMESPACES="default kube-system"
 for namespace in ${NAMESPACES}
 do
     while
```
kopiczko: I think this should become `if`.

xh3b4sd: I think we should stop using bash magic where possible. 🙈

corest: It is while for the network glitches case, when `get namespaces` fails just because of api being unresponsive

kopiczko: Does it matter? `$?` will be non-zero anyway.

PR #559, opened 2019-07-09
Make image pull progress deadline configurable for kubelets

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/6367

Makes `--image-pull-progress-deadline` configurable for kubelets with a default of 1 minute. We will increase this to 10 minutes for tenant clusters in AWS China regions. This mitigates the extremely slow image pulls we see there which are outside of our control.

This happened in control plane clusters a long time ago and we decided to implement for tenants as well.

> --image-pull-progress-deadline duration
>  If no pulling progress is made before this deadline, the image pulling will be cancelled. (default 1m0s)

https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,11 @@ version directory, and  then changes are introduced.
 
 ## [v4.6.0] WIP
 
+### Changed
+
+- Make --image-pull-progress-deadline configurable for kubelets so a longer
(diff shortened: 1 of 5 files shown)


rossf7:
Tested on godsmack.

PR #560, opened 2019-07-10
Make sure GiantSwarm user has id 1000

corest:
This is related to issue with node-exporter, where it mounts volumes as user `1000`, but on some nodes, this user has a different ID. So with this, we will explicitly create it with ID 1000.

After all clusters will be upgraded, we will remove a user from here https://github.com/giantswarm/installations/blob/master/default-draughtsman-configmap-values.yaml#L87

### DIFF
--- a/v_4_5_0/master_template.go
+++ b/v_4_5_0/master_template.go
@@ -4,9 +4,6 @@ const MasterTemplate = `---
 ignition:
   version: "2.2.0"
 passwd:
-  groups:
-    - name: giantswarm
-      gid: 1000
   users:
     - name: giantswarm
       shell: "/bin/bash"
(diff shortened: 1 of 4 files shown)


corest:
User is there already, so it just ensures the user has a proper id.

corest:
Will make one more test on aws (tested in `kvm`) and merge then with vendoring into operators

kopiczko:
> User is there already, so it just ensures the user has a proper id.

Then this is a fix.



> After all clusters will be upgraded, we will remove a user from here https://github.com/giantswarm/installations/blob/master/default-draughtsman-configmap-values.yaml#L87

Do we need to track it?

corest:
@kopiczko https://github.com/orgs/giantswarm/projects/9

PR #564, opened 2019-07-17
Custom cgroup structure

corest:
Towards https://github.com/giantswarm/giantswarm/issues/6270

### New `kubereserved` cgroup.
- `docker` daemon moved from root cgroup into kubereserved. 
- `kubelet` and `etcd` systemd services moved under `kubereserved` cgroup. This is not that important, as containers itself are placed into cgroups by docker daemon configuration. But this change was done to keep related things better structured
- with `--parent-cgroup` configuration, all the containers, started by docker without k8s context (e.g. kubelet/etcd) are started under `kubereserved` cgroup.
 
### Explicit reservation of the resources via kubelet config.

- `SystemReserved` configuration explicitly points to `system` cgroup and has explicitly reserved resources.
- `Kubereserved` configuration has explicitly reserved resources.

This means, that there will be less allocatable resources for nodes:
```
[Allocatable] = [Node Capacity] - [Kube-Reserved] - [System-Reserved]
```

Allocation enforcement was not changed. That means, that resource allocation forced only for pods (`kubepods` cgroup). If kubereserved/systemdreserved cgroup will need more memory - they can take more. But for pods it is enforced to stay in `Allocatable` limits

Exact numbers were taken from best practices docs, discussions on GitHub and some performance tests on test clusters. We can adjust those numbers over time while testing in real production.

----

New cgroup structure looks like:

```
+-- -.slice
+-- system.slice. Reserved: **250m** CPU, **384Mi** memory
|   +-- sshd
|   +-- systemd
|   +-- udevd
|   +-- networkd
|   +-- ...
+-- kubereserved.slice (new cgroup). Reserved(master/worker): **400m**/**300** CPU, **1280Mi**/**768Mi** memory
|   +-- docker daemon
|   +-- containerd
|   +-- docker containers, started out of k8s context, e.g. kubelet, etcd
+-- user.slice - no reserved resources explicitly
|   +-- user sessions
+-- kubepods
|   +-- besteffort
|   +-- burstable
|   +-- guaranteed 
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ version directory, and  then changes are introduced.
 ### Changed
 
 - Mount relevant directories so that the command `docker` can run in `Kubelet`. This is needed for `rbd` to mount `Ceph` volumes on the nodes.
+- Add explicit cgroups for better grained resources management over operating system components and container runtime.
 
 ### Fixed
 
--- a/v_4_6_0/files/config/kubelet-master.yaml.tmpl
+++ b/v_4_6_0/files/config/kubelet-master.yaml.tmpl
@@ -16,15 +16,15 @@ evictionSoftGracePeriod:
   memory.available: "5s"
 evictionMaxPodGracePeriod: 60
 kubeReserved:
-  cpu: 500m
-  memory: 512Mi
+  cpu: 250m
+  memory: 768Mi
   ephemeral-storage: 1024Mi
 kubeReservedCgroup: /podruntime.slice
 kubeletCgroups: /podruntime.slice
 runtimeCgroups: /podruntime.slice
 systemReserved:
-  cpu: 500m
-  memory: 512Mi
+  cpu: 250m
+  memory: 384Mi
 systemReservedCgroup: /system.slice
 authentication:
   anonymous:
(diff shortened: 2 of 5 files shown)


corest:
I've tried to move `user` cgroup under `system`, so we can control this as well under SystemReserved scope. But there are a lot of dependencies under that structure, so I've just added approximate number of resources into SystemReserved cgroup

CHANGELOG.md
```diff
@@ -17,7 +17,8 @@ version directory, and  then changes are introduced.
 
 ### Changed
 
-- Mount relevant directories so that the command `docker` can run in `Kubelet`. This is needed for `rbd` to mount `Ceph` volumes on the nodes.
+- Mount relevant directories so that the command `docker` can run in `Kubelet`. This is needed for `rbd` to mount `Ceph` volumes on the nodes.h
```
rossf7: My knowledge is shallow here. What is `rbd`?
```suggestion
- Mount relevant directories so that the command `docker` can run in `Kubelet`. This is needed for `rbd` to mount `Ceph` volumes on the nodes.
```

CHANGELOG.md
```diff
@@ -17,7 +17,8 @@ version directory, and  then changes are introduced.
...
+- Add explicit cgroups for better grained resources management over operating system components and container runtime.
```
rossf7:
```suggestion
- Add explicit cgroups for finer grained resources management over operating system components and container runtime.
```

Minor 🇺🇸 nit but this flows better for me.

PR #566, opened 2019-07-23
Harden default restricted PSP

corest:
Towards https://github.com/giantswarm/giantswarm/issues/5128


corest:
Life is a tough thing

PR #570, opened 2019-07-29
Node exporter systemd unit

corest:
Moving node-exporter from container into systemd unit. 

This will help us
 - solve all the issues related with access to OS info.
 - keeping CoreOS and node-exporter versions in sync (another thing which causes issues every now and then)

---

Unit is using little hack to copy binary from container so we can solve distribution issue for China.
With next PR I'll remove node-exporter from current WIP version in cluster-operator.
Would be nice to sync this back to control-plane

### DIFF
--- a/v_4_6_0/master_template.go
+++ b/v_4_6_0/master_template.go
@@ -325,6 +325,51 @@ systemd:
       ExecStopPost=-/usr/bin/docker rm -f $NAME
       [Install]
       WantedBy=multi-user.target
+  - name: node-exporter.service
+    enabled: true
+    contents: |
+      [Unit]
+      Description=Node Exporter
+
+      [Service]
+      ExecStartPre=/bin/sh -c "docker rm -f node-exporter-binarycopy; \
+                  docker create --name node-exporter-binarycopy {{ .RegistryDomain }}/node-exporter:v0.18.0-giantswarm && \
+                  docker cp node-exporter-binarycopy:/bin/node_exporter /opt/bin/node_exporter && \
+                  docker rm -f node-exporter-binarycopy"
+      ExecStart=/opt/bin/node_exporter \
+                --log.level=info \
+                --web.listen-address=:10300 \
+                --collector.arp \
+                --collector.bcache \
+                --collector.conntrack \
+                --collector.cpu \
+                --collector.edac \
+                --collector.entropy \
+                --collector.filefd \
+                --collector.filesystem \
+                --collector.hwmon \
+                --collector.loadavg \
+                --collector.mdadm \
+                --collector.meminfo \
+                --collector.netdev \
+                --collector.netstat \
+                --collector.sockstat \
+                --collector.stat \
+                --collector.systemd \
+                --collector.time \
+                --collector.timex \
+                --collector.uname \
+                --collector.vmstat \
+                --collector.xfs \
+                --no-collector.diskstats \
+                --no-collector.infiniband \
+                --no-collector.ipvs \
+                --no-collector.textfile \
+                --no-collector.wifi \
+                --no-collector.zfs
+
+      [Install]
+      WantedBy=multi-user.target
   - name: etcd2.service
     enabled: false
     mask: true
--- a/v_4_6_0/master_template.go
+++ b/v_4_6_0/master_template.go
@@ -330,6 +330,8 @@ systemd:
     contents: |
       [Unit]
       Description=Node Exporter
+      After=k8s-setup-network-env.service docker.service
+      Requires=k8s-setup-network-env.service docker.service
 
       [Service]
       ExecStartPre=/bin/sh -c "docker rm -f node-exporter-binarycopy; \
--- a/v_4_6_0/master_template.go
+++ b/v_4_6_0/master_template.go
@@ -334,6 +334,8 @@ systemd:
       Requires=k8s-setup-network-env.service docker.service
 
       [Service]
+      Restart=on-failure
+      RestartSec=5s
       ExecStartPre=/bin/sh -c "docker rm -f node-exporter-binarycopy; \
                   docker create --name node-exporter-binarycopy {{ .RegistryDomain }}/node-exporter:v0.18.0-giantswarm && \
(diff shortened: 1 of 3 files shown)


v_4_6_0/master_template.go
```diff
@@ -323,6 +323,51 @@ systemd:
       --v=2"
       ExecStop=-/usr/bin/docker stop -t 10 $NAME
       ExecStopPost=-/usr/bin/docker rm -f $NAME
+      [Install]
+      WantedBy=multi-user.target
+  - name: node-exporter.service
+    enabled: true
+    contents: |
+      [Unit]
+      Description=Node Exporter
```
mazzy89: On worker we require `docker.service` and other units before to start. here no. any reasons for that?

corest: typo

calvix:
maybe we could add some Restart policy for the service so it can restart in case of failure

we should also check if we somehow scrape node-exporter via api proxy to pods or if we simply scrape via node and rewrite port

If we scrape via API proxy we can have issues with firewall rules, not sure

MarcelMue:
Not blocking here:
I understand the issue that is solved with this, but shouldn't components be kept in sync in the future with the OS version because of the managed apps approach? Or it should be a lot easier to keep it in sync at least.

I am asking because this feels like we are creating a snowflake service here which we deploy differently from all others (e.g. all other exporters). I would prefer to stay aligned but like I said - no blocker for me now.

corest:
Yep, we scrapi via api-proxy. That can be an issue here

JosephSalisbury:
I get it, but having to roll clusters to upgrade node-exporter feels super weird

corest:
@JosephSalisbury I don't agree. We have a lot of issues because of operating system changes, which break node-exporter. So having different node-exporter versions deployed on different coreos versions just doesn't work. There was an issue about that https://github.com/giantswarm/giantswarm/issues/6428. So keeping those components in sync makes more sense to me

corest:
> I am asking because this feels like we are creating a snowflake service here which we deploy differently from all others (e.g. all other exporters). 

@MarcelMue node-exporter for me is more infrastructure component, than k8s component. Running it as docker container is weird. It requires a lot of information about system, therefore isolation here is useless. 

At the same time, I agree with the fact, that having it in daemonset simplifies management

corest:
Will wait for more thoughts. As I see here pros and cons and it might be still better to just run it as root in container and manage all those volumes and root fs hacks

teemow:
We've had this discussion a couple of times and I was always torn to move it to a systemd unit as well as there are so many weird workarounds to make it work. So I like this change. Yes of course it would be nice to have everything within the cluster and within containers. But as long as it isn't really supported and suggested upstream let's not overdo this.

That means node-exporter is dealt with similarly like the kubelet, etcd but also not that much differently than calico, kube-proxy and other k8s-addons in k8scloudconfig in terms of lockstep with the OS and tenant releases. So this isn't a unique way to manage node-exporter.

But do we need to adjust some of the rules in prometheus as well for this? Don't we have some kind of mapping to work around the container issues atm? eg mounts. Something like that: https://github.com/giantswarm/g8s-prometheus/blob/master/helm/g8s-prometheus-chart/prometheus-rules/disk.rules.yml#L41

kopiczko:
> Don't we have some kind of mapping to work around the container issues atm? eg mounts. Something like that: https://github.com/giantswarm/g8s-prometheus/blob/master/helm/g8s-prometheus-chart/prometheus-rules/disk.rules.yml#L41

This should be `(/rootfs)?` as everywhere else.

kopiczko:
I don't understand how it helps:

> keeping CoreOS and node-exporter versions in sync (another thing which causes issues every now and then)

This is <provider>-operator that decides which version of OS is running in the cluster. So this is decoupled.

Isn't it also enough to change directory permissions on the host and add r+x access to giantswarm group for /var/lib/kubelet?

corest:
It is enough to change permissions. But that’s why so didn’t reference my PM even. It is just unnatural to run it in container. Current approach we’re using is a hack itself. Hack which eg helps us to scrape target in easier way.
And now we can do this in proper way. Yes, it involves a lot of changes and probably dups in alerts for handling both approaches in parallel for migration period. But is just a proper way of using this. 

Anyway, if that’s a thing I need to convince all the people again in - I’m fine to abandon this and fix PM with permissions. It is even easier for me

kopiczko:
I won't block it. I understand this is the recommended way of running this thing. There are also benefits or running this in DS even if it's hackery.  I also don't see how it improves version sync between OS and node-exporter. But I don't have strong opinion here and I don't want to discourage you.

corest:
Majority decided this change will make node-exporter management too hard. Closing.

teemow:
Just out of interest. Where was that decided?

corest:
In sig-infrastructure sync

teemow:
@corest alright. Btw the kube-prometheus project also runs the node-exporter in a pod. Check their parameters for the filesystems. This would make much more sense imo.

https://github.com/coreos/kube-prometheus/blob/master/manifests/node-exporter-daemonset.yaml#L20

corest:
@teemow our only issues for now (for now as we every time have some during updates) is running as non-root. So, sorting permissions

teemow:
I disagree about the "only" issue. But yeah maybe that is the issue you have in your postmortem :)

teemow:
But sorting permissions sounds dangerous. So you open up permissions to node-exporter only I hope. Otherwise you make things readable for everyone that aren't supposed to be readable. What is better then? Running node-exporter as root or making everything readable? Would be good to discuss in sig-security first.

corest:
@teemow it is a complex thing. I open permissions for giantswarm group, which is then used to run node-exporter. 

> Otherwise you make things readable for everyone that aren't supposed to be readable.

It is readable for user `giantswarm`, which we are using anyway. In terms of containers - yes, you can run your container as `giantswarm` user and get the same access. But that's also the same as writing your own psp and running container as root

teemow:
As said please get consensus in @giantswarm/sig-security  about this. Don't convince only me :)

FYI the `giantswarm` user is in sudoers on the host. And it has a shell and is accessible from the outside. I don't think this is a good idea to re-use that user.

PR #571, opened 2019-07-30
Set kubelet folder ownership to root:giantswarm

corest:
Towards https://github.com/giantswarm/giantswarm/issues/6622

### DIFF
--- a/v_4_6_0/master_template.go
+++ b/v_4_6_0/master_template.go
@@ -52,16 +52,19 @@ systemd:
     contents: |
       {{range .Content}}{{.}}
       {{end}}{{end}}
-  - name: set-certs-group-owner-permission-giantswarm.service
+  - name: set-giantswarm-user-ownership.service
     enabled: true
     contents: |
       [Unit]
-      Description=Change group owner for certificates to giantswarm
+      Description=Set required for giantswarm user ownership
       Wants=k8s-kubelet.service k8s-setup-network-env.service
       After=k8s-kubelet.service k8s-setup-network-env.service
       [Service]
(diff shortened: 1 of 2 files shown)


v_4_6_0/master_template.go
```diff
@@ -52,16 +52,19 @@ systemd:
...
       [Unit]
-      Description=Change group owner for certificates to giantswarm
+      Description=Set required for giantswarm user ownership
       Wants=k8s-kubelet.service k8s-setup-network-env.service
       After=k8s-kubelet.service k8s-setup-network-env.service
       [Service]
       Type=oneshot
-      ExecStart=/bin/sh -c "find /etc/kubernetes/ssl -name '*.pem' -print | xargs -i  sh -c 'chown root:giantswarm {} && chmod 640 {}'"
+      # wait for kubelet dirs being created
+      ExecStartPre=/bin/sleep 30
```
calvix: can we rather do check if the dir is already with some loop and sleep, this can easily fail on some envs ( like in china pulling image is slow and can definitely take more than 30s

corest: I don't think this service will be running before `k8s-kubelet.service` actually starts (goes through all StartPre and run ExecStart). And pulling is done in StartPre, so this ownership service can't be activated before

corest: anyway added loop

corest:
That's a good catch. I'm pretty sure it sets new pods dirs with root

corest:
We will need to switch to root user :/

corest:
This will not work

PR #572, opened 2019-08-08
Kubernetes 1.14.5 

mazzy89:
Towards https://github.com/giantswarm/giantswarm/issues/6701

### DIFF
--- a/v_4_5_1/cloudconfig.go
+++ b/v_4_5_1/cloudconfig.go
@@ -1,0 +1,101 @@
+package v_4_5_0
+
+import (
+	"bytes"
+	"compress/gzip"
+	"encoding/base64"
+	"text/template"
+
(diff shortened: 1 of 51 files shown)


v_4_5_1/cloudconfig.go
```diff
@@ -0,0 +1,101 @@
+package v_4_5_0
```
schjan: Shouldnt the package name be v_4_5_1 for every file?

mazzy89: Amazing catch. thanks a lot

mazzy89:
@schjan Now I'm going to leave this opened and I'll merge once we have tested all the operators. Going to open a PR on aws-operator so you can then mirror it in another operator like azure

PR #579, opened 2019-08-21
Remove IC binding to privileged PSP

corest:
Towards https://github.com/giantswarm/giantswarm/issues/6765#event-2572506429


MarcelMue:
Do we need something in the changelog for this?

corest:
I’d say no. Nothing changed. One PR remove privileged binding, another - adds missing privilege in psp

PR #580, opened 2019-08-29
Use struct instead of interface for estension rendering

MarcelMue:
This allows us to clean up old TODO's (e.g. https://github.com/giantswarm/giantswarm/issues/4329) and makes rendering a bit more straight forward. So far we implicitly called function when rendering - instead we should just have a data structure which is passed to the rendering section.

Am I missing anything obvious here?

### DIFF
--- a/v_4_8_0/master_template_test.go
+++ b/v_4_8_0/master_template_test.go
@@ -24,11 +24,7 @@ func Test_MasterTemplate(t *testing.T) {
 
 	// Test with Params struct. This should contain all evaluated fields.
 	{
-		params := Params{
-			// Extension has to be set because it's interface and
-			// template evaluation will panic otherwise.
-			Extension: nopExtension{},
-		}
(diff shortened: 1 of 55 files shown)


v_4_8_0/master_template_test.go
```diff
@@ -24,11 +24,7 @@ func Test_MasterTemplate(t *testing.T) {
 
 	// Test with Params struct. This should contain all evaluated fields.
 	{
-		params := Params{
-			// Extension has to be set because it's interface and
-			// template evaluation will panic otherwise.
-			Extension: nopExtension{},
```
kopiczko: I guess you can also drop `nopExtension` type.

kopiczko:
Hey @MarcelMue look what I found :)

MarcelMue:
Dropping it as I thought I had already closed it :D

PR #583, opened 2019-09-26
Remove Get List Watch from audit logs per default

jgsqware:
It generate a lot of logs, not necessary useful.
I think we can drop it as those verbs are more related to having access to the cluster.

### DIFF
--- a/v_4_8_0/files/policies/audit-policy.yaml
+++ b/v_4_8_0/files/policies/audit-policy.yaml
@@ -109,7 +109,7 @@ rules:
     omitStages:
       - "RequestReceived"
   # Get repsonses can be large; skip them.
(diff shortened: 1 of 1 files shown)


pipo02mix:
yeah agree we need to do configurable by customer/installation. For my customer so far removing these actions are ok

tuommaki:
@jgsqware Can you make a roadmap story out of this (with configurability) and assign it to Team Batman?

corest:
@jgsqware is this still a thing? Can you create a separate story for batman?

corest:
@jgsqware is this still a thing or can be closed?

xh3b4sd:
I think this can be closed.

PR #584, opened 2019-09-27
Remove limits from calico-node DaemonSet

tuommaki:
Recently calico-node started OOM on several installations' tenant clusters and
on control plane. While the root cause is still unknown, crashing calico-node
occasionally breaks all workload from the cluster requiring POD recreations in
order to resume normal operation. This is very bad from a core component such
as Calico so replacing hard limits with monitoring & alert is less disruptive
way to reach same outcome.

When calico memory usage grows into unreasonable levels, we should investigate
further and manually fix the situation at hand, but not let the cluster go into
fully broken state due to automatic actions taken by container limits & kernel.

Limits from existing clusters will be removed now as well.

Towards https://github.com/giantswarm/giantswarm/issues/7022


tuommaki:
**Question:** How can we ensure that when we update Calico in the future, that we remember to remove the limits from deployment manifest?

calvix:
the main reason why calico has limits is that kubelet put it into  guaranteed class, and it does not get evicted

 Not sure if this can cause any issues, @puja108  ?

calvix:
changes that deviates from calico original manifest are usually summed here
https://github.com/giantswarm/k8scloudconfig/blob/master/v_4_8_0/files/k8s-resource/calico-all.yaml#L1-L8

tuommaki:
Crap: https://kubernetes.io/docs/tasks/configure-pod-container/quality-service-pod/#create-a-pod-that-gets-assigned-a-qos-class-of-guaranteed

> - Every Container in the Pod must have a memory limit and a memory request, and they must be the same.
> - Every Container in the Pod must have a CPU limit and a CPU request, and they must be the same.

So I guess we have to change them instead of removing. Memory from `150MB` -> `250MB` then? I don't think I've seen anything over `250` during this week when looking at prom metrics for this.

tuommaki:
...though we do have `priorityClassName: system-cluster-critical` on the daemonsets. That should do the required job here.

corest:
@tuommaki nope, it will not do the required job. QoS is still a thing https://kubernetes.io/docs/tasks/administer-cluster/out-of-resource/#evicting-end-user-pods

tuommaki:
Ok. Two different things, it seems. `system-cluster-critical` causes the POD to get evicted, but immediately re-scheduled where as having limits and getting into `Guaranteed` class, it delays eviction in the first place by killing lower priority stuff first.

So conclusion: We need limits.

teemow:
@tuommaki but we can create very high limits correct. So we still avoid running into OOMs too fast.

tuommaki:
Closing in favor of https://github.com/giantswarm/k8scloudconfig/pull/585

mazzy89:
This issue has been reopened and this is the way we go. Removing limits and create an ad-hoc alert

MarcelMue:
As discussed we should first test this on a CP which is affected before we touch tenantcluster related parts. Therefor k8scc is currently not needed IMO.

mazzy89:
Yep @MarcelMue I remember 😇 Just updating this task all over the places to communicate our decisions to all the subscribers to this ticket. Then we won't merge this now.

corest:
But still, let's not change it everywhere. Only WIP version

teemow:
Yes please. *Never* change an existing active release. We promise that those releases are immutable over time. This will get "fixed" with voo in terms of having a git tag instead of folders. But until then please never touch the folders once they are used in a published release.

mazzy89:
Closing this in favor of this https://github.com/giantswarm/k8scloudconfig/pull/597

PR #585, opened 2019-09-27
Increase memory limit for calico-node

tuommaki:
Due to recent OOMing in `calico-node` increase memory limit accordingly to
avoid problems. Crashing calico-node can potentially (happened few times) take
cluster workload into permanent CrashLoopBackOff requiring POD recreation after
calico network gets stable again.

Initial plan was to remove limits altogether, but that disables Guaranteed QoS
class from calico which then yields different problems if other workload on a
node is taking too much resources. Having Guaranteed QoS delays eviction to
later phase, effectively killing normal workload first and this behavior is
crucial for core component such as calico.

### DIFF
--- a/v_4_0_0/files/k8s-resource/calico-all.yaml
+++ b/v_4_0_0/files/k8s-resource/calico-all.yaml
@@ -215,7 +215,7 @@ spec:
               memory: 150Mi
             limits:
               cpu: 250m
-              memory: 150Mi
+              memory: 250Mi
           livenessProbe:
             httpGet:
               path: /liveness
--- a/v_4_0_0/files/k8s-resource/calico-all.yaml
+++ b/v_4_0_0/files/k8s-resource/calico-all.yaml
@@ -215,7 +215,7 @@ spec:
               memory: 150Mi
             limits:
               cpu: 250m
-              memory: 500Mi
+              memory: 150Mi
           livenessProbe:
             httpGet:
               path: /liveness
--- a/v_4_0_0/files/k8s-resource/calico-all.yaml
+++ b/v_4_0_0/files/k8s-resource/calico-all.yaml
@@ -212,7 +212,7 @@ spec:
           resources:
             requests:
               cpu: 250m
-              memory: 150Mi
+              memory: 250Mi
             limits:
               cpu: 250m
               memory: 250Mi
--- a/v_4_0_0/files/k8s-resource/calico-all.yaml
+++ b/v_4_0_0/files/k8s-resource/calico-all.yaml
@@ -212,10 +212,10 @@ spec:
           resources:
             requests:
               cpu: 250m
-              memory: 250Mi
+              memory: 150Mi
             limits:
               cpu: 250m
-              memory: 250Mi
+              memory: 500Mi
           livenessProbe:
             httpGet:
               path: /liveness
--- a/v_4_7_0/files/k8s-resource/calico-all.yaml
+++ b/v_4_7_0/files/k8s-resource/calico-all.yaml
@@ -460,7 +460,7 @@ spec:
               memory: 150Mi
             limits:
               cpu: 250m
-              memory: 150Mi
+              memory: 250Mi
           livenessProbe:
             httpGet:
               path: /liveness
--- a/v_4_7_0/files/k8s-resource/calico-all.yaml
+++ b/v_4_7_0/files/k8s-resource/calico-all.yaml
@@ -457,7 +457,7 @@ spec:
           resources:
             requests:
               cpu: 250m
-              memory: 150Mi
+              memory: 250Mi
             limits:
               cpu: 250m
               memory: 250Mi
--- a/v_4_7_0/files/k8s-resource/calico-all.yaml
+++ b/v_4_7_0/files/k8s-resource/calico-all.yaml
@@ -457,10 +457,10 @@ spec:
           resources:
             requests:
               cpu: 250m
-              memory: 250Mi
+              memory: 150Mi
             limits:
               cpu: 250m
-              memory: 250Mi
+              memory: 500Mi
           livenessProbe:
             httpGet:
               path: /liveness
--- a/v_4_0_1/files/k8s-resource/calico-all.yaml
+++ b/v_4_0_1/files/k8s-resource/calico-all.yaml
@@ -215,7 +215,7 @@ spec:
               memory: 150Mi
             limits:
               cpu: 250m
-              memory: 150Mi
+              memory: 250Mi
           livenessProbe:
             httpGet:
               path: /liveness
--- a/v_4_0_1/files/k8s-resource/calico-all.yaml
+++ b/v_4_0_1/files/k8s-resource/calico-all.yaml
@@ -215,7 +215,7 @@ spec:
               memory: 150Mi
             limits:
               cpu: 250m
-              memory: 500Mi
+              memory: 150Mi
           livenessProbe:
             httpGet:
               path: /liveness
--- a/v_4_0_1/files/k8s-resource/calico-all.yaml
+++ b/v_4_0_1/files/k8s-resource/calico-all.yaml
@@ -212,7 +212,7 @@ spec:
           resources:
             requests:
               cpu: 250m
-              memory: 150Mi
+              memory: 250Mi
             limits:
               cpu: 250m
               memory: 250Mi
--- a/v_4_0_1/files/k8s-resource/calico-all.yaml
+++ b/v_4_0_1/files/k8s-resource/calico-all.yaml
@@ -212,10 +212,10 @@ spec:
           resources:
             requests:
               cpu: 250m
-              memory: 250Mi
+              memory: 150Mi
             limits:
               cpu: 250m
-              memory: 250Mi
+              memory: 500Mi
           livenessProbe:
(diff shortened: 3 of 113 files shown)


v_4_0_0/files/k8s-resource/calico-all.yaml
```diff
@@ -215,7 +215,7 @@ spec:
               memory: 150Mi
             limits:
               cpu: 250m
-              memory: 150Mi
+              memory: 250Mi
```
tuommaki: This is ~67% increase should be plenty enough. Normally it only seldom goes above `150MB`, though I admit that prometheus scrape interval of 30s doesn't help with this. Nevertheless, so far manual adjustment of memory limit to `250MB` has removed the problem so hence I'm rather confident with this. The downside of having too large limit is that it prevents usage of small instance types because they cannot hold all workload. This overhead is then also present on every node.

Following screenshot is past week from `asgard`:
Screenshot from 2019-09-27 12-19-30

tuommaki:
And I just realized I almost failed here. Need to adjust requests as well... Changing...

corest:
Why do we change it everywhere? With this we change currently running clusters. And that's confusing

tuommaki:
> Why do we change it everywhere? With this we change currently running clusters. And that's confusing

Because I'm manually also patching currently running clusters and we don't want new clusters be created with low limits that won't work.

Do you think that only active releases should be changed?

kopiczko:
Yeah @corest is right here.

kopiczko:
> Do you think that only active releases should be changed?

I think yes. With VOO we won't even have an option.

corest:
I wouldn't even go with only active. Because that makes whole point about having release useless. As with master machine restarts on some clusters only we will have cluster with the same versions but different state.
@giantswarm/team-ludacris ping. Maybe we need to have patch release?

teemow:
How about increasing the limit to a much higher value?

teemow:
We absolutely don't want calico to OOM.

teemow:
And +1 on the patch release. Please never ever change an already active release.

corest:
@teemow increasing just limit makes this workload losing its QoS class. Increasing both requests/limits to high numbers again makes consumption higher, which again needs to be explained to customers

tuommaki:
> How about increasing the limit to a much higher value?

This requires then once again substantially larger master instances. We already did this once: https://github.com/giantswarm/kubernetesd/pull/449 but it's up to inventory if this is enough or not.

> And +1 on the patch release. Please never ever change an already active release.

Ok. I'll do that. Next week then. What about then the existing clusters?

teemow:
> @teemow increasing just limit makes this workload losing its QoS class. Increasing both requests/limits to high numbers again makes consumption higher, which again needs to be explained to customers

I know it switches from Guaranteed to Burstable. Where is the problem?

> This requires then once again substantially larger master instances

No it doesn't need a bigger master. Only if you create higher requests.

> What about then the existing clusters?

They need to get upgraded. As always. If you talk to the customer because a cluster is in trouble you can also manually change the value.

tuommaki:
> I know it switches from Guaranteed to Burstable. Where is the problem?

Before we made it `Guaranteed`, customer workload that didn't have proper limits, caused kubelet to evict workload and calico is then one of the potential targets to get killed. AFAIK this has happened often before we got this. If we can live with `Bustable`, we can remove limits altogether, but based on current understanding, this is not feasible.

> They need to get upgraded. As always. If you talk to the customer because a cluster is in trouble you can also manually change the value.

So effectively we won't be proactive here. I'm sort of good with this at least for now, because I ran out of time and it's not clear what to do.

corest:
> I know it switches from Guaranteed to Burstable. Where is the problem?

It affects the eviction order. Should not be the issue for the case with those high priorities configured. But as anything else requires some tests

teemow:
Yeah this should be done via priority classes.

tuommaki:
> Yeah this should be done via priority classes.

Based on my understanding, it's not the same thing. See here: https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/

We have currently that priority class, but
> Note that marking a pod as critical is not meant to prevent evictions entirely; it only prevents the pod from becoming permanently unavailable. For static pods, this means it can’t be evicted, but for non-static pods, it just means they will always be rescheduled.

My interpretation here is that it would get into restart / eviction loop. Definitely requires testing at least.

teemow:
https://kubernetes.io/docs/tasks/administer-cluster/out-of-resource/#daemonset

> It is never desired for kubelet to evict a DaemonSet Pod, since the Pod is immediately recreated and rescheduled back to the same node.

corest:
> Instead DaemonSet should ideally launch Guaranteed Pods

From the same doc

teemow:
Lol and reading further: 

> In general, it is strongly recommended that DaemonSet not create BestEffort Pods to avoid being identified as a candidate Pod for eviction. Instead DaemonSet should ideally launch Guaranteed Pods.

The only other option I see is that with the higher requests of the pod 250m we stay under the threshold and then the pod is not evicted. 

> Guaranteed pods and Burstable pods whose usage is beneath requests are evicted last. Guaranteed Pods are guaranteed only when requests and limits are specified for all the containers and they are equal. Such pods are guaranteed to never be evicted because of another Pod’s resource consumption. If a system daemon (such as kubelet, docker, and journald) is consuming more resources than were reserved via system-reserved or kube-reserved allocations, and the node only has Guaranteed or Burstable Pods using less than requests remaining, then the node must choose to evict such a Pod in order to preserve node stability and to limit the impact of the unexpected consumption to other Pods. In this case, it will choose to evict pods of Lowest Priority first.

teemow:
So it boils down to:

 * High limit, higher request, alert that matches request? => Pod is not evicted or very unlikely. If it grows further the likelihood its getting evicted is higher. 
 * higher limit and request but equal, alert that matches request? => Pod is not likely to evicted at all. But will be killed once the limit is reached.

tuommaki:
AFAIK k8s scheduling works purely on requests so that still affects instance types / how much workload fits on the same machine.

Also: When we have limit, why alert?

teemow:
> AFAIK k8s scheduling works purely on requests so that still affects instance types / how much workload fits on the same machine.

You anyway wanted to raise the requests to 250m. So there is no difference between both options. My point was just, that the burstable pod is less likely to get evicted if it stays underneath the request. Which will be the case since we anyway have to raise the request in the guaranteed option as well.

> Also: When we have limit, why alert?

Yes we should probably alert before we reach the limit. Calico is a critical pod and we should take care before they start crash looping.

v_4_0_0/files/k8s-resource/calico-all.yaml
```diff
@@ -215,7 +215,7 @@ spec:
               memory: 150Mi
             limits:
               cpu: 250m
-              memory: 150Mi
+              memory: 500Mi
```
tuommaki: Do we now have consensus on keeping requests (for scheduling) as-is but moving limit to `500MB`? Before merge I'll then figure out which ones are actually needed for required patch releases and copy those.

teemow: I thought requests: 250 and limit almost unlimited. Where does the 150 and 500 come from now?

tuommaki: `150MB` is the current limit which *most* the time is perfectly fine. It's only occasional peak that runs above it. `500MB` is ~233% more than what is normally enough so for me this is a lot. I don't understand reasoning from something like e.g. `4GB` limit for component like this. What purpose would it have?

teemow: The purpose is that we don't want calico to get killed. If there is a short spike for whatever reason calico should not get killed.

If the calico pods stays under 150MB for 99% of the time I am fine with that. If we now are more often between 150MB and 250MB then we should consider raising requests as this prevents eviction.

puja108:
> So it boils down to:
> 
> High limit, higher request, alert that matches request? => Pod is not evicted or very unlikely. If it grows further the likelihood its getting evicted is higher.
> higher limit and request but equal, alert that matches request? => Pod is not likely to evicted at all. But will be killed once the limit is reached.

But what is the difference then between high limit and no limit?

teemow:
@puja108 you are right. Both will be burstable and there is no difference. 

So we can remove the limit and then need to check how often the actual memory consumption will be higher than memory request, because this might lead to eviction.

v_4_7_0/files/k8s-resource/calico-all.yaml
```diff
@@ -460,7 +460,7 @@ spec:
               memory: 150Mi
             limits:
               cpu: 250m
-              memory: 150Mi
+              memory: 500Mi
```
kopiczko: Was that introduced in some previous commit?

tuommaki: No, this was a leftover from my earlier "reset all" command. PM currently on-hold in our team.

mazzy89:
Closing this and reopening this https://github.com/giantswarm/k8scloudconfig/pull/584 after a team discussion along with Timo where we decided that we are going to remove limits and create an ad-hoc alarms for this calico component

PR #587, opened 2019-10-01
Kubelet configmap

ferrandinand:
Towards: https://github.com/giantswarm/giantswarm/issues/4889

This PR allows to create dynamic configurations for kubelet as mentioned here: 

3 Steps need to be done to use dynamic kubelet configuration.
1.- Create kubelet configmaps for master and workers once the API is available. Executed once in master in addons service.
    https://github.com/giantswarm/k8scloudconfig/pull/587/files#diff-bc06789e9db2a732ce261d4b602cfe8fR65
2.- Start kubelets as usual without configmaps(there is no api connection yet). 
   https://github.com/giantswarm/k8scloudconfig/pull/587/files#diff-a503fa1db86f84fce8eb93e5e36bb677R313
3.- We patch all nodes to use configmaps instead of local configs.  Created a new service to do this when a node bootstraps with some templating.
https://github.com/giantswarm/k8scloudconfig/pull/587/files#diff-51cf641c8fe17e6f747aae1dbe7eb144R1
templating to patch
https://github.com/giantswarm/k8scloudconfig/pull/587/files#diff-819c3db2b80b99ddcc637bce82be913b

Some implications:
- Files can be unsync if we use different config templates for local bootstrap and configmaps.
- If we use same file template for local and configmap we need to drop some flags specific to nodes.


mazzy89:
Where was residing before the kubelet configuration? just wondering the `ConfigMap` valeus from where they come from?

ferrandinand:
Here: https://github.com/giantswarm/k8scloudconfig/blob/master/v_4_8_0/files/config/kubelet-master.yaml.tmpl
And here: https://github.com/giantswarm/k8scloudconfig/blob/master/v_4_8_0/files/config/kubelet-worker.yaml.tmpl

Anyway I will provide more info tomorrow when oppening the PR after more testing

mazzy89:
Ok i'll hold on.

v_4_8_0/files/config/kubelet-configmap.yaml.tmpl
```diff
@@ -0,0 +1,80 @@
+apiVersion: v1
+kind: ConfigMap
+metadata: 
+  name: kubelet-configuration-master-values
+  namespace: kube-system
+data:
+  kubelet: |
```
kopiczko: Do we have now to keep this and fallback configuration in sync? How is it done?

corest:
I got the "how" part. And I know there is a reason e.g. with local-dns. 

But can you describe typical workflow for this change to use? E.g. how this would be applied via coredns change? Just to be sure we're not doing any manual patches. Can this, for example, be done via user-config cm for coredns?

v_4_8_0/files/config/kubelet-configmap.yaml.tmpl
ferrandinand: No sync atm, they are different resources. If we are not fine with this we can reduce configmap configs to the minimum values required.

ferrandinand:
The idea in this PR is just about using dynamic kubelet configs for all use of cases.
In the case of local dns the idea would be to create a feature flag to install ds and other resources if enabled but I would keep kubelet configMap patch out of the app.

But user should be able to change the kubelet config just editing the configMap.

I am thinking on some other case of use. e.g Increasing the kubelet resources or eviction policies per cluster. So I don´t think we would like to mix configMap patch in the coredns chart.

v_4_8_0/files/config/kubelet-configmap.yaml.tmpl
kopiczko: It would be good to have at least bold comment in both places that those things need to be in sync.

kopiczko:
What happens when the kublet restarts and the CM is changed? Will it take it from etcd or local disk?

v_4_8_0/files/config/kubelet-configmap.yaml.tmpl
ferrandinand: Yes, a comment would be the minimum but this is an interesting topic if it would be better to have a full kubelet config or just the overrides.

ferrandinand:
It gets a local copy of the cm in what is called LastGoodKnown configuration so it can be started with last working config without API connection.

It is better explained here: 
https://kubernetes.io/docs/tasks/administer-cluster/reconfigure-kubelet/#understanding-how-the-kubelet-checkpoints-config

v_4_8_0/files/config/kubelet-configmap.yaml.tmpl
kopiczko: I can't tell which is better.

kopiczko: Maybe we should discuss in the SIG?

ferrandinand: My bad, you cannot select specific fields to override. A valid cm overrides a full config file.
https://github.com/giantswarm/opsctl/pull/409

corest: This brings some overhead. Why can't we at least render both configs from the same template?

ferrandinand: Do you mean refactor worker/master with same template passing different values depending if it is worker or master or creating a new template for configMap part and then there partially render the kubelet config template inside?

corest: I mean why can’t ypu create configmap from local kubelet config instead of rendering separate cm yaml?

ferrandinand: Understood I see the improvement there.
I will see how to integrate a template in the template and if it is possible.
 
```
apiVersion: v1
kind: ConfigMap
metadata: 
  name: kubelet-configuration-master-values
  namespace: kube-system
data:
  kubelet: |
```Render a template ehere```
```

corest: No, not render into cm
But create cm from file as we do for kube-proxy

kubectl create configmap kubelet --from-file=<local config>

You can add it into the kubelet config rendering service as a second command

ferrandinand: I got you now. Nice idea.
Thanks

v_4_8_0/files/conf/k8s-addons
```diff
@@ -161,3 +174,4 @@ do
     done
 done
 echo "Addons successfully installed"
+
```
kopiczko:
```suggestion
```

v_4_8_0/files/conf/k8s-addons
```diff
@@ -62,6 +62,19 @@ do
     sleep 3s
 done
 
+# create dynamic kubelet configmap
+for i in {worker,master}
```
kopiczko:
```suggestion
for role in {worker,master}
```

v_4_8_0/files/conf/k8s-addons
```diff
@@ -62,6 +62,19 @@ do
...
+do
+  KUBELET_CONFIGMAP="/etc/kubernetes/config/kubelet-${i}.yaml"
```
kopiczko:
```suggestion
  KUBELET_CONFIGMAP="/etc/kubernetes/config/kubelet-${role}.yaml"
```

v_4_8_0/files/config/kubelet-worker.yaml.tmpl
```diff
@@ -1,8 +1,6 @@
 kind: KubeletConfiguration
 apiVersion: kubelet.config.k8s.io/v1beta1
-address: ${DEFAULT_IPV4}
 port: 10250
-healthzBindAddress: ${DEFAULT_IPV4}
```
kopiczko: Will it be configured by coredns?

ferrandinand: These are configurations for kubelet and link node ips with kubelet.
They are configurations created per node in the local template but if we want to use the same template here we need to remove these configs and use them as flags becase each node will have his own IP and we cannot define them in the same configmap.

corest:
This patching looks suspicious a bit. Let's discuss this during infra sync as PR introduces so much indirect logic

v_4_8_0/files/config/kubelet-worker.yaml.tmpl
corest: I don't get this. Can we create configmap after local file is rendered? So that you don't have complex logic and multiple additional files rendered?

E.g. create configmap here https://github.com/giantswarm/k8scloudconfig/pull/587/files#diff-a503fa1db86f84fce8eb93e5e36bb677R102 instead of k8s-addons?

ferrandinand: That logic is executed on each node once it gets bootstrapped so we don´t need to create one config map per node, just one configmap for masters and one for workers. 

I liked to put it as an addon because it is executed once one master bootstrap and it is part of the kubernetes config at startup and It will reside in kube-system namespace.

The problem when using the same template for local and configmap is not all values are shared.So if we want to use the same template we need to drop some configs and put them as flags(some implications are here)

Refering to the node patch to use the configmap, we can use the node patch command literally in the service but I saw it very hacky to use something like this instead a template.
```kubectl patch node ${NODE_NAME} -p "{\"spec\":{\"configSource\":{\"configMap\":{\"name\":\"${CONFIG_MAP_NAME}\",\"namespace\":\"kube-system\",\"kubeletConfigKey\":\"kubelet\"}}}}"``` 

But if using this, we remove 2 templates and its rendering command.

**Comment by ferrandinand** [comment]
Yes, this is not an easy to adapt feature. It has some implications if we want to use it but it gives us something new as well.

Let´s talk about it in the SIG and agree which direction we must take. I put some notes in the description to allow everybody to get fast into the conversation.

File: v_4_8_0/master_template.go [hunk 6]
```diff
@@ -309,7 +310,11 @@ systemd:
       {{ . }} \
       {{ end -}}
       --node-ip=${DEFAULT_IPV4} \
-      --config=/etc/kubernetes/config/kubelet.yaml \
+      --address=${DEFAULT_IPV4} \
+      --healthz-bind-address=${DEFAULT_IPV4} \
```
kopiczko: I see. This is here now.

calvix:
old PR, closing

calvix:
if needed should be part of team story not SIG issue

PR #588, opened 2019-10-02
Update k8s, etcd, calico, debug-tools and CHANGELOG

tfussell:
Closing in favor of https://github.com/giantswarm/k8scloudconfig/pull/590


v_4_8_0/files/k8s-resource/calico-all.yaml
```diff
@@ -462,19 +660,19 @@ spec:
...
               memory: 150Mi
           livenessProbe:
-            httpGet:
-              path: /liveness
-              port: 9099
-              host: localhost
+            exec:
+              command:
+              - /bin/calico-node
+              - -felix-live
```
MarcelMue: Not sure about this one.

v_4_8_0/files/k8s-resource/calico-all.yaml
```diff
@@ -263,29 +264,226 @@ spec:
...
+  - apiGroups: [""]
+    resources:
+      - nodes/status
+    verbs:
+      # Needed for clearing NodeNetworkUnavailable flag.
+      - patch
+      # Calico stores some configuration information in node annotations.
+      - update
+  # Watch for changes to Kubernetes NetworkPolicies.
+  - apiGroups: ["networking.k8s.io"]
```
MarcelMue: Hard to tell what changed in these because they show up in the git diff weirdly.

v_4_8_0/files/k8s-resource/calico-all.yaml
```diff
@@ -42,17 +42,17 @@ data:
   cni_network_config: |-
     {
       "name": "k8s-pod-network",
-      "cniVersion": "0.3.0",
+      "cniVersion": "0.3.1",
       "plugins": [
         {
           "type": "calico",
           "log_level": "info",
+          "nodename": "__KUBERNETES_NODE_NAME__",
```
MarcelMue: Why move this from alphabetical ordering?

tfussell: Main goal was to keep it as close to https://docs.projectcalico.org/v3.9/manifests/calico.yaml as possible. That way we can compare changes in newer versions using `diff`.

v_4_8_0/files/k8s-resource/calico-all.yaml
tfussell: Sorry about that. The only change AFAIK is adding `blockaffinities` CRD access below.

v_4_8_0/files/k8s-resource/calico-all.yaml
tfussell: I'm not either. Shows up in [v3.9](https://docs.projectcalico.org/v3.9/manifests/calico.yaml). [v3.8](https://docs.projectcalico.org/v3.8/manifests/calico.yaml) uses the old `httpGet` way. I can do some research on pros and cons. Should this also be discussed with SIG-Infra?

PR #596, opened 2019-10-14
Use a systemd unit for kubelet instead of running in a container

tfussell:
Towards https://github.com/giantswarm/giantswarm/issues/6498

The method for getting the `kubelet` binary onto nodes is not yet decided. For now I am just trying to copy over from the docker image since it's already configured and known to be accessible from all machines. [Discussion is here](https://gigantic.slack.com/archives/C3S482Z98/p1571053603048600).

kubelet `--feature-gates` flag is deprecated in 1.16 so I also took care of that.


v_4_9_0/worker_template.go
```diff
@@ -138,15 +138,32 @@ systemd:
       ExecStopPost=-/usr/bin/docker rm -f $NAME
       [Install]
       WantedBy=multi-user.target
+  - name: k8s-setup-download-hyperkube.service
```
MarcelMue: Note: We want to change this later on, for now this is the simplest thing to do.

MarcelMue:
@giantswarm/sre Can we get some opinions on here? 

We found this to be the simplest way for now but uncertain of correctness / sensibility.

v_4_9_0/worker_template.go
kopiczko: Why is it bad? What are the ideas?

MarcelMue: This causes us to pull the entire hyperkube image just to then only use the binaries. It would be much better to only pull the binaries.

tfussell: It's mainly bad because Docker isn't intended as a software distribution platform. Alternatives include downloading binaries directly from Google (might have issues on nodes in China), baking it into our AMIs (not sure how this works on Azure/KVM), uploading/downloading via S3 (maybe proxied through giantswarm.io?).

v_4_9_0/master_template.go
```diff
@@ -322,8 +297,7 @@ systemd:
       --kubeconfig=/etc/kubernetes/kubeconfig/kubelet.yaml \
       --node-labels="node.kubernetes.io/master,node-role.kubernetes.io/master,kubernetes.io/role=master,role=master,ip=${DEFAULT_IPV4},{{.Cluster.Kubernetes.Kubelet.Labels}}" \
       --v=2"
-      ExecStop=-/usr/bin/docker stop -t 10 $NAME
-      ExecStopPost=-/usr/bin/docker rm -f $NAME
+      ExecStop=-/usr/bin/pkill kubelet
```
JosephSalisbury: Why do we need this? Shouldn't systemd stop the kubelet?

tfussell: I'm not sure why that's there. I removed it and tested and the process was stopped just fine. Fixed!

v_4_9_0/worker_template.go
mazzy89: It would have been great to bake it into our AMI but we don't maintain AMIs. We simply pull CoreOS AMIs as they are and use them and then configure accordingly. I agree anyway with the current approach. I've seen it around and all the community is struggling with that and with the deprecation of kubelet containerized so 👍 for me

v_4_9_0/worker_template.go
```diff
@@ -138,15 +138,32 @@ systemd:
...
+      Requires=docker.service
+      [Service]
+      Type=oneshot
+      RemainAfterExit=yes
+      TimeoutStartSec=0
+      Environment="IMAGE={{ .RegistryDomain }}/{{ .Images.Kubernetes }}"
+      Environment="NAME=%p.service"
+      ExecStart=/bin/bash -c "/usr/bin/docker create --name $NAME $IMAGE && \
+        /usr/bin/docker cp $NAME:/hyperkube /opt/bin/hyperkube && \
+        /usr/bin/docker rm $NAME"
```
mazzy89: I'd write this in a systemd way by leveraging more the features that come out of the box with systemd.

Something like

```
ExecPreStart=/usr/bin/docker create --name $NAME $IMAGE
ExecStart=/usr/bin/docker cp $NAME:/hyperkube /opt/bin/hyperkube
ExecStop=/usr/bin/docker rm $NAME"
```

More elegant I'd say

WDYT?

tfussell: Good idea. I wasn't sure how that would interact with a oneshot unit, but it does look nicer. I'll give it a try.

v_4_9_0/files/config/kubelet-master.yaml.tmpl
```diff
@@ -20,7 +20,6 @@ kubeReserved:
   memory: 1280Mi
   ephemeral-storage: 1024Mi
 kubeReservedCgroup: /kubereserved.slice
-kubeletCgroups: /kubereserved.slice
```
corest: That is something we should fix instead of removing

tfussell: This is what I'm trying to test now. Kubelet still runs in the kubereserved slice without this as it's in the systemd unit. How can you check if this affects the resource calculation in Kubernetes?

corest: Why can’t we check what is wrong with logs instead of calculating allocatable?
There is some bug which we first need to prove, then decide on a way to proceed.

corest: I’m happy to help if you give me a cluster tomorrow

tfussell: I can reliably reproduce the bug by toggling this `kubeletCgroups: /kubereserved.slice` in the kubelet config. `journalctl -u k8s-kubelet` only shows the first two seconds of logs and then stops. `journalctl -o verbose` shows the missing kubelet log lines without the SYSTEMD_UNIT annotation. I'm not sure if it's a systemd bug or some improper configuration. Help tomorrow would be appreciated.

corest: I've looked into this and didn't find what could be wrong. But even removing `kubeletCgroups` didn't fix the issue for me. It freezes every time

MarcelMue: What are our options here?

We need to pick this up again soonish for 1.16 but not having logs in journalctl is unacceptable. Is there somewhere upstream where we could raise this?

mazzy89: I've run a quick search through upstream issues and I could not find anything. Perhaps we need to dig more on upstream and open it there.

tfussell: I am leaning towards it being a systemd bug. I was thinking of looking in the source code to see how kubelet changes its own cgroup and see if I can reproduce without kubelet. I get the impression that these advanced cgroup options aren't used very often.

tfussell:
I updated and test this today today, and the cluster started fine. I compared node allocatable between a TC created with this branch (called 9.1.0 for now) and a 9.0.0 (i.e. containerized) TC. They are identical so I think removing `kubeletCgroups` from the kubelet config and leaving it set only in the systemd unit `Slice` is sufficient for ensuring kubelet resources are accounted for:
```
Master 9.1:
Capacity:
 cpu:                2
 ephemeral-storage:  5110Mi
 hugepages-1Gi:      0
 hugepages-2Mi:      0
 memory:             8169076Ki
 pods:               110
Allocatable:
 cpu:                1400m
 ephemeral-storage:  4086Mi
 hugepages-1Gi:      0
 hugepages-2Mi:      0
 memory:             6260340Ki
 pods:               110

Master 9.0:
Capacity:
 cpu:                2
 ephemeral-storage:  5110Mi
 hugepages-1Gi:      0
 hugepages-2Mi:      0
 memory:             8169076Ki
 pods:               110
Allocatable:
 cpu:                1400m
 ephemeral-storage:  4086Mi
 hugepages-1Gi:      0
 hugepages-2Mi:      0
 memory:             6260340Ki
 pods:               110

Worker 9.1:
Capacity:
 cpu:                2
 ephemeral-storage:  40940Mi
 hugepages-1Gi:      0
 hugepages-2Mi:      0
 memory:             8169076Ki
 pods:               110
Allocatable:
 cpu:                1500m
 ephemeral-storage:  39916Mi
 hugepages-1Gi:      0
 hugepages-2Mi:      0
 memory:             6784628Ki
 pods:               110

Worker 9.0:
Capacity:
 cpu:                2
 ephemeral-storage:  40940Mi
 hugepages-1Gi:      0
 hugepages-2Mi:      0
 memory:             8169076Ki
 pods:               110
Allocatable:
 cpu:                1500m
 ephemeral-storage:  39916Mi
 hugepages-1Gi:      0
 hugepages-2Mi:      0
 memory:             6784628Ki
 pods:               110
```

kopiczko:
This makes sense to me too:

```
	// kubeletCgroups is the absolute name of cgroups to isolate the kubelet in
	// Dynamic Kubelet Config (beta): This field should not be updated without a full node
	// reboot. It is safest to keep this value the same as the local config.
	// Default: ""
	// +optional
	KubeletCgroups string `json:"kubeletCgroups,omitempty"`
```

I guess this makes sense only when it isn't started in the cgroup already?

tfussell:
> This makes sense to me too:
> 
> ```
> 	// kubeletCgroups is the absolute name of cgroups to isolate the kubelet in
> 	// Dynamic Kubelet Config (beta): This field should not be updated without a full node
> 	// reboot. It is safest to keep this value the same as the local config.
> 	// Default: ""
> 	// +optional
> 	KubeletCgroups string `json:"kubeletCgroups,omitempty"`
> ```
> 
> I guess this makes sense only when it isn't started in the cgroup already?

Exactly. If we run `kubelet` in a container, `kubeletCgroups` allows the `kubelet` to change its cgroup after it starts, but we can just use `Slice=` in the systemd service to get the same effect as far as I can tell.

By the way, I believe that this cgroup change after start is what's breaking the `journald` logs. I tried following the `kubelet` source code for this config, but it's very circuitous. I still can't find out how it changes its own cgroup. I may open an issue in `k/k` later on to see if anyone else is having the problem.

PR #604, opened 2019-11-11
IPVS mode

mazzy89:
This PR enables IPVS mode on kube-proxy.

the default IPVS scheduler algorithm is round-robin (rr).

Besides the change of the `mode` in the kube-proxy manifest, this PR wires some fine-grained tuning on some kernel parameters.

There are a lot of discussions running upstream about these two parameters and the reason is because of how IPVS connection tracking works.

**TL;DR**; in case of a backend dies unexpectedly without sending a `TRACE` back, for instance, a node crash, the client keeps sending data to that conn that it is still not closed and hence blackholed.

Datadog used them to prevent some issues. Discussion here https://twitter.com/jpetazzo/status/1189293822271983623?s=20

Also in k8s upstream repo, some discussions are running about this. https://github.com/kubernetes/kubernetes/issues/77903#issuecomment-510793748

I have already tested it on our test clusters but nothing there all good. I don't expect to spot any obvious issue.

I spoke already with our customer to test it in the proposed cluster on-prem. Discussion here https://gigantic.slack.com/archives/C6L8J93N0/p1573466656317400

### DIFF
--- a/v_5_0_0/files/conf/hardening.conf
+++ b/v_5_0_0/files/conf/hardening.conf
@@ -17,3 +17,5 @@ net.ipv4.ip_local_port_range=1024 65535
 net.ipv4.conf.all.rp_filter = 1
 net.ipv4.conf.all.arp_ignore = 1
 net.ipv4.conf.all.arp_announce = 2
+net.ipv4.vs.expire_nodest_conn = 1
+net.ipv4.tcp_retries2 = 5
--- a/v_5_0_0/files/config/kube-proxy.yaml
+++ b/v_5_0_0/files/config/kube-proxy.yaml
@@ -2,7 +2,7 @@ apiVersion: kubeproxy.config.k8s.io/v1alpha1
 clientConnection:
   kubeconfig: /etc/kubernetes/config/proxy-kubeconfig.yaml
 kind: KubeProxyConfiguration
-mode: iptables
+mode: ipvs
 resourceContainer: /kube-proxy
 clusterCIDR: {{.Cluster.Calico.Subnet}}/{{.Cluster.Calico.CIDR}}
 metricsBindAddress: 0.0.0.0:10249


mazzy89:
tested this and works. merging.

corest:
What was the test? How many customers were checked with this change on their clusters? Last time we did such a quick switch with a single test for a single customer - we then spend ~2 months reverting changes. 

It would be nice to have at least a few performance tests (especially with big number of services) with comparing metrics and health of the kube-proxy in iptables/ipvs modes. 

If those tests were performed - please provide some data. For now, it looks like blind switch. Correct me if I'm wrong.

corest:
More thoughts from @giantswarm/sig-infrastructure also appreciated

mazzy89:
https://github.com/giantswarm/giantswarm/issues/6562#issuecomment-552933531

corest:
IMO, that's not enough for a switch. Previous issues with ipvs were related to tables overflow. The test you had can't cause enough load. Therefore, I think we don't have enough data to do this switch. Also, this should be tested on all platforms

mazzy89:
Fair enough. then I would propose to narrow down this a bit. We would like to release this only for KVM. so I'd say to flag this and have enabled only for KVM. Other platforms later on. WDYT?

We need then to recreate the scenarios we had. I was not around at that time so I can't add anything.

From my understanding tables overflow could be then recreated by having many nodes in a cluster and of course many pods.  Hard scenario to recreate I'd say. any idea?

corest:
> We would like to release this only for KVM. so I'd say to flag this and have enabled only for KVM.

Agree. But still tests are missing.

Tables overflow is caused by huge number of services. As services are the entities kube-proxy mainly takes care about. So it is easy to reproduce high load with big number of services.

mazzy89:
Sure. we would provide tests beforehand. Rollback this, keeping aside and provide tests.

So just creating `Services` objects in the cluster would overload IPVS? it does not convince me that. I had under my hands' cluster with ~20k `Services` and everything was fine. But perhaps the assumptions were completely different (k8s version, CNI, etc...). 

But yeah only if we test we can know more. So here the thing.

Let me rollback this and taking out of the git master the config. I'll create a PR and send this PR out of our tests. How does it sound?

corest:
Maybe this overload thing is not a thing anymore. I don't know. I just think we should have proved stability for such changes. As rolling this back is super painful. 

You can skip reverting while we're not releasing this. If tests are fine - leave it like this. If there are issues - then revert. I think it should be safe unless you'll postpone this for later. Then it is better to revert

mazzy89:
Gotcha. let me get back here for the next week with tests in my hands.

PR #607, opened 2019-11-18
Make restricted PSP more secure

whites11:
Make restricted PSP more secure

See https://github.com/giantswarm/giantswarm/issues/7643


mazzy89:
Can you fix here the title?

whites11:
> Can you fix here the title?

done

PR #609, opened 2019-11-20
Add default Network Policy deny-all

mazzy89:
Towards https://github.com/giantswarm/giantswarm/issues/6413

Still, need to test this.

One thing that I need to verify is the presence of the `giantswarm` ns meanwhile the creation of the Network Policy.

Don't recall at the moment at which point of the TC bootstrap the `giantswarm` ns is created.

PR #612, opened 2019-11-27
ensure namespaces exist

whites11:
needed in clusters where the app-operator is not running)

see https://github.com/giantswarm/k8scloudconfig/pull/609

### DIFF
--- a/v_5_0_0/files/conf/k8s-addons
+++ b/v_5_0_0/files/conf/k8s-addons
@@ -25,7 +25,7 @@ do
 done
 
 # label namespaces (required for network egress policies)
-NAMESPACES="default kube-system" 
+NAMESPACES="default kube-system giantswarm"
 for namespace in ${NAMESPACES}
 do
     if ! /usr/bin/docker run -e KUBECONFIG=${KUBECONFIG} --net=host --rm -v /etc/kubernetes:/etc/kubernetes $KUBECTL get namespaces -l name=${namespace} | grep ${namespace}; then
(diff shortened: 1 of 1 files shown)


mazzy89:
Not fully convinced here. 

That code is put at the beginning of the k8s-addon when the giantswarm namespace still does not exist.

so this code will run always also in the case when app-operator is present.

mazzy89:
Does this make sense here? if e2e does not carry app-operator with them, then it would make sense to create giantswarm ns in azure e2e tests because this is an external dependency

whites11:
That's an option for sure. But we caught only one case (the e2e with missing app-operator). What if there are other cases as well? This unit fails completely if that prerequisite is not met for any reason, leading to non working clusters.
If we can, I reckon we should ensure all prerequisites are fullfilled within this script.
Other opinions welcome!

mazzy89:
I've tested this case for cluster startup on aws azure and kvm but not run e2e tests. the cluster startup which is what we are interested in works. no recall any other cases here whre this could fails.

tfussell:
My initial feeling is that this would make sense to do in `e2e-harness` since it's responsible for setting up clusters in e2e tests. Maybe somewhere around here https://github.com/giantswarm/e2e-harness/blob/master/pkg/framework/guest.go What do you all think?

mazzy89:
agree with @tfussell

whites11:
Sounds good to me. `e2e-harness` is deprecated and we will definitely see this issue again, but it's 2 against one so I'll make the change there.
Thanks for the feedack.

PR #614, opened 2019-11-29
Use registry domain param for k8s-api-healthz and wait for domains

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/7773

Fixes 2 instances where the registry domain is hardcoded to quay.io.


rossf7:
Closing in favor of #622

PR #617, opened 2019-12-04
Increase fs.inotify.max_user_instances to 8192

tfussell:
For https://github.com/giantswarm/giantswarm/issues/7711

Changing in `v_4_8_1` and `v_5_0_0` as these are unreleased versions.


tfussell:
I spent some time researching how other platforms/systems handle this setting. It seems like 8192 is fairly well established as a reasonable number. See for example https://github.com/moby/moby/issues/1044 and https://github.com/kubernetes/kubernetes/issues/32526 and https://github.com/kubernetes/kops/issues/2912 I'll modify this PR, retest, re-request review, and merge.

tfussell:
Tested on `kvm-operator`. Cluster started and is healthy. Confirmed that the value is changed: 
```
cat /proc/sys/fs/inotify/max_user_instances
8192
```

PR #627, opened 2020-01-21
v5.1.0: copy

MarcelMue:


v_5_1_0/cloudconfig.go
```diff
@@ -0,0 +1,103 @@
+package v_5_0_0
```
omegas27: Should that not be v_5_1_0 in for every file?

MarcelMue: Next PR will sort imports :)

MarcelMue:
Going with this - not really controversial.

PR #631, opened 2020-01-30
Lowercase $(hostname) when matching k8s node name with it

tuommaki:
Starting from release v_5_0_0 e.g. node labels were set by separate systemd
unit with kubectl and that must match running hostname with k8s node name.

That matching is case sensitive and since Azure sets hostname with uppercase
letters in Base36 encoding, the hostname must be convered to lowercase in order
to match k8s node name.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ version directory, and  then changes are introduced.
 
 ## [v5.1.0] - 2020-01-21
 
+- Lowercase $(hostname) to match k8s node name e.g. when using with kubectl.
+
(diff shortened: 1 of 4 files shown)


tuommaki:
```
$ kubectl --context giantswarm-ps9qs get nodes
NAME                  STATUS   ROLES    AGE     VERSION
ps9qs-master-000000   Ready    master   6h21m   v1.16.3
ps9qs-worker-00000b   Ready    worker   16m     v1.16.3
ps9qs-worker-00000d   Ready    worker   16m     v1.16.3
ps9qs-worker-00000e   Ready    worker   16m     v1.16.3
```

PR #635, opened 2020-02-12
Increase kubelet hard memory eviction

mazzy89:
Towards https://github.com/giantswarm/giantswarm/issues/8361


mazzy89:
Do you think this could have an impact on nodes and be disruptive?

MarcelMue:
> Do you think this could have an impact on nodes and be disruptive?

I don't think so - just something where we need to aware that customers might complain?

mazzy89:
then let's communicate in the CHANGELOG of the next release. sound it resonable?

MarcelMue:
> then let's communicate in the CHANGELOG of the next release. sound it resonable?

Works for me :+1:

PR #636, opened 2020-02-14
Avoid checking for domain validity of IP addresses

whites11:
So the `wait-for-domains` service/script is used on a master node to wait for the DNS records to be ready before going on with the setup of the kubernetes control plane.
In azure, the script looked like this (for a long time):

```
#!/bin/bash
domains="127.0.0.1 api.yguz7.k8s.godsmack.westeurope.azure.gigantic.io quay.io"

for domain in $domains; do
until nslookup $domain; do
    echo "Waiting for domain $domain to be available"
    sleep 5
done

echo "Successfully resolved domain $domain"
done
```

At some point, `nslookup 127.0.0.1` started failing and the master node started not getting boostrapped correctly.

This PR adds a check to avoid checking the "DNS validity" for ip addresses.
Tested on a live cluster on azure and it works (cluster comes correctly up with this change).

### DIFF
--- a/v_5_0_0/files/conf/wait-for-domains
+++ b/v_5_0_0/files/conf/wait-for-domains
@@ -2,10 +2,13 @@
 domains="{{.Cluster.Etcd.Domain}} {{.Cluster.Kubernetes.API.Domain}} {{ .RegistryDomain }}"
 
 for domain in $domains; do
-until nslookup $domain; do
-    echo "Waiting for domain $domain to be available"
-    sleep 5
-done
-
-echo "Successfully resolved domain $domain"
+  if [[ $domain =~ ^[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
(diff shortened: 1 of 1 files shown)


whites11:
> I wouldn’t perform this check. It is better to find out why do we have ip in nslookup check. And maybe exclude then localhost from check instead of applying such regex checks

Added 127.0.0.1 check + a note on celestial inbox to check the reason behind the presence of 127.0.0.1 in that script.

corest:
That’s not what I meant. It is templated from somewhere. So maybe for azure it just needs to be templates differently. 
Also you’re changing stuff in v_5_0_0 which is released already with operators. So it can’t be changed anymore

whites11:
@corest OK, I give up this way (even though it would be much quicker to solve our current issues).
For my understanding of how this repo is handled, if I have to make a (legit) fix on any script, I need to create a new version?

PR #637, opened 2020-02-18
Revert "Increase kubelet hard memory eviction (#635)"

mazzy89:
This reverts commit 0a6035168fd670351f1204ad3444ecf67b681cd0.

According to @sslavic's findings this change does not help to fix https://github.com/giantswarm/giantswarm/issues/8361 so reverting


mazzy89:
Yep :-(

PR #638, opened 2020-02-25
Ephemeral port collisions

stone-z:
Towards https://github.com/giantswarm/giantswarm/issues/8798

Reserve ports 30000-32767 from ephemeral port range so they don't get assigned and then denied to kube-apiserver.


MarcelMue:
Change is okay - I added sig infra as reviewers as I can't judge the consequences and would like some more opinions here.

calvix:
Looks fine, if I get this correct kernel won't use this port range for connections so they are reserved for nodePort

CHANGELOG.md
```diff
@@ -9,10 +9,11 @@ The latest version is considered WIP and it is a subject of change. All other
 versions are frozen. To freeze current version all files are copied to a new
 version directory, and  then changes are introduced.
 
-## [v5.1.0] - 2020-01-21
+## [v5.1.0] - Unreleased
```
calvix: I got a feeling this version is already released so this change should go to the new version of k8scloudconfig

stone-z: @MarcelMue can you confirm -- should I add a 5.2.0 version or is 5.1.0 still waiting?

MarcelMue: kvm is using `5.0.0` https://github.com/giantswarm/kvm-operator/blob/2fd4e6e377b279837f89b915b465d579510c4afb/service/controller/cloudconfig/master_template.go#L8
aws is using `5.0.0` https://github.com/giantswarm/aws-operator/blob/master/service/controller/internal/cloudconfig/tccp.go#L11

BUT azure uses `5.1.0` https://github.com/giantswarm/azure-operator/blob/b23abfe0195bdfc7ee99d533fdc39cd186d0d479/service/controller/v13/cloudconfig/master_template.go#L9

So would be best to ask @giantswarm/team-celestial why they didn't create a new `k8scc` version when they created their azure release :P

v_5_2_0/files/conf/hardening.conf
```diff
@@ -0,0 +1,24 @@
...
+net.ipv4.tcp_timestamps = 0
+net.ipv6.conf.all.accept_redirects = 0
+net.ipv6.conf.default.accept_redirects = 0
+# Increased mmapfs because some applications, like ES, need higher limit to store data properly
+vm.max_map_count = 262144
+# Ingress controller performance improvements
+# See https://github.com/kubernetes/ingress-nginx/issues/1939
+net.core.somaxconn=32768
+net.ipv4.ip_local_port_range=1024 65535
+# Reserved to avoid conflicts with kube-apiserver, which allocates within this range
```
stone-z: This is the only substantive change for this PR

stone-z:
Moved the change to 5.2.0

CHANGELOG.md
stone-z: Added v_5_2_0

stone-z:
> Looks fine, if I get this correct kernel won't use this port range for connections so they are reserved for nodePort

Right, it just means the kernel won't use this range when it offers randomly-selected local ports. They can still be bound explicitly if some application requests it.

CHANGELOG.md
tuommaki: :see_no_evil: Need to add that as part of release checklist. Totally forgot.

v_5_2_0/files/conf/hardening.conf
calvix: so the way how we do things is that adding a new version is always separate PR and then you create a new PR where you add changes. The reason for this is that in this PR its completely impossible to find any changes and when someone is trying to figure out where the change happened or what went wrong this change is un traceable.

stone-z: My bad, opened https://github.com/giantswarm/k8scloudconfig/pull/639 to add the new version

stone-z:
This rebase got ugly. Closing this in favor of https://github.com/giantswarm/k8scloudconfig/pull/640

PR #639, opened 2020-02-26
Add new version v_5_2_0

stone-z:
Add folder for 5.2.0


MarcelMue:
Can you also add the changelog - just to make it nice and obvious to see right away that the new WIP exists.

PR #642, opened 2020-03-03
First boot always

corest:
We have the issue in aws, where reboot doesn't reprovision a node. It is not consistent behavior. 
This PR should fix it

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ version directory, and  then changes are introduced.
 ### Changed
 
 - Reserve ports `30000-32767` from ephemeral port range for `kube-apiserver` use.
+- Make provisioning idempotent by generating `/boot/coreos/first_boot` file on every boot.
 
 ## [v5.1.0] - 2020-01-21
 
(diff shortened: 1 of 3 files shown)


tfussell:
Are reboots supposed to reprovision? I thought cloud-init was only intended to run on first boot.

MarcelMue:
This is a bigger change - some of the SE workflows rely on not reprovisioning on reboot AFAIK. cc @pipo02mix

corest:
> Are reboots supposed to reprovision? I thought cloud-init was only intended to run on first boot.
Yes. It is happening in kvm and azure, but not in aws. THat's a whole thing - a reboot is a clean install. 

> This is a bigger change - some of the SE workflows rely on not reprovisioning on reboot AFAIK

They should not rely on this as we do reprovision in kvm/azure. Regarding OIDC for tenant - it should be implemented in proper way and can't be a blocker for this change

mazzy89:
This should follow the common convention set up by CoreOS as in here https://github.com/flatcar-linux/scripts/blob/9e1c23f3f44d2751076e770f43f7a6db05d49652/build_library/grub.cfg#L68-L71

So it is fine for me.

mazzy89:
We'll move this once we'll migrate to Flatcar following the convention in here https://github.com/flatcar-linux/scripts/blob/flatcar-master/build_library/grub.cfg#L70

PR #650, opened 2020-03-05
AWS VPC CNI

mazzy89:
Towards https://github.com/giantswarm/giantswarm/issues/9259

This PR introduces pretty important changes to our AWS provider.

It replaces Calico as CNI and introduces AWS VPC CNI however it leaves Calico in place for ensuring network policies.

Spread around there are comments I have added to make review easier.


v_5_2_0/files/config/kube-proxy.yaml
```diff
@@ -4,5 +4,4 @@ clientConnection:
 kind: KubeProxyConfiguration
 mode: iptables
 resourceContainer: /kube-proxy
-clusterCIDR: {{.Cluster.Calico.Subnet}}/{{.Cluster.Calico.CIDR}}
```
mazzy89: If we're using the AmazonVPC networking, we should omit the ClusterCIDR because pod IPs are real, routable IPs in the VPC, and they are not in a specific CIDR range that allows us to distinguish them from other IPs.  Omitting the ClusterCIDRcauses kube-proxy never to SNAT when proxying clusterIPs, which is the behavior we want for pods.

v_5_2_0/master_template.go
```diff
@@ -293,6 +293,10 @@ systemd:
         --register-with-taints=node-role.kubernetes.io/master=:NoSchedule \
         --kubeconfig=/etc/kubernetes/kubeconfig/kubelet.yaml \
         --node-labels="node.kubernetes.io/master,role=master,ip=${DEFAULT_IPV4},{{.Cluster.Kubernetes.Kubelet.Labels}}" \
+        {{ if eq .Cluster.Kubernetes.CloudProvider "aws" -}}
```
mazzy89: From upstream https://github.com/aws/amazon-vpc-cni-k8s#setup

v_5_2_0/worker_template.go
```diff
@@ -190,6 +190,10 @@ systemd:
         --register-node=true \
         --kubeconfig=/etc/kubernetes/kubeconfig/kubelet.yaml \
         --node-labels="node.kubernetes.io/worker,role=worker,ip=${DEFAULT_IPV4},{{.Cluster.Kubernetes.Kubelet.Labels}}" \
+        {{ if eq {{.Cluster.Kubernetes.CloudProvider}} "aws" -}}
```
mazzy89: Same as https://github.com/giantswarm/k8scloudconfig/pull/650/files#r388368501

v_5_2_0/master_template.go
```diff
@@ -293,6 +293,10 @@ systemd:
         --register-with-taints=node-role.kubernetes.io/master=:NoSchedule \
         --kubeconfig=/etc/kubernetes/kubeconfig/kubelet.yaml \
         --node-labels="node.kubernetes.io/master,role=master,ip=${DEFAULT_IPV4},{{.Cluster.Kubernetes.Kubelet.Labels}}" \
+        {{ if eq .Cluster.Kubernetes.CloudProvider "aws" }}
```
mazzy89: From upstream https://github.com/aws/amazon-vpc-cni-k8s#setup

v_5_2_0/files/config/kube-proxy.yaml
mazzy89: This requires changes and it must be gated.

mazzy89:
This PR requires changes according to to values that we are going to introduce in here https://github.com/giantswarm/installations/pull/1094

mazzy89:
`DisableCalico` has been added and used by Celestial for Azure? Is it like that? @whites11 could you advice here?

whites11:
> `DisableCalico` has been added and only used by Celestial for Azure? Is it like that? @whites11 could you advice here?

I know that in Azure we use calico for policies only, because overlay network uses the native azure features.
This came way before I joined the company so I can't add much more to the above.

mazzy89:
I see. thanks, @whites11 to shed some light in here. need to get why that flag has been added and for which reason then. Missing the reason why disabling completely Calico. In which use cases we could ever use that?

v_5_2_0/files/conf/k8s-addons
```diff
@@ -88,29 +88,29 @@ $KUBECTL delete pods -l k8s-app=kube-proxy -n kube-system
 
 {{ if not .DisableCalico -}}
 
-# apply calico
-CALICO_FILE="calico-all.yaml"
+{{ if eq .Cluster.Kubernetes.CloudProvider "aws" -}}
```
kopiczko: This is unrelated but @giantswarm/team-celestial how about doing the same for Azure and removing `.DisableCalico` setting?

It would be easier to update calico that way because we'd have all manifests in a single repo.

v_5_2_0/files/conf/k8s-addons
```diff
@@ -88,29 +88,29 @@ $KUBECTL delete pods -l k8s-app=kube-proxy -n kube-system
...
-    echo "Waiting for calico to be ready . . "
-    sleep 3s
-done
+{{ if eq .Cluster.Kubernetes.CloudProvider "aws" -}}
+echo "Waiting for aws-node to be ready..."
+$KUBECTL -n kube-system -l k8s-app=aws-node wait --for=condition=Ready --timeout=10m pods
+{{ end -}}
+
+echo "Waiting for calico-node to be ready..."
+$KUBECTL -n kube-system -l k8s-app=calico-node wait --for=condition=Ready --timeout=10m pods
```
kopiczko: This is so much better.

v_5_2_0/files/config/kube-proxy.yaml
kopiczko: Why are you adding `DisableCalico` here? AFAIK this is used by azure-operator which is basically doing the same thing. Using Azure route tables and using Calico in policy mode.

v_5_2_0/files/config/kubelet-master.yaml.tmpl
```diff
@@ -21,6 +21,7 @@ kubeReserved:
   memory: 1280Mi
   ephemeral-storage: 1024Mi
 kubeReservedCgroup: /kubereserved.slice
+maxPods: 110
```
kopiczko: Is it related?

v_5_2_0/files/k8s-resource/calico-policy-only.yaml
```diff
@@ -0,0 +1,515 @@
+---
```
kopiczko: I guess we have something similar in azure-operator now. It would be really good to have that aligned.

v_5_2_0/files/config/kubelet-master.yaml.tmpl
calvix: its just placeholder now, I will add some logic to set this dynamically for AWS based on instance type

v_5_2_0/files/conf/k8s-addons
mazzy89: Actually I have spent some time to refactor the entire k8s-adodns file using helpers and only kubectl and it looks much leaner. but never had the chance to commit it.

paurosello:
Based on docs in the CNI repo I think we should also setup MINIMUM_IP_TARGET and WARM_IP_TARGET values.

WARM_IP_TARGET = 20
MINIMUM_IP_TARGET = 5

Sound reasonable values to avoid throttling the AWS API, most of our nodes are running around 20 pods and we can add in 5 blocks as we are not so "short" in IPs

> MINIMUM_IP_TARGET is for pre-scaling, WARM_IP_TARGET is for dynamic scaling. For example, suppose a cluster has an expected pod density of approximately 30 pods per node. If WARM_IP_TARGET is set to 30 to ensure there are enough IPs allocated up front by the CNI, then 30 pods are deployed to the node, the CNI will allocate an additional 30 IPs, for a total of 60, accelerating IP exhaustion in the relevant subnets. If instead MINIMUM_IP_TARGET is set to 30 and WARM_IP_TARGET to 2, after the 30 pods are deployed the CNI would allocate an additional 2 IPs. This still provides elasticity, but uses roughly half as many IPs as using WARM_IP_TARGET alone (32 IPs vs 60 IPs).
> 
> This also improves reliability of the EKS cluster by reducing the number of calls necessary to allocate or deallocate private IPs, which may be throttled, especially at scaling-related times.

calvix:
@paurosello @mazzy89  Any last comments? this is working fine I am ready to merge

v_5_2_0/files/conf/setup-kubelet-environment
```diff
@@ -0,0 +1,22 @@
...
+# https://github.com/aws/amazon-vpc-cni-k8s#eni-allocation
+# https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-eni.html
+
+instance_size=$(curl http://169.254.169.254/latest/meta-data/instance-type 2>/dev/null| cut -d\. -f2)
+
+if [[ "${instance_size}" =~ ^(4xlarge|8xlarge|9xlarge|12xlarge|16xlarge|18xlarge|24xlarge)$ ]]; then
+	MAX_PODS=110
+else 
+	MAX_PODS=40
+fi
```
xh3b4sd: We need to document this properly. /cc @marians

v_5_2_0/master_template.go
```diff
@@ -294,7 +294,6 @@ systemd:
         --kubeconfig=/etc/kubernetes/kubeconfig/kubelet.yaml \
         --node-labels="node.kubernetes.io/master,role=master,ip=${DEFAULT_IPV4},{{.Cluster.Kubernetes.Kubelet.Labels}}" \
         {{ if eq .Cluster.Kubernetes.CloudProvider "aws" }}
-        --cni-config-dir=/etc/cni/net.d \
```
mazzy89: Just personal curiousity. Why are this not needed?

calvix: this is the default settings, so there is no point of defining it again with the same value

mazzy89: oh right, man. I did not look at the default parameters. thank you

v_5_2_0/files/k8s-resource/calico-policy-only.yaml
```diff
@@ -495,6 +495,10 @@ spec:
             # Disable IPv6 on Kubernetes.
             - name: FELIX_IPV6SUPPORT
               value: "false"
+            - name: FELIX_IPTABLESMANGLEALLOWACTION
+              value: Return
+            - name: FELIX_INTERFACEPREFIX
+              value: eni
```
mazzy89: I'm pretty sure that our interfaces are not named `eni`. the right prefix here is `cali`

paurosello:
for some reason I cannot comment on your thread @mazzy89, we had to chnage the name of the interfaces in AWS CNI as it has some hardcoded iptables rules :( check https://github.com/giantswarm/aws-operator/pull/2197

mazzy89:
I see. thanks for insights. this deviation is pretty important and i believe it is necessary to be documented.

corest:
i hope this was tested with onprem tenant clusters (all those calico related conditions rules)

calvix:
KVM looks good
```
[remote][calvix@master-8cr6f-745c8cb97d-8hz5s] ~ | geckon | 0sitz | master
 #: cat /etc/kubernetes/config/proxy-config.yml 
apiVersion: kubeproxy.config.k8s.io/v1alpha1
clientConnection:
kind: KubeProxyConfiguration
mode: iptables
clusterCIDR: 172.24.0.0/16
```

and cluster was healthy overall

v_5_2_0/master_template.go
```diff
@@ -382,12 +397,25 @@ storage:
         source: "data:text/plain;base64,{{ index .Files "conf/trusted-user-ca-keys.pem" }}"
 
     {{- if not .DisableCalico }}
+    {{- if eq .Cluster.Kubernetes.CloudProvider "aws" }}
+    - path: /srv/aws-cni.yaml
+      filesystem: root
+      mode: 0644
+      contents:
+        source: "data:text/plain;charset=utf-8;base64,{{  index .Files "k8s-resource/aws-cni.yaml" }}"
```
tuommaki: I think this file might be missing from this PR. I was working on `v_6_0_0` which was copied from `v_5_2_0` but the file is not anywhere:
```
~/go/src/github.com/giantswarm/k8scloudconfig $ git checkout master
Switched to branch 'master'
Your branch is up to date with 'origin/master'.
~/go/src/github.com/giantswarm/k8scloudconfig $ git pull
Already up to date.
~/go/src/github.com/giantswarm/k8scloudconfig $ find . -name aws-cni.yaml 
~/go/src/github.com/giantswarm/k8scloudconfig $ 
```

PR #656, opened 2020-03-10
v4.9.1: changelog

tfussell:
Part 3 of https://github.com/giantswarm/k8scloudconfig/pull/654

`aws-operator@5.5.0` uses [k8scloudconfig v_4_9_0](https://github.com/giantswarm/aws-operator/blob/d0af87284deb59eca4035772db2c68866bfe251a/Gopkg.lock#L418). To implement "versions from releases", we therefore need to create a new patch `k8scloudconfig` version based on `v_4_9_0` which allows all image fields to be templated.


tfussell:
Going with a ping.

PR #657, opened 2020-03-10
v4.9.1: Make images and versions all template-able

tfussell:


tfussell:
I have tested this in `aws-operator` successfully here https://github.com/giantswarm/aws-operator/pull/2180 Cluster creation and upgrades both worked. I think this is ready to go.

v_4_9_1/release.go
```diff
@@ -0,0 +1,62 @@
+package v_4_9_1
+
+import (
+	"fmt"
+
+	"github.com/giantswarm/apiextensions/pkg/apis/release/v1alpha1"
+)
+
+func BuildImages(registryDomain string, versions Versions) Images {
```
tfussell: It's not totally clear where this code should live. This seemed like as good a place as any.

v_4_9_1/release.go
```diff
@@ -0,0 +1,62 @@
...
+	return Images{
+		CalicoCNI:             buildImage(registryDomain, "giantswarm/cni", versions.Calico),
```
tfussell: These are just for defaulting. The provider operator will call this, but it's free to override any image before sending them to the templating step.

v_4_9_1/files/conf/k8s-addons
```diff
@@ -1,21 +1,20 @@
 #!/bin/bash
 
 export KUBECONFIG=/etc/kubernetes/kubeconfig/addons.yaml
-# kubectl 1.12.2
-KUBECTL={{ .RegistryDomain }}/giantswarm/docker-kubectl:f5cae44c480bd797dc770dd5f62d40b74063c0d7
+export KUBECTL="/usr/bin/docker run -e KUBECONFIG=${KUBECONFIG} --net=host --rm -v /srv:/srv -v /etc/kubernetes:/etc/kubernetes {{ .Images.Kubectl }}"
```
tfussell: This refactoring wasn't totally necessary, but it makes the file much more readable and therefore easier to debug/edit.

v_4_9_1/files/k8s-resource/calico-all.yaml
```diff
@@ -7,13 +7,9 @@
 #  - Added resource limits to install-cni.
 #  - Added 'priorityClassName: system-cluster-critical' to calico daemonset.
 #
-# Calico Version v3.9.1
+# Originally intended for Calico version v3.9.1
```
tfussell: This might work for new minor/patch releases of Calico. This just makes it clear that it was originally tested for this version.

v_4_9_1/master_template.go
```diff
@@ -362,18 +362,6 @@ systemd:
       [Install]
       WantedBy=multi-user.target
 
-  - name: debug-tools.service
```
tfussell: This service was complicating the process and it's value is questionable for me. If we have SSH access, we can SFTP these binaries in instead.

MarcelMue: We already changed this in v_5_X_X - instead of removing the whole unit, please just set enabled to false.

If you have network issues, it is very cumbersome IMO to use SFTP. A disabled unit that can just be enabled when debugging is easier to deal with.

PR #659, opened 2020-03-13
Increase conntrack max

tfussell:
Towards https://github.com/giantswarm/giantswarm/issues/9154

### DIFF
--- a/v_4_9_1/files/config/kube-proxy.yaml
+++ b/v_4_9_1/files/config/kube-proxy.yaml
@@ -5,4 +5,7 @@ kind: KubeProxyConfiguration
 mode: iptables
 resourceContainer: /kube-proxy
 clusterCIDR: {{.Cluster.Calico.Subnet}}/{{.Cluster.Calico.CIDR}}
+{{- if .Cluster.KubeProxy.ConntrackMaxPerCore }}
+conntrackMaxPerCore: {{.Cluster.KubeProxy.ConntrackMaxPerCore}}
+{{- end }}
 metricsBindAddress: 0.0.0.0:10249
--- a/v_4_9_1/files/config/kube-proxy.yaml
+++ b/v_4_9_1/files/config/kube-proxy.yaml
@@ -5,7 +5,7 @@ kind: KubeProxyConfiguration
 mode: iptables
 resourceContainer: /kube-proxy
 clusterCIDR: {{.Cluster.Calico.Subnet}}/{{.Cluster.Calico.CIDR}}
-{{- if .Cluster.KubeProxy.ConntrackMaxPerCore }}
-conntrackMaxPerCore: {{.Cluster.KubeProxy.ConntrackMaxPerCore}}
+{{- if .Cluster.Kubernetes.NetworkSetup.KubeProxy.ConntrackMaxPerCore }}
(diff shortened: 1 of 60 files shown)


mazzy89:
It is not what we want. Conntrack number depends by instance type. Bigger instances can afford higher conn track.

we would like to give our customers the possibility to configure conntrack. This is possible by using certain kube-proxy flags [1] or property in the kube-proxy [2] that on the ground they tune up the kernel running sysctl.

Leaving to the user the possiblity to increase is the best option here.

[1] https://kubernetes.io/docs/reference/command-line-tools-reference/kube-proxy/#options
[2] https://godoc.org/k8s.io/kubernetes/pkg/proxy/apis/config#KubeProxyConntrackConfiguration

calvix:
this should probably go to all other bigger versions as well

v_4_9_1/files/config/kube-proxy.yaml
```diff
@@ -5,4 +5,7 @@ kind: KubeProxyConfiguration
 mode: iptables
 resourceContainer: /kube-proxy
 clusterCIDR: {{.Cluster.Calico.Subnet}}/{{.Cluster.Calico.CIDR}}
+{{- if .Cluster.KubeProxy.ConntrackMaxPerCore }}
+conntrackMaxPerCore: {{.Cluster.KubeProxy.ConntrackMaxPerCore}}
```
MarcelMue: Does this actually work? Where is `ConntrackMaxPerCore` actually set?

mazzy89: We need to set up `ConntrackMaxPerCore` outside of the k8scc

MarcelMue:
Can we merge this now? Do we potentially break something by merging into 5_2_0?

tfussell:
> this should probably go to all other bigger versions as well

I agree, but I think we'll do it in one version at a time to keep PRs more digestible. We'll probably end up adding it to all unreleased k8scc versions.

tfussell:
As far as merging to v5.2.0, I went through all the active operator versions to see which k8scc is being used:
- aws master (wip) -> v5.2.0
- aws master (v8.2.1) -> v5.2.0
- aws legacy (v5.6.0) -> v5.0.0
- aws legacy-1-15 (wip) -> v4.9.1
- aws thiccc (v30 [latest]) -> v4.8.0
- kvm master (wip) -> v5.0.0
- kvm thiccc (v27 [latest]) -> v5.0.0
- azure master (v3.0.0) -> v5.1.1
- azure master (wip) -> v6.0.0
- azure thiccc (v13 [latest]) -> v5.1.0

So... this is complicated and I don't know, but maybe this will help us decide.

MarcelMue:
Okay - moving this from 5.2.0 to 6.0.0 as 5.2.0 is already used in a aws operator release.

MarcelMue:
@giantswarm/team-ludacris PTAL

tfussell:
Moving it to v6.0.0 makes sense. Current targets then would be:

- aws master - v6.0.0
- aws legacy - v6.0.0
- aws legacy-1-15 - v4.9.1
- azure master - v6.0.0
- azure thiccc v14 - v5.1.1
- kvm master - v6.0.0
- kvm legacy-1-15 - v4.9.1

This means we would need a new v5.1.1 based on v5.1.0. I'm not sure if it's worth the trouble for Azure v14 though since our main target was AWS.

PR #663, opened 2020-03-16
backport-aws-cni-to-v_4_x_x

calvix:
backporting PR https://github.com/giantswarm/k8scloudconfig/pull/650/ to `v_4_10_0`


calvix:
the customer decided they don't need it in k8s .1.15 so this is not needed anymore

PR #667, opened 2020-03-19
v6.0.0: Make images and versions all template-able

tuommaki:

### DIFF
--- a/v_6_0_0/master_template.go
+++ b/v_6_0_0/master_template.go
@@ -149,8 +149,7 @@ systemd:
       [Service]
       Type=oneshot
       TimeoutStartSec=0
-      # TODO: What's up with this?
-      Environment="IMAGE={{.Cluster.Kubernetes.NetworkSetup.Docker.Image}}"
+      Environment="IMAGE={{ .Images.KubernetesNetworkSetupDocker }}"
       Environment="NAME=%p.service"
       ExecStartPre=/usr/bin/mkdir -p /opt/bin/
       ExecStartPre=/usr/bin/docker pull $IMAGE
--- a/v_6_0_0/master_template.go
+++ b/v_6_0_0/master_template.go
@@ -149,6 +149,7 @@ systemd:
       [Service]
       Type=oneshot
       TimeoutStartSec=0
+      # TODO: What's up with this?
       Environment="IMAGE={{.Cluster.Kubernetes.NetworkSetup.Docker.Image}}"
       Environment="NAME=%p.service"
       ExecStartPre=/usr/bin/mkdir -p /opt/bin/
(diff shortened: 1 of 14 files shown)


tuommaki:
> I assume these are the same changes that @tfussell did earlier for 9.0.1?

Yep, all the same except that optional refactoring in here: https://github.com/giantswarm/k8scloudconfig/pull/657/files#diff-1529c1dd58ed9cdb742092415f5db5f1R4

v_6_0_0/master_template.go
```diff
@@ -149,6 +149,7 @@ systemd:
       [Service]
       Type=oneshot
       TimeoutStartSec=0
+      # TODO: What's up with this?
       Environment="IMAGE={{.Cluster.Kubernetes.NetworkSetup.Docker.Image}}"
```
tuommaki: Though this one I was wondering - what should it be? `.Images.NetworkSetupDockerImage` ?

tfussell: That seems fine as far as naming. I guess I missed this in v4.9.1. I passed the other tags in through [key.DefaultVersions in the operator](https://github.com/giantswarm/aws-operator/blob/61ba949f6da8acde9da99f379c40c85ecb8b42a4/service/controller/key/key.go#L224) so we can override them in the future. It would be nice if they had semver tags.

corest:
Why does it major version update? Is this a new Kubernetes or breaking change?
P.S. I don’t see aws cni images, but calico only

tuommaki:
> Why does it major version update? Is this a new Kubernetes or breaking change?

"1. Increment MAJOR version number when breaking library API changes introduced."

I consider this breaking the API as the `types.go` and the way how registry domain is configured for images changes.

> P.S. I don’t see aws cni images, but calico only

Whow! It seems that there's file missing even from original PR. I made a comment there.

v_6_0_0/master_template.go
tuommaki: It didn't exist in `v4.9.1` yet, I think. I'll use that one :+1:

tfussell:
I don't really understand the comment about CNI. Otherwise this looks fine to me once `.Images.NetworkSetupDockerImage` is sorted and it's all tested. Also I noticed that the `[6.0.0]` link in the changelog is broken from the previous PR. 

As a separate issue, we don't share any code between k8scc versions, but I wonder if it might make sense to share this `Images`, `Versions`, and templating function code in `pkg/template` or something since it will be in v4.9.1, v6.0.0, and maybe one or two others for legacy support. What do you think?

tuommaki:
> I don't really understand the comment about CNI.

That seems to be some work-in-progress inconsistency. I talked with Vaclav and it turns out that AWS CNI is added by `aws-operator` and is not supposed to exist in `k8scloudconfig` at all. Hence everything is fine in here. There's a bit cleanup work left for that, but that's topic for a different PR.

> I wonder if it might make sense to share this Images, Versions, and templating function code in pkg/template or something since it will be in v4.9.1, v6.0.0, and maybe one or two others for legacy support. What do you think?

At first it sounds good, but I wonder how it would look like in the future if we need to change the API for some reason for upcoming template versions? In that regard it might be easier to maintain these as-is. It's pros and cons, but I'm not sure which one is better approach. For now I'd maybe go with this and then refactor later if it seems that API stays as-is.

tuommaki:
I've addressed the open concerns. Going with this now to unblock work on next steps. Give me a shout when there's more to improve.

PR #684, opened 2020-03-30
filemanager: handler initial filepath.Walk error

wtty-fool:
This helps us avoid panic and returns a nice error when something is
**really** wrong with the file structure e.g. it is not there.

FYI: This popped up while debugging go-module-based aws-operator.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,15 +13,15 @@ version directory, and  then changes are introduced.
 
 ## [Unreleased]
 
+### Fixed
 
+- better error handling for filemanager
 
 ## [v6.0.1]
 
 ### Fixed
 
 - Fix go module.
-
-
 
 ## [v6.0.0]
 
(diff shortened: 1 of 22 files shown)


CHANGELOG.md
```diff
@@ -13,16 +13,16 @@ version directory, and  then changes are introduced.
 
 ## [Unreleased]
 
+### Fixed
 
+- better error handling for filemanager
```
kopiczko: I think we can skip changelog here. We don't know the bug it fixes so it isn't really interesting for the end user.

v_4_0_0/filemanager.go
```diff
@@ -30,6 +30,9 @@ func RenderFiles(filesdir string, ctx interface{}) (Files, error) {
 	files := Files{}
 
 	err := filepath.Walk(filesdir, func(path string, f os.FileInfo, err error) error {
+		if err != nil {
+			return microerror.Mask(err)
+		}
```
kopiczko: We shouldn't change that in already released versions.

wtty-fool: Just for `v_6_0_0` then?

kopiczko: I'm confused. See https://gigantic.slack.com/archives/C0251MXL5/p1585566502077900

wtty-fool:
I'll push my fix to the flattening PR instead.

PR #688, opened 2020-04-02
'HA masters'

calvix:
## Checklist

### DIFF
--- a/v_6_0_0/types.go
+++ b/v_6_0_0/types.go
@@ -43,7 +43,9 @@ type Params struct {
 	// cancelled if no progress has been made.
 	ImagePullProgressDeadline string
 	// Container images used in the cloud-config templates
-	Images         Images
+	Images Images
+	// MultiMaster spec
+	MultiMasters   MultiMastersSpec
 	Node           v1alpha1.ClusterNode
 	RegistryDomain string
 	SSOPublicKey   string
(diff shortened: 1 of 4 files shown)


v_6_0_0/types.go
```diff
@@ -115,6 +117,16 @@ type HyperkubePodHostMount struct {
...
 }
 
+type MultiMastersSpec struct {
+	// If set to true, k8scloduconfig will render master template for cluster of 3 masters.
+	// Defaults to false.
+	Enabled bool
+	// Etcd initial cluster is config which define which etcd are members of the cluster.
+	// The format look like to this: `etcd0=https://10.1.1.1:2380,etcd1=https://10.1.1.2:2380,etcd2=https://10.1.1.3:2380`
+	// Where 10.1.1.1, 10.1.1.2, 10.1.1.3 can be either IP or DNS  of master machine where is etcd listening.
+	EtcdInitialCluster string
```
calvix, line 127: I wanted initially to build this based on `BaseDomain` but then this would complicate things for China as there is no route53, so the operator needs to provide this value-based either on DNS or IP. It should be quite easy to do via some `key` function.

calvix: Hmm maybe we need to use DNS anyway because of the SNI in SSL certs

PR #692, opened 2020-04-09
Flatten

tfussell:
Now that provider operators are all using go modules, we can get rid of old k8scc versions and only have the current version. This is a stopgap to make changes easier and faster (no more copying/wiring versions to change a single configuration line) until we get to ignition-operator.

I have tested this in `aws-operator@legacy` successfully.

There could be some difficulties if we need to backport/hotfix old versions of provider operators, but we should be able to use `git checkout -b <working branch> <tag>` to create a branch module version in the future. Hopefully this won't be a common use case.

## Checklist


omegas27:
Wow Thomas :)

pkg/template/cloudconfig_test.go
```diff
@@ -1,11 +1,11 @@
-package v_4_0_0
+package template
```
kopiczko, line 1: How about moving stuff to `pkg/k8cloudconfig/...` instead of `pkg/...`?

My reasoning here is dealing with go modules backward compatibility packages while keeping consistency between repos.

E.g. when we release v2 of [helmclient](https://github.com/giantswarm/helmclient) then the import will look like `github.com/giantswarm/helmclient/v2`. Moving to`/pkg` doesn't solve anything because it'd become `github.com/giantswarm/helmclient/v2/pkg`, but `github.com/giantswarm/helmclient/v2/pkg/helmclient` makes sense.

kopiczko:
@nprokopic no, this is only defined in `go.mod` file. This is already done [here](https://github.com/giantswarm/k8scloudconfig/blob/f2bfdd840123bc3417d9c6235764ca98ae0f9320/go.mod#L1). So we have to import it as v6 now.

pkg/template/cloudconfig_test.go
tfussell: I've been thinking about this for a few days. I'm not sure if I understand how this would help. It feels like it just makes the import path longer since packages would repeat the module name (`import "github.com/giantswarm/k8scloudconfig/v6/pkg/k8scloudconfig/template"` vs `import "github.com/giantswarm/k8scloudconfig/v6/pkg/template"`).

I even thought about dropping `pkg/` so it would become `import "github.com/giantswarm/k8scloudconfig/v6/template"`. I'm not sure if we have guidelines on when to use `pkg/`. Some thoughts can be found here https://github.com/golang-standards/project-layout/issues/10.

I also want to keep these things consistent between repos and backwards compatible so let's try to get this right.

kopiczko: I started scribing some [docs on that](https://github.com/giantswarm/fmt/pull/109).

TL;TR I think `github.com/giantswarm/k8scloudconfig/v6/pkg/template` is fine but I'd like to keep `pkg` because `github.com/giantswarm/k8scloudconfig/k8scloudconfig` looks like a glitch.

tfussell: Okay, this makes much more sense now. The main motivation then is not having to rename the package while importing. Works for me.

PR #694, opened 2020-04-10
Remove unused kube-proxy component label

omegas27:
In an effort to have homogeneous labels, I removed this one as it is not used.


omegas27:
@MarcelMue fixed

PR #695, opened 2020-04-10
Add common labels

omegas27:
I am keeping this as a draft because it will probably have a lot of things to do.

What would be the next steps?

### DIFF
--- a/v_6_0_0/files/k8s-resource/calico-all.yaml
+++ b/v_6_0_0/files/k8s-resource/calico-all.yaml
@@ -286,12 +286,13 @@ metadata:
   name: calico-node
   namespace: kube-system
   labels:
+    app.kubernetes.io/name: calico-node
     giantswarm.io/service-type: managed
     k8s-app: calico-node
 spec:
   selector:
     matchLabels:
-      k8s-app: calico-node
+      app.kubernetes.io/name: calico-node
   updateStrategy:
     type: RollingUpdate
     rollingUpdate:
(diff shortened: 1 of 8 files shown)


v_6_0_0/files/k8s-resource/calico-all.yaml
```diff
@@ -286,21 +286,23 @@ metadata:
   name: calico-node
   namespace: kube-system
   labels:
+    app.kubernetes.io/name: calico-node
     giantswarm.io/service-type: managed
```
MarcelMue, line 290: We might also want to set the `app.kubernetes.io/version` label for managed apps. I think it would be helpful information, uncertain if it would be misleading though.

omegas27: I imagine it would be useful but people would have to change that information in multiple places which would at some point not be forgotten. Do you know if we could automate adding that label?

omegas27: ping @MarcelMue when you have some time :)

v_6_0_0/files/k8s-resource/calico-all.yaml
```diff
@@ -286,21 +286,23 @@ metadata:
   name: calico-node
   namespace: kube-system
   labels:
+    app.kubernetes.io/name: calico-node
```
kopiczko, line 289: Please make sure this is communicated. Peoples shell history will stop working.

omegas27: It will keep working as I did not remove the old label :) I will communicate it though

PR #696, opened 2020-04-10
Add AWS CNI variable

tfussell:
`aws-operator@legacy` is using `k8scloudconfig` v6.0.0 which currently assumes that all AWS clusters will use AWS CNI. Since we don't want to backport AWS CNI to non-node pools clusters, I added a new variable to enable AWS CNI. This should be set to `true` in `aws-operator@master`. I will make that corresponding change once this is accepted.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,11 @@ version directory, and  then changes are introduced.
 
 ## [Unreleased]
 
+### Changed
+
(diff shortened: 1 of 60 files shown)


tfussell:
Tested in https://github.com/giantswarm/aws-operator/pull/2250 successfully.

tfussell:
I tested this on aws-operator legacy and master so I'm going with this. Will work on operator PRs next.

PR #697, opened 2020-04-14
v4.9.2 - Copy

stone-z:
v4.9.1 does not include `KubernetesNetworkSetupDocker` as a templated image. Adding v4.9.2 to add this


stone-z:
Going with this

PR #700, opened 2020-04-15
Remove profiling from Controller Manager and Scheduler

stone-z:
Towards https://github.com/giantswarm/giantswarm/issues/4813
Removes the debug profiling endpoint at /debug/pprof for Controller Manager and Scheduler

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,9 @@ versions are frozen. To freeze current version all files are copied to a new
 version directory, and  then changes are introduced.
 
 
-
 ## [Unreleased]
+
+- Disable profiling for Controller Manager and Scheduler
 
 ## [v6.0.2]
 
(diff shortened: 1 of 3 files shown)


stone-z:
Also double-checking with @giantswarm/sig-monitoring that these debug endpoints aren't being scraped or used somewhere for monitoring.

tuommaki:
>  aren't being scraped or used somewhere for monitoring.

Nope. Only occasionally used for debugging resource overuse.

omegas27:
@tfussell was releasing v6.0.3. You might want to merge into it but it looks fine to me. It would be nice to be able to enable it if needed in the future though

stone-z:
> It would be nice to be able to enable it if needed in the future though

Is it helpful enough to be made configurable? If we use it often enough then I could revisit it as an argument. 
I don't think changing it at this level would be super useful because this would be set at node creation time, so if you wanted to turn it on for debugging you'd either have to create a new node or restart the component manually with the flag re-applied (which you can already do).

omegas27:
Maybe doing it manually works in this case, It was just a thought of mine

stone-z:
Sure, if it's worth making it configurable let's revisit

PR #704, opened 2020-04-29
Use Local as NGINX external traffic policy

sslavic:
Towards https://github.com/giantswarm/giantswarm/issues/8877


sslavic:
Dropped in favor of https://github.com/giantswarm/k8scloudconfig/pull/709

PR #708, opened 2020-05-01
1.17 support with backwards compatibility

tfussell:
## Checklist


files/conf/k8s-install-binaries
```diff
@@ -0,0 +1,24 @@
...
+
+declare -a binaries=(kubectl kubelet)
+
+docker create --name hyperkube "$image"
+if [ "$minor" -lt "17" ]; then
+  docker cp hyperkube:/hyperkube /opt/bin/hyperkube
+fi
+for binary in "${binaries[@]}"; do
+  if [ "$minor" -lt "17" ]; then
+    echo "hyperkube $binary "'"$@"' > /opt/bin/"$binary"
```
MarcelMue, line 18: I don't understand how this works / what is happening here. Can you explain it to me?

tfussell: Sure. This creates two wrapper scripts, `/opt/bin/kubectl` and `/opt/bin/kubelet`. They have a single line, `hyperkube kubectl "$@"` and `hyperkube kubelet "$@"` respectively, which just redirects all arguments to `hyperkube` setting the first argument to the name of the `hyperkube` bundled binary. `kubelet --arg1` then becomes `hyperkube kubelet --arg1`.

MarcelMue: Okay - that makes sense to me. Can you maybe add a comment explaining this? It might be only me but this is some serious bash magic to me.

tfussell: Good call--I'll try to spin up a cluster with this, add some comments, then open it up for more review.

MarcelMue:
Is this ready to merge? Or is there something that we still need to do here?

tfussell:
> Is this ready to merge? Or is there something that we still need to do here?

I'm seeing constant restarts on the master. I thought this was the known `torcx` bug, but it doesn't seem to stop (left it running over the weekend) so I'm afraid there's something else going on. I will check if this is happening in 1.16 as well and then escalate to the team.

tfussell:
This has been tested in `kvm-operator`.

PR #709, opened 2020-05-04
Use Local as NGINX external traffic policy

sslavic:
Towards giantswarm/giantswarm#8877

## Checklist


tfussell:
Parent issue is closed. We'll handle this in Rocket if it becomes necessary.

PR #715, opened 2020-05-11
'ha-masters-flattened'

calvix:
For HA masters story

non-HA masters should stay the same and the only change is how the etcd cluster is bootstrapped when the HA masters flag is TRUE.

### DIFF
--- a/files/manifests/k8s-api-server.yaml
+++ b/files/manifests/k8s-api-server.yaml
@@ -27,6 +27,9 @@ spec:
     {{ range .Hyperkube.Apiserver.Pod.CommandExtraArgs -}}
     - {{ . }}
     {{ end -}}
+    {{- if .MultiMasters.Enabled }}
+    - --apiserver-count=3
+    {{- end }}
     - --allow-privileged=true
     - --anonymous-auth=false
     - --insecure-port=0
--- a/pkg/template/cloudconfig.go
+++ b/pkg/template/cloudconfig.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"compress/gzip"
 	"encoding/base64"
+	"fmt"
 	"log"
 	"text/template"
 
@@ -58,7 +59,9 @@ func NewCloudConfig(config CloudConfigConfig) (*CloudConfig, error) {
 	if config.Template == "" {
 		return nil, microerror.Maskf(invalidConfigError, "config.Template must not be empty")
 	}
-
+	if config.Params.MultiMasters.Enabled && config.Params.MultiMasters.EtcdInitialCluster == "" {
+		config.Params.MultiMasters.EtcdInitialCluster = defaultEtcdMultiCluster(config.Params.BaseDomain)
+	}
 	c := &CloudConfig{
 		config:   "",
 		params:   config.Params,
@@ -111,4 +114,8 @@ func (c *CloudConfig) Base64() string {
 
 func (c *CloudConfig) String() string {
 	return c.config
+}
+
+func defaultEtcdMultiCluster(baseDomain string) string {
+	return fmt.Sprintf("etcd1=etcd1.%s,etcd2=etcd2.%s,etcd3=etcd3.%s", baseDomain, baseDomain, baseDomain)
 }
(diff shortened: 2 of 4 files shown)


pkg/template/types.go
```diff
@@ -46,7 +46,9 @@ type Params struct {
 	// cancelled if no progress has been made.
 	ImagePullProgressDeadline string
 	// Container images used in the cloud-config templates
-	Images         Images
+	Images Images
+	// MultiMaster spec
+	MultiMasters   MultiMastersSpec
```
kopiczko, line 51:
```suggestion
	MultiMasters   MultiMasters
```

pkg/template/types.go
```diff
@@ -118,6 +120,16 @@ type HyperkubePodHostMount struct {
 	ReadOnly bool
 }
 
+type MultiMastersSpec struct {
```
kopiczko, line 123:
```suggestion
type MultiMasters struct {
```

pkg/template/cloudconfig.go
```diff
@@ -54,7 +59,9 @@ func NewCloudConfig(config CloudConfigConfig) (*CloudConfig, error) {
 	if config.Template == "" {
 		return nil, microerror.Maskf(invalidConfigError, "config.Template must not be empty")
 	}
-
+	if config.Params.MultiMasters.Enabled && config.Params.MultiMasters.EtcdInitialCluster == "" {
+		config.Params.MultiMasters.EtcdInitialCluster = defaultEtcdMultiCluster(config.Params.BaseDomain)
+	}
```
calvix, line 64: not sure if this si the best place to put the this

kopiczko: Looks good.

pkg/template/types.go
```diff
@@ -118,6 +120,16 @@ type HyperkubePodHostMount struct {
 	ReadOnly bool
 }
 
+type MultiMasters struct {
+	// If set to true, k8scloduconfig will render master template for cluster of 3 masters.
+	// Defaults to false.
+	Enabled bool
+	// Etcd initial cluster is config which define which etcd are members of the cluster.
```
kopiczko, line 127:
```suggestion
	// EtcdInitialCluster is config which define which etcd are members of the cluster.
```

pkg/template/types.go
```diff
@@ -118,6 +120,16 @@ type HyperkubePodHostMount struct {
 	ReadOnly bool
 }
 
+type MultiMasters struct {
+	// If set to true, k8scloduconfig will render master template for cluster of 3 masters.
```
kopiczko, line 124: Godoc should start with the varuable name.

```suggestion
	// Enabled when set to true will cause rendering master template for cluster of 3 masters. Single master otherwise.
```

pkg/template/types.go
```diff
@@ -46,7 +46,9 @@ type Params struct {
 	// cancelled if no progress has been made.
 	ImagePullProgressDeadline string
 	// Container images used in the cloud-config templates
-	Images         Images
+	Images Images
+	// MultiMaster spec
```
kopiczko, line 50:
```suggestion
```

files/manifests/k8s-api-server.yaml
```diff
@@ -27,6 +27,9 @@ spec:
     {{ range .Hyperkube.Apiserver.Pod.CommandExtraArgs -}}
     - {{ . }}
     {{ end -}}
+    {{- if .MultiMasters.Enabled }}
+    - --apiserver-count=3
```
JosephSalisbury, line 31: For my understanding, why do we need to tell the apiserver how many replicas there are?

calvix: I dont know, i would assume it needs it in order to not fuck up things? Its in the config so it might be good idea to set it up :p

```

--apiserver-count int     Default: 1
--
  | The  number of apiservers running in the cluster, must be a positive number.  

```

calvix: the description is not that much usefull

JosephSalisbury: lol, w/e. thanks for pointers <3

PR #718, opened 2020-05-18
Drop index suffix for non-HA etcd

tfussell:
After a little discussion, @calvix and I decided to drop the index suffix from the etcd node name for single master clusters. `etcd1` becomes `etcd`. This helps us with backwards compatibility in other operators where we currently create DNS records like `etcd.<domain>`.

## Checklist

### DIFF
--- a/pkg/template/cloudconfig.go
+++ b/pkg/template/cloudconfig.go
@@ -59,21 +59,21 @@ func NewCloudConfig(config CloudConfigConfig) (*CloudConfig, error) {
 	if config.Template == "" {
 		return nil, microerror.Maskf(invalidConfigError, "config.Template must not be empty")
 	}
+
 	if config.Params.Etcd.NodeName == "" {
 		if config.Params.Etcd.HighAvailability {
+			// We can't guess the node name in this case so must return an error
 			return nil, microerror.Maskf(invalidConfigError,
 				"config.%T must be specified for HA etcd",
 				config.Params.Etcd.NodeName)
 		}
-		config.Params.Etcd.NodeName = nodeName(1)
+		config.Params.Etcd.NodeName = etcdNodeName(1, 1)
 	}
+
 	if config.Params.Etcd.InitialCluster == "" {
-		etcdClusterSize := 1
-		if config.Params.Etcd.HighAvailability {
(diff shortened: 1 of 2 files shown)


kopiczko:
Will it be possible to go from ha to non ha and does it have impact?

tfussell:
> Will it be possible to go from ha to non ha and does it have impact?

Only non-HA to HA will be supported. I asked @calvix about upgrades and we decided that calling the single non-HA etcd node `etcd` (this PR) would be good and that we should number nodes starting from 1 (e.g. `etcd1,etcd2,etcd3`) and the rest will be figured out later.

calvix:
migration from HA to non-HA wont be supported as etcd migration from multi cluster to single is not recommended operation with big posbility to corruption.

pkg/template/cloudconfig.go
```diff
@@ -135,15 +135,26 @@ func (c *CloudConfig) String() string {
 	return c.config
 }
 
-func nodeName(index int) string {
-	return fmt.Sprintf("etcd%d", index)
+func etcdClusterSize(highAvailability bool) int {
+	if highAvailability {
+		return 3
```
MarcelMue, line 140: I think it's a bad idea in the long run to hardcode this here. Fine for now.

calvix:
@tfussell Any mroe changes you want to do or can we merge now?

MarcelMue:
I believe this will be merged today - couldn't do it yesterday because of quay. We can merge already here if needed @calvix - Thomas should be online in 2 or 3 hours.

calvix:
End of today is fine :)

pkg/template/cloudconfig.go
tfussell: I agree. Might want 5 or 7 in the future. I guess we could pass this in as a parameter rather than the boolean `HighAvailability`. We'll work it out as we go.

tfussell:
Testing this now.

tfussell:
Cluster creation with this change worked fine. I am not able to test cluster upgrades on KVM currently. @calvix I will go ahead and merge to unblock your work, but please check upgrades (non-HA to non-HA and non-HA to HA) before releasing in an operator.

PR #722, opened 2020-05-22
dont-delete-master-nodes-on-HA

calvix:
This old part of code is removing the other master nodes in case of node reboot. We don't need it for HA masters and AWS cloud provider should anyway clean old non-existing nodes.

## Checklist

### DIFF
--- a/files/conf/k8s-addons
+++ b/files/conf/k8s-addons
@@ -40,6 +40,7 @@ do
     done
 done
 
+{{- if not .Etcd.HighAvailability }}
 # check for other master and remove it
 THIS_MACHINE=$HOSTNAME
 for master in $(kubectl get nodes --no-headers=true --selector role=master | awk '{print $1}')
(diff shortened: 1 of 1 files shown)


kopiczko:
> for HA masters and AWS cloud provider should anyway clean old non-existing nodes.

Because ASG?

calvix:
> Because ASG? No, that's general cloud provider function, it check api for nodes and if hte  node does not exists in AWS node anymore it will remove it, it does not matter if its instance from ASG or not

kopiczko:
> > Because ASG?
> 
> No, that's general cloud provider function, it check api for nodes and if the node does not exist in AWS node anymore it will remove it, it does not matter if its instance from ASG or not

Why it doesn't work that way for a single master?

calvix:
> Why it doesn't work that way for a single master?

I does, I think this was more for KVM or other providers than AWS

calvix:
tested bot HA and non ha cluster, looks fine

PR #725, opened 2020-05-26
Fix regression in kubelet installation systemd unit

tfussell:
This library now checks the passed in kubernetes version to determine which binaries should be copied out of the `hyperkube` image in the `k8s-extract` service. When I set this to 1.17, this led previously-working operators to stop working when upgrading to k8scloudconfig v6.2.0+ because it wasn't necessary to pass in kubernetes version. To get the 1.17 compatibility, the version needs to be passed in (implemented in AWS and KVM already).

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ version directory, and  then changes are introduced.
 
 ## [Unreleased]
 
+### Fixed
+
+- Regression in kubelet installation for 1.16 clusters.
+
 ## [v6.2.1] 2020-05-26
 
 ### Added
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ version directory, and  then changes are introduced.
 
 ## [Unreleased]
 
+### Fixed
+
+- Regression in kubelet installation systemd unit for 1.16 clusters.
```
kopiczko, line 16:
```suggestion
- Fix regression in kubelet installation systemd unit for 1.16 clusters.
```

PR #726, opened 2020-05-26
Set TLS cipher suites for apiserver

stone-z:
Towards https://github.com/giantswarm/giantswarm/issues/4812

Allows the following TLS ciphers:
```
TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384
TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305
TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305
TLS_RSA_WITH_AES_256_GCM_SHA384
TLS_RSA_WITH_AES_128_GCM_SHA256
```

Based on recommendations from @pipo02mix, [SSLLabs](https://github.com/ssllabs/research/wiki/SSL-and-TLS-Deployment-Best-Practices) and [rancher](https://rancher.com/docs/rancher/v2.x/en/installation/options/tls-settings/).

## Checklist


files/manifests/k8s-api-server.yaml
```diff
@@ -53,6 +53,7 @@ spec:
     - --logtostderr=true
     - --tls-cert-file=/etc/kubernetes/ssl/apiserver-crt.pem
     - --tls-private-key-file=/etc/kubernetes/ssl/apiserver-key.pem
+    - --tls-cipher-suites=TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305,TLS_RSA_WITH_AES_256_GCM_SHA384,TLS_RSA_WITH_AES_128_GCM_SHA256
```
MarcelMue, line 56: This needs to be tested - I am uncertain if the apiserver will be happy with this way of passing them in.

stone-z: I believe you are correct, updated to use the kubelet config files. Still have to try it out

stone-z: Actually, it looks like this method is deprecated for the kubelet but not for apiserver. So I think I need both
https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/

stone-z:
Tested by Fer on `geckon` https://gigantic.slack.com/archives/CHAHY6VMF/p1590652277159200

PR #728, opened 2020-05-27
Remove deprecated resourceContainer from kube-proxy config

tfussell:
This field was removed in https://github.com/kubernetes/kubernetes/pull/78294. This resolves the following error in `kube-proxy` logs (k8s 1.17.6):

```
W0527 16:33:40.443251       1 server.go:427] using lenient decoding as strict decoding failed: strict decoder error for apiVersion: kubeproxy.config.k8s.io/v1alpha1
clientConnection:
kind: KubeProxyConfiguration
mode: iptables
clusterCIDR: 172.24.0.0/16
: v1alpha1.KubeProxyConfiguration.ReadObject: found unknown field: resourceContainer, error found in #10 byte of ...|Container":"/kube-pr|..., bigger context ...|0.0.0:10249","mode":"iptables","resourceContainer":"/kube-proxy"}|...
```

## Checklist


MarcelMue:
What did the flag initially do? Do we need to do something else?

tfussell:
> What did the flag initially do? Do we need to do something else?

As far as I can tell, it used to be normal to run `kube-proxy` on nodes outside of Kubernetes (like `kubelet`). In that case, it needed to be able to be configured to run inside a cgroup. Now most people run it in a DaemonSet as we do so it's implicitly inside the kubernetes cgroup.

PR #729, opened 2020-05-27
Set etcd logger to zap for 3.4 compatibility

tfussell:
The default was capnslog which was deprecated in 3.4. See https://github.com/etcd-io/etcd/issues/11426 Fixes the following warning in `etcd` logs: `[WARNING] Deprecated '--logger=capnslog' flag is set; use '--logger=zap' flag instead`

## Checklist


calvix:
FYI this breaks etcd for etcd version < 3.4

tfussell:
> FYI this breaks etcd for etcd version < 3.4

Thanks for reporting this. I have been working with my team to determine how we should test k8scloudconfig now that we have versions from releases. I guess we'll either add an etcd version check or make it clear that etcd < 3.4 isn't supported in the next release.

PR #730, opened 2020-05-27
Move runtime cgroups config to kubelet flags

tfussell:
`runtimeCgroups` is not allowed in the `kubelet` config file. See https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/kubelet/config/v1beta1/types.go It can still be set via a flag so I'm moving it there.

## Checklist


tfussell:
> Interesting / disturbing. I am certain that it was functional previously.

Same. The systemd unit defines `Slice=kubereserved.slice` so I think this may actually be redundant. I don't see an issue with specifying it in both places though.

PR #735, opened 2020-06-09
enable-calico-metrics-for-policy-only

calvix:
enable metrics endpoint for calico policy only
move calico-policy only manifest to the  new image schema
remove typha deployment from calico-policy only - that is not needed, nto sure why its in the offcicial manifest

towards https://github.com/giantswarm/giantswarm/issues/11336


calvix:
tested creating a few clusters including updates, all green

PR #736, opened 2020-06-09
WIP: use env variables for ETCD clustering

whites11:
WIP

@calvix I'd like to have your early feedback for this PR.

The goal of this PR is making the etcd3.service unit more easily configurable.
By default, it works out of the box as the current `k8scloudconfig` master is working (taking clustering setup data from the provider operator) but it can easily be extended by writing an environment file in `/etc/etcd-cluster-environment` to override the default values.

LMK what you think, thank you

### DIFF
--- a/pkg/template/master_template.go
+++ b/pkg/template/master_template.go
@@ -199,13 +199,13 @@ systemd:
           $IMAGE \
           etcd \
           --name ${ETCD_NAME} \
-          --trusted-ca-file /etc/etcd/server-ca.pem \
-          --cert-file /etc/etcd/server-crt.pem \
-          --key-file /etc/etcd/server-key.pem\
+          --trusted-ca-file=/etc/etcd/server-ca.pem \
+          --cert-file=/etc/etcd/server-crt.pem \
+          --key-file=/etc/etcd/server-key.pem\
           --client-cert-auth=true \
-          --peer-trusted-ca-file /etc/etcd/server-ca.pem \
-          --peer-cert-file /etc/etcd/server-crt.pem \
-          --peer-key-file /etc/etcd/server-key.pem \
+          --peer-trusted-ca-file=/etc/etcd/server-ca.pem \
+          --peer-cert-file=/etc/etcd/server-crt.pem \
+          --peer-key-file=/etc/etcd/server-key.pem \
           --peer-client-cert-auth=true \
           --advertise-client-urls=https://{{ .Cluster.Etcd.Domain }}:{{ .Etcd.ClientPort }} \
           --initial-advertise-peer-urls=http://0.0.0.0:2380 \
--- a/pkg/template/master_template.go
+++ b/pkg/template/master_template.go
@@ -208,7 +208,7 @@ systemd:
           --peer-key-file /etc/etcd/server-key.pem \
           --peer-client-cert-auth=true \
           --advertise-client-urls=https://{{ .Cluster.Etcd.Domain }}:{{ .Etcd.ClientPort }} \
-          --initial-advertise-peer-urls=http://${DEFAULT_IPV4}:2380 \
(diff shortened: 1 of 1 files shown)


pkg/template/master_template.go
```diff
@@ -204,12 +208,12 @@ systemd:
           --peer-key-file /etc/etcd/server-key.pem \
           --peer-client-cert-auth=true \
           --advertise-client-urls=https://{{ .Cluster.Etcd.Domain }}:{{ .Etcd.ClientPort }} \
-          --initial-advertise-peer-urls=https://{{ .Etcd.NodeName }}.{{ .BaseDomain }}:2380 \
+          --initial-advertise-peer-urls=http://${DEFAULT_IPV4}:2380 \
```
whites11, line 211: Is there a reason not to listen on the LAN address here and use the hostname instead?

calvix: Yes this is a problem for AWS, as the DEFAULT IPV4 is not the same ip that is etcd using for connecting

pkg/template/master_template.go
```diff
@@ -204,12 +208,12 @@ systemd:
...
           --listen-client-urls=https://0.0.0.0:2379 \
           --listen-peer-urls=https://0.0.0.0:2380 \
           --initial-cluster-token k8s-etcd-cluster \
-          --initial-cluster {{ .Etcd.InitialCluster }} \
-          --initial-cluster-state {{ .Etcd.InitialClusterState }} \
+          --initial-cluster=${ETCD_INITIAL_CLUSTER} \
+          --initial-cluster-state ${ETCD_INITIAL_CLUSTER_STATE} \
```
calvix, line 216: one uses `=` other uses ` `  which is weird, I would go fro consistency :) Maybe for the whole etcd (so either use `=` everywhere or ` `)

pkg/template/master_template.go
calvix: if you need this for azure than put it into the ENV as well and in azure you can override it

pkg/template/master_template.go
```diff
@@ -194,7 +198,7 @@ systemd:
           --name $NAME \
           $IMAGE \
           etcd \
-          --name {{ .Etcd.NodeName }} \
+          --name ${ETCD_NAME} \
```
calvix, line 201: I would suggest to test his as I fear that systemd itself won't fill the ${} values as you might need to run it whoel in bash wrapper `/bin/bash -c "docker run ...."`.

Or at least I know this was a problem somewhere. (maybe just in defining pod args)

pkg/template/master_template.go
whites11: what if we use 0.0.0.0?

pkg/template/master_template.go
whites11: sure

pkg/template/master_template.go
whites11: I tested this in the last days, it should work.

whites11:
> For my information: what is it going to be used for?

allow a different strategy to setup the etcd cluster in an HA master than the one of passing static cluster data from the provider operator.
I'm experimenting a bit with HA masters on azure and think this would make life easier when we'll implement it.
Have a look here for more details if you want https://drive.google.com/file/d/1iy832bHIpDahMHGVy-oFBvpYG5Vye-3s/view

pkg/template/master_template.go
calvix: that is not a "listen address" that is IP that is sent as Peer URL to other etcds

you need to be able to connect to that

pkg/template/master_template.go
calvix: Cool

pkg/template/master_template.go
calvix: and definitely not http but https

PR #739, opened 2020-06-12
Add quay.io as registry mirror in dockerd config

wtty-fool:
Towards https://github.com/giantswarm/giantswarm/issues/11327

## Checklist

### DIFF
--- a/files/conf/docker-daemon.json
+++ b/files/conf/docker-daemon.json
@@ -1,0 +1,5 @@
+{
+    "registry-mirrors": [
+        "https://quay.io"
+    ]
+}
--- a/files/conf/docker-daemon.json
+++ b/files/conf/docker-daemon.json
@@ -1,5 +1,6 @@
 {
     "registry-mirrors": [
-        "https://quay.io"
+        "https://quay.io",
+        "registry-intl.cn-shanghai.aliyuncs.com"
     ]
 }
--- a/files/conf/docker-daemon.json
+++ b/files/conf/docker-daemon.json
@@ -1,6 +1,5 @@
 {
     "registry-mirrors": [
-        "https://quay.io",
-        "registry-intl.cn-shanghai.aliyuncs.com"
+        "https://quay.io"
     ]
 }
--- a/pkg/template/worker_template.go
+++ b/pkg/template/worker_template.go
@@ -353,7 +353,7 @@ storage:
 
     - path: /etc/docker/daemon.json
       filesystem: root
-      mode: 0600
+      mode: 0644
       contents:
         source: "data:text/plain;charset=utf-8;base64,{{  index .Files "conf/docker-daemon.json" }}"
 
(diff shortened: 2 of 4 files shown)


files/conf/docker-daemon.json
```diff
@@ -0,0 +1,5 @@
+{
+    "registry-mirrors": [
+        "https://quay.io"
+    ]
```
wtty-fool, line 4: I felt adding a list of mirrors to config file was more readable than adding a bunch of flags to `docker.service`. This is because the order of mirrors is v. important here, and it's not immediately apparent when using flags.

pkg/template/worker_template.go
```diff
@@ -351,6 +351,12 @@ storage:
       contents:
         source: "data:text/plain;charset=utf-8;base64,{{  index .Files "conf/10-docker.rules" }}"
 
+    - path: /etc/docker/daemon.json
+      filesystem: root
+      mode: 0600
+      contents:
+        source: "data:text/plain;charset=utf-8;base64,{{  index .Files "conf/docker-daemon.json" }}"
+
```
wtty-fool, line 359: Just for worker now. Should we add mirror for control planes as well?

JosephSalisbury: We need it for tenant cluster masters, sure. In case of a machine going down during an outage.

kopiczko: Let's do all. Why not?

files/conf/docker-daemon.json
kopiczko: Should we also add conditional aliyun?

kopiczko: Or later? CC @JosephSalisbury

wtty-fool: This would cause `dockerd` to resolve in this order (for the record):
1. quay.io
1. aliyun.com
1. docker.io

JosephSalisbury: I'd rather worry about aliyun later - let's get it working with just quay / outside china first

wtty-fool: Ack, reverted commits adding aliyun.

wtty-fool:
I just remembered we are using `k8scloudconfig/v6` everywhere in `aws-operator` and extra magic in its `Dockerfile`. I'll migrate it to "flat" `k8scloudconfig`, but until then let's not merge this one.

kopiczko:
> I just remembered we are using `k8scloudconfig/v6` everywhere in `aws-operator` and extra magic in its `Dockerfile`. I'll migrate it to "flat" `k8scloudconfig`, but until then let's not merge this one.

I don't get this. What needs to be migrated?

wtty-fool:
Nevermind, I got confused by `k8scloudconfig` versioning.

wtty-fool:
Tested on ginger, requires `etcd >= 3.4`

PR #740, opened 2020-06-15
Drop kube-proxy clusterCIDR for Azure as well

tuommaki:
## Checklist

### DIFF
--- a/files/config/kube-proxy.yaml
+++ b/files/config/kube-proxy.yaml
@@ -3,7 +3,7 @@ clientConnection:
   kubeconfig: /etc/kubernetes/config/proxy-kubeconfig.yaml
 kind: KubeProxyConfiguration
 mode: iptables
-{{- if and (ne .Cluster.Kubernetes.CloudProvider "aws") (not .DisableCalico) }}
+{{- if and (and (ne .Cluster.Kubernetes.CloudProvider "aws") (ne .Cluster.Kubernetes.CloudProvider "azure")) (not .DisableCalico) }}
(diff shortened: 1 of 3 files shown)


files/config/kube-proxy.yaml
```diff
@@ -3,7 +3,7 @@ clientConnection:
   kubeconfig: /etc/kubernetes/config/proxy-kubeconfig.yaml
 kind: KubeProxyConfiguration
 mode: iptables
-{{- if and (ne .Cluster.Kubernetes.CloudProvider "aws") (not .DisableCalico) }}
+{{- if and (and (ne .Cluster.Kubernetes.CloudProvider "aws") (ne .Cluster.Kubernetes.CloudProvider "azure")) (not .DisableCalico) }}
```
kopiczko, line 6: Won't that be the same? Do we disable calico on kvm?

```suggestion
{{- if eq .Cluster.Kubernetes.CloudProvider "kvm" }}
```

calvix:
that `if` look crazy, but if there is not an easier solution, let's use that

files/config/kube-proxy.yaml
calvix: there might be a problem with AWS legacy release where we would use this k8scc as on legacy AWS we still sue calico for CNI

tuommaki: Yes, precisely now, but it would be annoying to hunt down when someone adds another provider. I'd rather explicitly exclude the specific known ones and let others decide.

I did have problems parsing that logic in my head as well, but I think it's smaller issue than having to hunt weird problems down on new provider.

kopiczko: 👍

PR #742, opened 2020-06-17
Use short names for bootstrap images

corest:
Towards https://github.com/giantswarm/giantswarm/issues/11326

## Checklist

### DIFF
--- a/files/conf/docker-daemon.json
+++ b/files/conf/docker-daemon.json
@@ -1,5 +1,5 @@
 {
     "registry-mirrors": [
-        "https://quay.io"
+        "https://{{ .RegistryDomain }}"
     ]
 }
--- a/files/conf/wait-for-domains
+++ b/files/conf/wait-for-domains
@@ -1,5 +1,5 @@
 #!/bin/bash
-domains="{{.Cluster.Etcd.Domain}} {{.Cluster.Kubernetes.API.Domain}} registry.hub.docker.com"
+domains="{{.Cluster.Etcd.Domain}} {{.Cluster.Kubernetes.API.Domain}}"
 
 for domain in $domains; do
 until nslookup $domain; do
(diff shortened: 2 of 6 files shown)


files/conf/wait-for-domains
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-domains="{{.Cluster.Etcd.Domain}} {{.Cluster.Kubernetes.API.Domain}} quay.io"
+domains="{{.Cluster.Etcd.Domain}} {{.Cluster.Kubernetes.API.Domain}}"
```
kopiczko, line 2: Why were we waiting for quay?

files/conf/docker-daemon.json
```diff
@@ -1,5 +1,5 @@
 {
     "registry-mirrors": [
-        "https://quay.io"
+        "https://{{ .RegistryDomain }}"
```
kopiczko, line 3: I have mixed feelings if we should configure that. What is the use case? If this is configured then it should be clear that this is the list of mirrors IMO. From variable description I'd try to set it to empty string to use dockerhub.

files/conf/wait-for-domains
corest: dunno

files/conf/docker-daemon.json
kopiczko: Discussed on Slack https://gigantic.slack.com/archives/C010L893ZFW/p1592393492040000.

TL;TR: we need to make AWS China work.

files/conf/wait-for-domains
calvix: this was quite important as it fixed flaky errors that were caused by AWS DNS that was at startup being able to resolve internal AWS DNS names but not public DNS so waiting until it could resolve something from outside was helping  work around such buggy state

This will probabbly cause issues again,  so I dont really agree here with the removal.

kopiczko: It would be good to add it back with a comment and maybe a different domain (because we probably gonna abandon quay).

PR #744, opened 2020-06-18
Expose kubelet health check endpoint

fiunchinho:
## Checklist

I'm testing rolling upgrades on Azure instances. For this, I need a health check endpoint so Azure can tell whether or not an instance is healthy. We thought using the kubelet health endpoint may be a good idea.
Azure makes requests to this endpoint using `localhost`, but we our current configuration we are binding the health endpoint just to the private IP address. With this PR, I'm binding to all IPv4 addresses so we can use both `localhost` and the private IP address.

### DIFF
--- a/files/config/kubelet-worker.yaml.tmpl
+++ b/files/config/kubelet-worker.yaml.tmpl
@@ -2,7 +2,7 @@ kind: KubeletConfiguration
 apiVersion: kubelet.config.k8s.io/v1beta1
 address: ${DEFAULT_IPV4}
 port: 10250
-healthzBindAddress: ${DEFAULT_IPV4}
+healthzBindAddress: "0.0.0.0"
 healthzPort: 10248
 clusterDNS:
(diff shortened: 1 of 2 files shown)


files/config/kubelet-worker.yaml.tmpl
```diff
@@ -2,7 +2,7 @@ kind: KubeletConfiguration
 apiVersion: kubelet.config.k8s.io/v1beta1
 address: ${DEFAULT_IPV4}
 port: 10250
-healthzBindAddress: ${DEFAULT_IPV4}
+healthzBindAddress: "0.0.0.0"
```
kopiczko, line 5: It would be good to drop a comment why we are doing so. Otherwise someone will revert it one day because they will think it's unnecessary. Speaking from experiences: https://github.com/giantswarm/k8scloudconfig/pull/742#discussion_r444879887

PR #747, opened 2020-06-19
Revert registry mirrors over security concerns

wtty-fool:
## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ version directory, and  then changes are introduced.
 
 ## [Unreleased]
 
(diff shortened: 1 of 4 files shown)


jgsqware:
What was the security concern?

wtty-fool:
Public mirrors allow for MITM attack by creating images which fit the resolution. 
For example:
1. The order of registry resolution is quay -> dockerhub
2. You create `quay.io/foo` organization and push malicious image `bar` there
3. If client uses `foo/bar`, hoping to get image from dockerhub, he will run your malicious image instead.

PR #755, opened 2020-07-01
Orphan and adopt pods with problematic selectors

tfussell:
Fixes https://github.com/giantswarm/giantswarm/issues/11154#issuecomment-645578936

## Checklist


files/conf/k8s-addons
```diff
@@ -70,6 +70,18 @@ do
...
+# The k8s-app selector was changed to app.kubernetes.io/name in k8scloudconfig v6.0.3/v6.1.0.
+# This change was reverted in k8scloudconfig v6.1.1+.
+# matchLabels is immutable in apps/v1 so `kubectl apply -f kube-proxy-ds.yaml` will fail
+# when upgrading clusters using k8scloudconfig v6.0.3/v6.1.0.
+# To mitigate this without causing downtime, we can orphan the existing pods with `kubectl delete --cascade=false`
+# The pods will continue to run and the new daemonset will inherit the pods so they can be
+# upgraded if needed in a graceful manner.
+if [[ $(kubectl get ds -n kube-system kube-proxy -o jsonpath='{.spec.selector.matchLabels}') == "map[app.kubernetes.io/name:kube-proxy]" ]]; then
+  echo "Orphaning kube-proxy pods to be inherited by new deployment"
+  kubectl delete ds -n kube-system kube-proxy --cascade=false
```
kopiczko, line 82: How is it supposed to work. Are the labels for the new selector already present on the pods?

tfussell: Yes, here was the PR https://github.com/giantswarm/k8scloudconfig/pull/695/files You can see that the new label was added but the old label, `k8s-app`, wasn't removed.

tfussell:
I just updated this branch and fixed some broken links in the CHANGELOG. The code is unchanged. I just confirmed that aws-operator and azure-operator didn't have any releases with the problematic selector label `app.kubernetes.io/name` which means this should only affect KVM. I have tested upgrades on KVM so I believe this is safe to merge and release.

PR #757, opened 2020-07-02
files, template: allow setting registry mirrors

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/11736

## Checklist


files/conf/docker-daemon.json
```diff
@@ -0,0 +1,7 @@
+{
+    "registry-mirrors": [
+        {{- range $i, $e := .RegistryMirrors }}
```
corest, line 3: download (9)
 is it only for me formatting looks like this?

kopiczko: It's for everyone. This is probably because it isn't a valid JSON before templating.

kopiczko:
Sorry for the noise. I'm turning this into draft and let you know when I think it's ready.

kopiczko:
Tested here https://github.com/giantswarm/aws-operator/pull/2572

PR #760, opened 2020-07-03
template: validate Images registry domain

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/11736

This change ensures that all images in `Parameters.Images` has the same
registry domain as set in `Parameters.RegistryDomain`.

## Checklist


kopiczko:
I have a different idea. `.RegistryDomain` can be removed. We can verify that images don't start with anything else than `docker.io` if mirrors are set.

PR #766, opened 2020-07-15
add monitoring labels for kube-proxy, k8s-scheduler, k8s-controller-manager

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/11724

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
+### Added
+
+- Add monitoring annotations `prometheus.io/*` to kube-proxy, k8s-scheduler, k8s-controller-manager and calico.
+
 ## [8.0.0] - 2020-08-11
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,10 +17,14 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 - Added validation of versions in the cloud config Params struct. Versions outside of supported ranges will cause an
   error to be returned from cloud config-related functions.
+<<<<<<< HEAD
   
 ### Changed
 
 - Updated backward incompatible Kubernetes dependencies to v1.18.5.
+=======
+- Add monitoring annotations `prometheus.io/*` and `giantswarm.io/monitoring*` to kube-proxy, k8s-scheduler, k8s-controller-manager and calico.
+>>>>>>> 18c862e... add-missing-monitoring-annotation-to-changelog
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ### Added
 
-- Add monitoring annotations `prometheus.io/*` to kube-proxy, k8s-scheduler, k8s-controller-manager and calico.
+- Add monitoring annotations `prometheus.io/*` and `giantswarm.io/monitoring*` to kube-proxy, k8s-scheduler, k8s-controller-manager and calico.
 
 ## [8.0.0] - 2020-08-11
 
@@ -17,14 +17,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 - Added validation of versions in the cloud config Params struct. Versions outside of supported ranges will cause an
   error to be returned from cloud config-related functions.
-<<<<<<< HEAD
   
 ### Changed
 
 - Updated backward incompatible Kubernetes dependencies to v1.18.5.
-=======
-- Add monitoring annotations `prometheus.io/*` and `giantswarm.io/monitoring*` to kube-proxy, k8s-scheduler, k8s-controller-manager and calico.
->>>>>>> 18c862e... add-missing-monitoring-annotation-to-changelog
 
 ### Removed
 
--- a/files/k8s-resource/calico-all.yaml
+++ b/files/k8s-resource/calico-all.yaml
@@ -309,6 +309,11 @@ spec:
         # priority scheduling and that its resources are reserved
         # if it ever gets evicted.
         scheduler.alpha.kubernetes.io/critical-pod: ''
+        prometheus.io/scrape: "true"
+        prometheus.io/port: "9091"
+        giantswarm.io/monitoring: "true"
+        giantswarm.io/monitoring-path: "/metrics"
+        giantswarm.io/monitoring-port: "9091"
     spec:
       nodeSelector:
         kubernetes.io/os: linux
--- a/files/k8s-resource/calico-all.yaml
+++ b/files/k8s-resource/calico-all.yaml
@@ -303,6 +303,7 @@ spec:
         app.kubernetes.io/name: calico-node
         giantswarm.io/service-type: managed
         k8s-app: calico-node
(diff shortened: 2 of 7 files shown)


mig4:
Regarding waiting for tenant clusters to be upgraded, our rough idea was to also have a _"legacy"_ selector which would use any existing labels if possible but would be mutually exclusive with the monitor using the new labels.

It may not be possible to have a single _"legacy"_ `ServiceMonitor` if there's no single label on all components that we could use, but we should be able to do it with a handful of them.

To explain this a bit, this labels approach is something we came up with to avoid having to create a `ServiceMonitor` for every single target we want to monitor for two reasons:

1. it would be a bit of a maintenance nightmare
2. we'd like a more dynamic system where if we deploy a new service it will be monitored automatically as long as it's labelled correctly without having to create a `ServiceMonitor` for it in PMO

So we were thinking to have a primary `ServiceMonitor` with a `matchLabels` `selector` that uses `giantswarm.io/monitoring` label to find (mostly) everything that it needs to monitor.

Granted. the labels are less flexible than `ServiceMonitor`s but we should be able to cover ~80% cases with them and use additional `ServiceMonitor`s for the rest.

omegas27:
I definitely agree on having a generic solution and add necessary service monitor if we need them.

As we discussed, maybe we could create services for the components we want to monitor adding the labels we want in Kubernetes but that would mean we have access to all the tenant clusters prometheus the same way the prometheus operator is creating a service for the kubelet: https://github.com/helm/charts/blob/f019fd53a81e7124df65be0d4da620795c251a2e/stable/prometheus-operator/templates/prometheus-operator/deployment.yaml#L46

omegas27:
@giantswarm/team-atlas would you mind taking another look?

JosephSalisbury:
dumbass question: `giantswarm.io/monitoring` is the managed apps label, right?

mig4:
> dumbass question: `giantswarm.io/monitoring` is the managed apps label, right?

It is. I was trying to think if using the same label for "system" components we want to monitor may cause any problems but I can't think of any at the moment. Semantically it also seems all right to have them aligned as both system components and managed apps are managed by us. WDYT?

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 - Added validation of versions in the cloud config Params struct. Versions outside of supported ranges will cause an
   error to be returned from cloud config-related functions.
+- Add monitoring annotations `prometheus.io/*` to kube-proxy, k8s-scheduler, k8s-controller-manager and calico.
```
mig4, line 14: We'll need to update the prefix here.

files/k8s-resource/calico-all.yaml
```diff
@@ -309,6 +309,11 @@ spec:
         # priority scheduling and that its resources are reserved
         # if it ever gets evicted.
         scheduler.alpha.kubernetes.io/critical-pod: ''
+        prometheus.io/scrape: "true"
+        prometheus.io/port: "9091"
```
mig4, line 313: These two should probably be dropped in favour of the ones below.

files/k8s-resource/calico-all.yaml
```diff
@@ -309,6 +309,11 @@ spec:
...
+        giantswarm.io/monitoring: "true"
```
mig4, line 314: Note that this will need to be a label instead of an annotation, see note in the [draft doc](https://github.com/giantswarm/giantswarm/blob/245a6a193e7e0f14b87563c139e6e31b8966be8c/content/docs/architecture-specs-adrs/specs/configuration-of-targets-in-tc-prometheus.md#labels-and-annotations).

files/k8s-resource/calico-all.yaml
```diff
@@ -309,6 +309,11 @@ spec:
...
+        giantswarm.io/monitoring-path: "/metrics"
```
mig4, line 315: IIRC this shouldn't start with a `/`, I'll double-check later, but see [monitoring of apps](https://intranet.giantswarm.io/docs/monitoring/monitoring_of_apps/).

files/k8s-resource/calico-all.yaml
omegas27: I am really adding the default ones as well as some customers might want to observe calico with their own prometheus if they experience networking issues. WDYT?

files/k8s-resource/calico-all.yaml
omegas27: I think we need to adjust here as prometheus is scraping according to annotations only:
https://www.weave.works/docs/cloud/latest/tasks/monitor/configuration-k8s/#per-pod-prometheus-annotations Also, the path starts with a /

files/k8s-resource/calico-all.yaml
mig4: OK, yeah it wouldn't hurt 👍

files/k8s-resource/calico-all.yaml
mig4: Yeah _Prometheus_ itself is only using annotations, because internally it just gets a list of all resources and then filters it and applies relabelings to drop some according to expressions, etc. to narrow it down further.

We can't do that because we're using a selector to only get a subset of resources, and with selector you need a label, won't work with annotations.

As for the path, I suppose we can adjust our replacement expression so it expects the `/`, the only thing is I don't know how many managed apps we have already labelled without it and so how many we'd need to change.

CHANGELOG.md
mig4: Thanks!

files/k8s-resource/calico-all.yaml
omegas27: Alright, let me change it to a label. I am not sure they added this annotations in a lot of places though but maybe we can support both schemes?

mig4: Do you mean to have `giantswarm.io/monitoring` as both annotation and a label? That could work too, and for apps it would mean they work with the current `g8s-prometheus` and our new _TC Prometheus_.

But for system components like calico, kube-proxy, etc. it would be good to test that this doesn't break the current `g8s-prometheus` setup, I have a notion it may cause errors because they would be picked up by two scrape configs - the custom workload one that matches system components, as well as the managed apps one.

omegas27: I was talking about the /metrics path, sorry for the confusion :)

omegas27: Yes, we still need to test it first. I will try tomorrow morning but testing k8scc is a pain

mig4: Oh, right, with regards to monitoring path, it should work either way cause we rewrite that to `/api/v1/namespaces/${NS}/pods/${POD}:${PORT}/proxy/${PATH}` so if PATH starts with `/` it will just end up being doubled, e.g. `/api/v1/namespaces/${NS}/pods/${POD}:${PORT}/proxy//metrics`, which still works, it's just not as neat.

mig4: Actually I also just realised that we don't need this at all here because `/metrics` is the default path.

omegas27: Then we just need to test it now :)

omegas27 (approved):

omegas27:
@giantswarm/team-atlas can we merge this one?

mig4 (approved):
LGTMIIWFY

PR #767, opened 2020-07-20
Remove v6 reference

whites11:
This PR updates a local module reference that was still using the v6 release.


whites11:
Tested on azure, all seems to be working fine

PR #769, opened 2020-07-21
Configure Docker in one place

omegas27:

### DIFF
--- a/files/conf/docker-daemon.json
+++ b/files/conf/docker-daemon.json
@@ -3,5 +3,10 @@
         {{- range $i, $e := .RegistryMirrors }}
         {{- if ne $i 0 }},{{ end }}{{ "\n" }}        "https://{{ $e }}"
         {{- end }}
-    ]
+    ],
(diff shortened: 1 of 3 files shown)


files/conf/docker-daemon.json
```diff
@@ -3,5 +3,10 @@
         {{- range $i, $e := .RegistryMirrors }}
         {{- if ne $i 0 }},{{ end }}{{ "\n" }}        "https://{{ $e }}"
         {{- end }}
-    ]
+    ],
+    "metrics-addr" : "0.0.0.0:9323",
```
MarcelMue, line 7: Should be fine but please test it. When you create a tenant cluster with this change, then you should still be able to curl this port from the node.

omegas27: I will try following https://gigantic.slack.com/archives/CHAHY6VMF/p1595338496088700

omegas27:
Tested in Gauss

PR #772, opened 2020-07-23
set-etcd-data-dir-permissions-to-0700

calvix:
etcd 3.4.10 needs 0700 permission on the data dir
https://github.com/etcd-io/etcd/blob/master/CHANGELOG-3.4.md

chmod is the only way as we need to ensure older clusters will have this fixed as well.


calvix:
tested, etcd is happy now

PR #776, opened 2020-07-27
Improve validation of Params

tfussell:
Towards https://github.com/giantswarm/giantswarm/issues/12005

## Checklist


pkg/template/cloudconfig.go
```diff
@@ -28,29 +25,6 @@ type CloudConfigConfig struct {
...
 }
 
-func DefaultCloudConfigConfig() CloudConfigConfig {
-	return CloudConfigConfig{
-		Params:   Params{},
-		Template: "",
-	}
-}
-
-func DefaultParams() Params {
```
tfussell, line 38: This feels like a problematic function for me. It allows operators to use the library without considering all of the parameters that are passed in. For this reason, I decided to remove it while working on this PR. Interested to hear what others think.

pkg/key/key.go
```diff
@@ -0,0 +1,9 @@
+package key
+
+const (
+	CalicoVersionConstraint                       = ">= 3.10.0 < 3.15.0"
+	EtcdVersionConstraint                         = ">= 3.4.0 < 3.5.0"
+	KubernetesVersionConstraint                   = ">= 1.16.0 < 1.18.0"
```
MarcelMue, line 6: Why strictly lower than 1.18?
We already kind of successfully tested it. IMO most constraints should only enforce a minimum version - we want to test newer versions as easily as possible.

pkg/template/cloudconfig_test.go
```diff
@@ -55,8 +58,17 @@ func TestCloudConfig(t *testing.T) {
 		if err != nil {
 			t.Errorf("failed to render ignition files, %v:", err)
 		}
+		tc.params.Extension = nopExtension{}
 		tc.params.Files = files
-		tc.params.Images = BuildImages("docker.io", Versions{})
+		tc.params.Versions = Versions{
```
tfussell, line 63: I don't really like arbitrarily defining these versions in the middle of the test, but it will be clear enough when these need to be updated if we change constraints and this test starts to fail I guess.

pkg/template/cloudconfig_test.go
```diff
@@ -55,8 +58,17 @@ func TestCloudConfig(t *testing.T) {
...
+			Calico:                       "3.14.1",
+			CRITools:                     "1.17.0",
```
MarcelMue, line 65: Why default a different version here than kubernetes if we use the same version constraint to verify?

pkg/template/validation_test.go
```diff
@@ -0,0 +1,119 @@
...
+	Calico:                       "3.9.1",
+	CRITools:                     "1.15.0",
+	Etcd:                         "3.3.15",
+	Kubernetes:                   "1.15.5",
+	KubernetesAPIHealthz:         "0.1.0",
+	KubernetesNetworkSetupDocker: "0.1.0",
+}
+
+func editVersions(versions Versions, fieldName string, version string) Versions {
+	entityType := reflect.TypeOf(versions)
```
tfussell, line 31: This function allows a field in the struct to be edited by name saving a good bit of boilerplate as we want all fields to be valid except for one in each test so we're checking the right thing.

pkg/template/cloudconfig.go
MarcelMue: We got rid of these functions in other places - so totally fine with me.

pkg/key/key.go
tfussell: My thinking is that we want to constrain versions to those that we have tested and already ensured that there are no errors. We found 3 warnings in component logs when testing 1.18 which I plan to fix today and release with wider version constraints. Worth discussing how we expect this to work though--changing constraints will require releasing a new operator which is extra work.

MarcelMue: I get your thought process - I just want to prevent the extra effort, when our team will be the ones raising these versions most of the time.
Limiting here makes it harder for us to try out new versions which I want to prevent at all cost :D

pkg/template/cloudconfig_test.go
tfussell: CRI-tools is kind of weird in that they only release one per Kubernetes minor, so 1.16.0, 1.17.0, etc. See https://github.com/kubernetes-sigs/cri-tools/releases I wanted a real version here in case we want to add more validation in code in the future. It feels arbitrary either way.

pkg/template/cloudconfig.go
tfussell: Sounds good. I will also add this decision to `fmt` so we can stay aligned across projects.

pkg/key/key.go
tfussell: I agree. It will limit our ability to automatically or manually test new Kubernetes minors. I would like to keep the maximum version constraint for etcd and calico as it will force us to think about possible issues and test first, but I don't mind dropping the maximum for Kubernetes here.

pkg/template/cloudconfig_test.go
tfussell: Actually there is a cri-tools@1.16.1 so even that rule doesn't hold. I'll just make it match the Kubernetes version for now--maybe we can add a check that the image exists in the docker registry in the future.

pkg/key/key.go
```diff
@@ -0,0 +1,9 @@
+package key
+
+const (
+	CalicoVersionConstraint                       = ">= 3.10.0 < 3.15.0"
+	EtcdVersionConstraint                         = ">= 3.4.0 < 3.5.0"
```
xh3b4sd, line 5: Are these constraints universally valid? I could see that we have quite different requirements across providers. Also IIRC for HA Masters we needed `3.4.10` or something like that. Maybe Mr @calvix can check back on this. I would further appreciate some godoc to explain what these constants are about just to provide more context.

tfussell: These constraints should allow etcd `3.4.10`. I did want to open this up to all of SIG Operator to make sure I wasn't missing any use cases though--just let me know. And I'll add some godoc comments in a bit.

pkg/key/key.go
```diff
@@ -0,0 +1,9 @@
+package key
+
+const (
+	CalicoVersionConstraint                       = ">= 3.10.0 < 3.15.0"
```
stone-z, line 4: Azure 12.0.0 is already calico 3.15.1, next 12.x releases for other providers will be as well

pkg/key/key.go
tfussell: Done.

pkg/key/key.go
tfussell: I added a comment for this block of constants.

pkg/key/key.go
tfussell: Good call. I changed this to allow 3.15.

pkg/key/key.go
tfussell: I checked AWS 11.5.0 and 12.0.0 and this constraint should work for both (etcd 3.4.9 and 3.4.10 respectively).

Issue #10, opened 2017-03-14
Support for SSH public keys


nhlfr:
We need to inject SSH keys into cloudconfig.

Most probably, we will do it by providing ssh keys via extension, so the external operators can fill it with data received from cert-operator. Since cert-operator is not implemented yet, the aws- and kvm-operator can temporarily provide the keys by themselves.

teemow:
Even more: There should be distinct users. Nobody should connect with the core user. Everybody should have it's own user.

It would be awesome if we could hook this up to dex and read users and keys from there. 

ping @puja108

puja108:
+1 we need distinct users, on the one side it's about managing access, on the other side it's about being able to identify who did what (i.e. for audits)

I don't think Dex has that kind of functionality, it's pretty dumb from what I've seen, we would need a real IdM solution for that kind of foo. Or sth interfacing with Vault or so that gives us revokability of keys maybe?

asymmetric:
Does this issue need to be specced more? It's not clear to me if it's actionable at this point, i.e. should we expand the issue to using multiple users?

puja108:
Yes, might need some minor speccing at least as @teemow was also mentioning maybe using the ssh backend of Vault for this.

puja108:
The comment I meant from Timo:
> just stumbled upon the ssh backend in vault. let's keep this in mind. would be totally awesome to have this in vault too. so once you are in the github group you can connect to the vaults and issue yourself certificates and ssh keys

webwurst:
We could use GitHub to read users and keys like https://github.com/webwurst.keys.

puja108:
Keybase might also be an option?

calvix:
@puja108 @teemow @xh3b4sd 

We need to continue on this.
Separate OS account for each person is requested by security (because of audit reasons).
So we need to have a way how to provide list of users and their ssh to the k8scloudconfig.

xh3b4sd:
[...]

PR #780, opened 2020-07-30
Remove old scheduler configs

tfussell:
Towards https://github.com/giantswarm/giantswarm/issues/12349. Kubernetes 1.16 adds strict decoding of the scheduler config which surfaced that we are setting some fields removed in Kubernetes 1.13. This leads to a warning and slightly less safe config loading. I'm deleting the deleted fields and those where the value matches the default.

## Checklist


files/config/scheduler.yaml
```diff
@@ -1,8 +1,4 @@
-kind: KubeSchedulerConfiguration
-algorithmSource:
-  provider: DefaultProvider
 apiVersion: kubescheduler.config.k8s.io/v1alpha1
+kind: KubeSchedulerConfiguration
 clientConnection:
   kubeconfig: /etc/kubernetes/kubeconfig/scheduler.yaml
-failureDomains: kubernetes.io/hostname,failure-domain.beta.kubernetes.io/zone,failure-domain.beta.kubernetes.io/region
-hardPodAffinitySymmetricWeight: 1
```
tfussell, line 8: Moved to scheduler policy config in https://github.com/kubernetes/kubernetes/pull/44159

files/config/scheduler.yaml
```diff
@@ -1,8 +1,4 @@
-kind: KubeSchedulerConfiguration
-algorithmSource:
-  provider: DefaultProvider
 apiVersion: kubescheduler.config.k8s.io/v1alpha1
+kind: KubeSchedulerConfiguration
 clientConnection:
   kubeconfig: /etc/kubernetes/kubeconfig/scheduler.yaml
-failureDomains: kubernetes.io/hostname,failure-domain.beta.kubernetes.io/zone,failure-domain.beta.kubernetes.io/region
```
tfussell, line 7: Removed in k8s 1.13 in change from `componentconfig/v1alpha1` to `kubescheduler.config.k8s.io/v1alpha1`.

files/config/scheduler.yaml
```diff
@@ -1,8 +1,4 @@
-kind: KubeSchedulerConfiguration
-algorithmSource:
```
tfussell, line 2: Deprecated, defaults to `DefaultProvider`.

tfussell:
Tested on `geckon`. Before change:
```
I1124 12:16:58.250247       1 serving.go:312] Generated self-signed cert in-memory
W1124 12:16:58.255259       1 configfile.go:65] using lenient decoding as strict decoding failed: strict decoder error for kind: KubeSchedulerConfiguration
algorithmSource:
  provider: DefaultProvider
apiVersion: kubescheduler.config.k8s.io/v1alpha1
clientConnection:
  kubeconfig: /etc/kubernetes/kubeconfig/scheduler.yaml
failureDomains: kubernetes.io/hostname,failure-domain.beta.kubernetes.io/zone,failure-domain.beta.kubernetes.io/region
hardPodAffinitySymmetricWeight: 1
: v1alpha1.KubeSchedulerConfiguration.HardPodAffinitySymmetricWeight: ReadObject: found unknown field: failureDomains, error found in #10 byte of ...|reDomains":"kubernet|..., bigger context ...|netes/kubeconfig/scheduler.yaml"},"failureDomains":"kubernetes.io/hostname,failure-domain.beta.kuber|...
W1124 12:16:58.866815       1 authentication.go:267] No authentication-kubeconfig provided in order to lookup client-ca-file in configmap/extension-apiserver-authentication in kube-system, so client certificate authentication won't work.
W1124 12:16:58.872418       1 authentication.go:291] No authentication-kubeconfig provided in order to lookup requestheader-client-ca-file in configmap/extension-apiserver-authentication in kube-system, so request-header client certificate authentication won't work.
```
after change:
```
I1124 14:33:02.274304       1 flags.go:33] FLAG: --write-config-to=""
I1124 14:33:03.554689       1 serving.go:312] Generated self-signed cert in-memory
W1124 14:33:05.327813       1 authentication.go:267] No authentication-kubeconfig provided in order to lookup client-ca-file in configmap/extension-apiserver-authentication in kube-system, so client certificate authentication won't work.
W1124 14:33:05.327884       1 authentication.go:291] No authentication-kubeconfig provided in order to lookup requestheader-client-ca-file in configmap/extension-apiserver-authentication in kube-system, so request-header client certificate authentication won't work.
```

MarcelMue (approved):

calvix (approved):
lgtm

yulianedyalkova:
Did we made sure that the behaviour that those flags were defining is preserved (by other flags/options)?

tfussell:
> Did we made sure that the behaviour that those flags were defining is preserved (by other flags/options)?

I think we can tell from the code and the warning message (`found unknown field: failureDomains, error found in #10 byte of ...|reDomains":"kubernet`) that the config here wasn't being used. It's worth looking into whether we should be setting this  on a global or pod-specific level. I think the idea is to use the built-in `requiredDuringSchedulingIgnoredDuringExecution` option for `nodeAffinity` on pods when needed.

yulianedyalkova:
I'm not questioning this PR, I'm just assuming we added those flags for a reason in the first place. And it would be nice to eventually evaluate if we still care about them and if there are alternative ways to ensure the behaviour stays the same :) I have no idea what the flags are doing though so I could be talking nonsense 🤷‍♀️

tfussell:
> I'm not questioning this PR, I'm just assuming we added those flags for a reason in the first place. And it would be nice to eventually evaluate if we still care about them and if there are alternative ways to ensure the behaviour stays the same :) I have no idea what the flags are doing though so I could be talking nonsense 🤷‍♀️

For sure. These were deprecated several years ago so it's hard to tell what they did exactly. I tried reading through the Kubernetes git history, but they don't talk about why, only what. My guess is that the scheduler used to handle failure domains in a special way from other affinity/anti-affinity (if affinity even existed back then) but it has been subsumed by new features.

PR #789, opened 2020-09-14
Remove init container that installs CNI binaries

paurosello:
## Checklist

The CNI running on a K8S should be in charge of installing it's own CNI binaries (portmap, bridge...)

Currently Calico and AWS CNI where fighting to setup this files:
https://github.com/giantswarm/giantswarm/issues/11077#issuecomment-685457567

And were also causing problems as nodes were being marked as Ready when AWS CNI was not even able to boot:
https://github.com/giantswarm/giantswarm/issues/13086

@whites11 not sure if this file is also being used in Azure, would this change still be fine?

### DIFF
--- a/files/k8s-resource/calico-policy-only.yaml
+++ b/files/k8s-resource/calico-policy-only.yaml
@@ -388,6 +388,7 @@ spec:
           securityContext:
             privileged: true
 {{ end -}}
+
       containers:
         # Runs calico-node container on each Kubernetes node.  This
         # container programs network policy and routes on each
--- a/files/k8s-resource/calico-policy-only.yaml
+++ b/files/k8s-resource/calico-policy-only.yaml
@@ -351,6 +351,7 @@ spec:
       # deletion": https://kubernetes.io/docs/concepts/workloads/pods/pod/#termination-of-pods.
       terminationGracePeriodSeconds: 0
       priorityClassName: system-node-critical
+{{ if not .EnableAWSCNI -}}
       initContainers:
         # This container installs the CNI binaries
         # and CNI network config file on each node.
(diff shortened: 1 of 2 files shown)


njuettner (approved):

calvix:
I thnik this could cause trouble for Azure if they want to use  the calico-policy only manifest from k8scc ( they had one of their own in operator).

One option would be a flag that would enable/disable this

paurosello:
Yes, although I think the Azure CNI might be installing this binaries too, not sure...

calvix:
There is no `azure CNI` pod or app running on Azure cluster, k8s simply assign IP from a predefined range and the azure cloud provider adds routes for each node to the  VNET so there is nothing that would install it on each node.

files/k8s-resource/calico-policy-only.yaml
```diff
@@ -351,41 +351,6 @@ spec:
       # deletion": https://kubernetes.io/docs/concepts/workloads/pods/pod/#termination-of-pods.
       terminationGracePeriodSeconds: 0
       priorityClassName: system-node-critical
-      initContainers:
-        # This container installs the CNI binaries
-        # and CNI network config file on each node.
-        - name: install-cni
-          image: {{ .Images.CalicoCNI }}
```
kopiczko, line 358: Can also the image configuration be removed?

calvix: image is still necessary for KVM ( and probabbly azure)

paurosello: I have disabled only for AWS CNI now

calvix (approved):

njuettner (approved):

PR #793, opened 2020-09-17
Fix line break when selecting HA

fiunchinho:
I was getting a weird line break, so I tested the template in the playground and I belive it's not correct. You can see it in action here https://play.golang.org/p/SNTCNgB4mHp

## Checklist

### DIFF
--- a/files/manifests/k8s-api-server.yaml
+++ b/files/manifests/k8s-api-server.yaml
@@ -26,9 +26,9 @@ spec:
     {{ range .Kubernetes.Apiserver.CommandExtraArgs -}}
     - {{ . }}
     {{ end -}}
-    {{- if .Etcd.HighAvailability }}
+    {{- if .Etcd.HighAvailability -}}
     - --apiserver-count=3
-    {{- end }}
+    {{ end -}}
     - --allow-privileged=true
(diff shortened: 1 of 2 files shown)


calvix (approved):
lgtm iwfy

files/manifests/k8s-api-server.yaml
```diff
@@ -26,9 +26,9 @@ spec:
     {{ range .Kubernetes.Apiserver.CommandExtraArgs -}}
     - {{ . }}
     {{ end -}}
-    {{- if .Etcd.HighAvailability }}
+    {{- if .Etcd.HighAvailability -}}
```
tfussell, line 29: This doesn't collapse too much whitespace (e.g. putting multiple flags on one line) when used with or without `.Kubernetes.Apiserver.CommandExtraArgs`?

fiunchinho: I think it's fine. Take a look at the updated example https://play.golang.org/p/DxKxmnrURtG

tfussell: Cool, thanks.

tfussell (approved):

PR #795, opened 2020-09-29
Allow parallel image pulls in kubelet

yulianedyalkova:
## Checklist

### DIFF
--- a/files/config/kubelet-master.yaml.tmpl
+++ b/files/config/kubelet-master.yaml.tmpl
@@ -44,3 +44,6 @@ tlsCipherSuites:
   - TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305
   - TLS_RSA_WITH_AES_256_GCM_SHA384
   - TLS_RSA_WITH_AES_128_GCM_SHA256
+serializeImagePulls: false
+registryPullQPS: 2
+registryBurst: 3
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
+### Changed
+
+- Allow parallel download of images in kubelet templates
+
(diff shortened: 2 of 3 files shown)


files/config/kubelet-master.yaml.tmpl
```diff
@@ -44,3 +44,6 @@ tlsCipherSuites:
   - TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305
   - TLS_RSA_WITH_AES_256_GCM_SHA384
   - TLS_RSA_WITH_AES_128_GCM_SHA256
+serializeImagePulls: false
```
MarcelMue, line 47: Can we do this in kind of alphabetical order? Even though the rest of the file is just kind of alphabetical

yulianedyalkova: I find this a lot easier to follow logically than having them ordered alphabetically and it keeps the consistency with the templates in [giantnetes-terraform](https://github.com/giantswarm/giantnetes-terraform/blob/master/templates/files/config/kubelet-master.yaml.tmpl) (I don't know if that's intentional so far). But I can change it if you really hate it.

MarcelMue: I am fine with keeping the order - I wonder why you find it easier to follow, as I don't really see the logic :sweat_smile:

MarcelMue (approved):

files/config/kubelet-master.yaml.tmpl
yulianedyalkova: Since they more or less configure the same thing, it's just nice to have them grouped together than spread all over the whole file. And then the order within the group - you just enable a feature and then fine-tune it. Maybe? 😄

MarcelMue: Okay I can understand wanting them in roughly the same place. Shame we can't structure this whole thing nicer though.

Anyway, please ask for review of sig-infrastructure.

corest (approved):

PR #799, opened 2020-10-02
k8s-label-node-timer

calvix:
## Checklist

# context
towards https://github.com/giantswarm/giantswarm/issues/13510

to ensure the important labels are always set even in case the customer does some bad voodoo

### DIFF
--- a/pkg/template/master_template.go
+++ b/pkg/template/master_template.go
@@ -355,7 +355,6 @@ systemd:
       Wants=k8s-kubelet.service
       [Service]
       Type=oneshot
-      RemainAfterExit=yes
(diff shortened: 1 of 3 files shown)


pkg/template/master_template.go
```diff
@@ -355,7 +355,6 @@ systemd:
       Wants=k8s-kubelet.service
       [Service]
       Type=oneshot
-      RemainAfterExit=yes
```
calvix, line 358: this has to be removed so timer see it as (exited/inactive)  otherwise timer would not do anything

whites11 (approved):
I love this

njuettner (approved):
goo catch

PR #801, opened 2020-10-05
Update version compatibility and certs location

tfussell:
## Checklist

- [x] Update changelog in CHANGELOG.md.

PR #808, opened 2020-10-20
Update kubelet calico etcd cert paths

tfussell:
Split from https://github.com/giantswarm/k8scloudconfig/pull/801

## Checklist

### DIFF
--- a/pkg/template/master_template.go
+++ b/pkg/template/master_template.go
@@ -61,7 +61,7 @@ systemd:
       After=k8s-kubelet.service k8s-setup-network-env.service
       [Service]
       Type=oneshot
-      ExecStart=/bin/sh -c "find /etc/kubernetes/ssl -name '*.pem' -print | xargs -i  sh -c 'chown root:giantswarm {} && chmod 640 {}'"
+      ExecStart=/bin/sh -c "find /etc/kubernetes/ssl -type f -print | xargs -i  sh -c 'chown root:giantswarm {} && chmod 640 {}'"
       [Install]
       WantedBy=multi-user.target
   - name: wait-for-domains.service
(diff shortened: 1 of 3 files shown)


pkg/template/master_template.go
```diff
@@ -300,9 +300,9 @@ systemd:
       Slice=kubereserved.slice
       CPUAccounting=true
       MemoryAccounting=true
-      Environment="ETCD_CA_CERT_FILE=/etc/kubernetes/ssl/etcd/server-ca.pem"
-      Environment="ETCD_CERT_FILE=/etc/kubernetes/ssl/etcd/server-crt.pem"
-      Environment="ETCD_KEY_FILE=/etc/kubernetes/ssl/etcd/server-key.pem"
+      Environment="ETCD_CA_CERT_FILE=/etc/kubernetes/ssl/calico/etcd-ca"
+      Environment="ETCD_CERT_FILE=/etc/kubernetes/ssl/calico/etcd-cert"
+      Environment="ETCD_KEY_FILE=/etc/kubernetes/ssl/calico/etcd-key"
```
stone-z, line 305: I'm confused about the variable naming. For my knowledge, why is the `calico` cert referred to as the `ETCD_X_FILE` / why isn't [the `etcd` bundle](https://github.com/giantswarm/certs/blob/05a37d211a36f3e1bdef367279311f06f6753bf8/pkg/certs/files.go#L44) used?

tfussell: This is a weird one. On KVM, we use the etcd backend for Calico. This means that each node needs a client certificate to get state directly from etcd. As far as I can tell, this TLS configuration is passed into kubelet with these environment variables. Then the Calico CNI has access to them from the kubelet. I might have this wrong, but I believe Calico fails without them (I tried removing them in the past). This simply aligns the path with our other libraries. The better approach would be to use the kubernetes datastore so we don't have an etcd client cert on every node--don't think we have a story for this yet.

tfussell: To more directly answer your question, I think that file only exists on the master, but we need it on all of the workers as well so we need to use this one.

MarcelMue (approved):

stone-z (approved):

tfussell:
Tested upgrades on KVM and AWS. Going with this.

PR #818, opened 2020-10-28
Add dockerhub authentication for kubelet (v9)

wtty-fool:
Towards https://github.com/giantswarm/giantswarm/issues/13981

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
+### Added
+
+- Add dockerhub authentication for `kubelet`.
+
 ## [9.0.0] - 2020-10-27
 
 ### Changed
--- a/pkg/template/master_template.go
+++ b/pkg/template/master_template.go
@@ -293,6 +293,7 @@ systemd:
       Description=k8s-kubelet
       StartLimitIntervalSec=0
       [Service]
+	  User=root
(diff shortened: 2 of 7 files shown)


pkg/template/master_template.go
```diff
@@ -293,6 +293,7 @@ systemd:
       Description=k8s-kubelet
       StartLimitIntervalSec=0
       [Service]
+	  User=root
```
corest, line 296: we don't need an explicit user. It is running as root already.

wtty-fool: Ok, let's test it without explicit user. I based this on the following thread https://github.com/kubernetes/kubernetes/issues/45487#issuecomment-464516064

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
+### Added
+
+- Add dockerhub authentication for `kubelet`.
```
kopiczko, line 12: This is a breaking change. We need to be clear about that. Otherwise dependabot will make a mess.

```suggestion
- Add dockerhub authentication for `kubelet`.

### Changed

- Add new required `DockerhubToken` parameter for templating.
```

kopiczko: Ah but we also want to release it for previus majors. This sucks. We need to break the versioning then...

kopiczko (approved):

PR #824, opened 2020-10-30
Set root user for k8s-kubelet explicitly (v7)

corest:
## Checklist

### DIFF
--- a/pkg/template/master_template.go
+++ b/pkg/template/master_template.go
@@ -293,6 +293,7 @@ systemd:
       Description=k8s-kubelet
       StartLimitIntervalSec=0
       [Service]
+      User=root
       TimeoutStartSec=300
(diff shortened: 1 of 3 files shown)


wtty-fool (approved):

pkg/template/master_template.go
```diff
@@ -293,6 +293,7 @@ systemd:
       Description=k8s-kubelet
       StartLimitIntervalSec=0
       [Service]
+      User=root
```
kopiczko, line 296: So if you don't specify it what is the user running the service?

corest: It is still the root. But without `HOME` env variable. Here it is explained https://github.com/kubernetes/kubernetes/issues/45487#issuecomment-312042754

PR #833, opened 2020-11-05
Set configurable node labels in k8s-label-node unit

tuommaki:
When upgrading a node so that the node identity doesn't change, old
labels need to be overwritten in order to become effective. Therefore
those must be applied in `k8s-label-node` unit.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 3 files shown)


tuommaki:
Tested this on Azure where we improved master instance upgrade. Previously it was updated, drained (and deleted) and reimaged, but now we only need to update it (Azure reboots the node with new ignition template) and reimage if necessary (it's not). Due to dropping intermediate draining with node deletion, the node object stays the same in k8s api and hence we need to overwrite existing labels with updated ones to recognize e.g. operator version change.

fiunchinho (approved):
LGTM

PR #848, opened 2020-12-01
Change kubelet runtime cgroup flag

yulianedyalkova:
Towards https://github.com/giantswarm/giantswarm/issues/14290

Use the systemd cgroup created for the docker service. Another options is to drop the flag but then dockerd will end up in the root cgroup for `hugetlb`, `net_cls`, `net_prio`, `freezer`, `cpuset` and `perf_event`

## Checklist


whites11 (approved):
LGTMIWFY when tested on all providers

tfussell:
Does this fix the `unknown container kubereserved.slice` warning in the kubelet logs? I didn't realize that kubelet actually changes the cgroup of the running docker service. How does this compare to setting `Slice=/kubereserved.slice` in the `docker.service` unit so that it's already in the right slice?

yulianedyalkova:
It does fix the warning, yes. I didn't know it changes the cgroup either but the [flag description](https://github.com/kubernetes/kubernetes/blob/1cdcc8338fb1451cc34e89e9b5a7e473bbb0755b/pkg/kubelet/config/flags.go#L90) hints that it does (I also tested it).

We already have the [`Slice=/kubereserved.slice`](https://github.com/giantswarm/k8scloudconfig/blob/master/pkg/template/master_template.go#L138) thing and it ensures that the dockerd process is running in the `/kubereserved.slice/docker.service` cgroup. The reason I added the flag explicitly is that it was supposed to add it to all the cgroups. What I mean:

Once you start docker service, the process is in the following cgroups:
```

11:devices:/kubereserved.slice/docker.service
10:memory:/kubereserved.slice/docker.service
9:cpuset:/
8:pids:/kubereserved.slice/docker.service
7:hugetlb:/
6:cpu,cpuacct:/kubereserved.slice/docker.service
5:net_cls,net_prio:/
4:perf_event:/
3:freezer:/
2:blkio:/kubereserved.slice/docker.service
1:name=systemd:/kubereserved.slice/docker.service
0::/
```

When we have the old value of the flag, kubelet changes them to:
```
11:devices:/kubereserved.slice
10:memory:/kubereserved.slice
9:cpuset:/kubereserved.slice
8:pids:/kubereserved.slice
7:hugetlb:/kubereserved.slice
6:cpu,cpuacct:/kubereserved.slice
5:net_cls,net_prio:/kubereserved.slice
4:perf_event:/kubereserved.slice
3:freezer:/kubereserved.slice
2:blkio:/kubereserved.slice
1:name=systemd:/kubereserved.slice
0::/
```

When I change the flag to the new value, kubelet doesn't do anything (I guess it does some smart stuff and sees that the process is already in the right cgroup so it's the same as just skipping the flag). Which means that we run in the root cgroup for hugetlb, net_cls, net_prio, freezer, cpuset and perf_event. **Does anyone think that's an issue?**

MarcelMue:
> Which means that we run in the root cgroup for hugetlb, net_cls, net_prio, freezer, cpuset and perf_event. Does anyone think that's an issue?

I don't see an issue but I am also not very knowledgeable on the topic.

MarcelMue (approved):

tfussell:
As far as I understand, the pods run in a separate hierarchy so these only apply to the dockerd process itself (and child processes). From reading the descriptions of those cgroups, they wouldn't seem to be useful for us. I think this is fine.

tfussell:
Sorry, to be clear, I think it's fine if those controllers run in the root cgroup, however, if this PR puts all of them into the correct `/kubereserved.slice/docker.service`, let's use this flag for consistency.

tfussell (approved):

kopiczko:
Maybe adding a comment above the flag that if the value is changed then we also need to change `Slice=` in docker service + maybe even link to [this comment](https://github.com/giantswarm/k8scloudconfig/pull/848#issuecomment-736671371) why that is.

yulianedyalkova:
Passing `--runtime-cgroups=/kubereserved.slice/docker.service` and skipping the flag both result in the same cgroup setup for dockerd:
```
11:devices:/kubereserved.slice/docker.service
10:memory:/kubereserved.slice/docker.service
9:cpuset:/
8:pids:/kubereserved.slice/docker.service
7:hugetlb:/
6:cpu,cpuacct:/kubereserved.slice/docker.service
5:net_cls,net_prio:/
4:perf_event:/
3:freezer:/
2:blkio:/kubereserved.slice/docker.service
1:name=systemd:/kubereserved.slice/docker.service
0::/
```

So if we're all ok with the fact that docker will run in the root cgroup for some systems, I'll just drop the flag. In that way we'll define the cgroup in one place only ([service file](https://github.com/giantswarm/k8scloudconfig/blob/master/pkg/template/master_template.go#L138)) and skip the coupling between the flag and the service file. This will also limit the possibility for unpleasant surprises like [this one](https://github.com/giantswarm/giantswarm/issues/14290).

Thanks everyone for the input!

PR #855, opened 2020-12-09
Update k8s-extract for k8s 1.19+

tfussell:
## Checklist


tfussell:
This has been tested in https://github.com/giantswarm/kvm-operator/pull/1105. The cluster still doesn't start completely due to other issues, but k8s-extract works with 1.19 with this PR.

files/conf/k8s-extract
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -euo pipefail
+
+container=$1
+
+docker cp "$container":/kubelet /opt/bin/kubelet
```
tfussell, line 6: Changed from `docker cp "$container":/usr/local/bin/kubelet /opt/bin/kubelet` corresponding to https://github.com/giantswarm/hyperkube/blob/master/Dockerfile#L14

files/conf/k8s-extract-binaries
```diff
@@ -1,7 +0,0 @@
-#!/bin/bash
```
tfussell, line 1: Renamed to `k8s-extract` above.

pkg/template/cloudconfig.go
```diff
@@ -67,20 +66,6 @@ func NewCloudConfig(config CloudConfigConfig) (*CloudConfig, error) {
 		)
 	}
 
-	{
-		kubernetesVersion, err := semver.NewVersion(config.Params.Versions.Kubernetes)
```
tfussell, line 71: No longer need the 1.17 version check.

pkg/template/master_template.go
```diff
@@ -279,7 +279,7 @@ systemd:
       ExecStartPre=/usr/bin/mkdir -p /opt/bin/
       ExecStartPre=/usr/bin/docker pull $IMAGE
       ExecStartPre=-/usr/bin/docker rm $CONTAINER_NAME
-      ExecStartPre=-/usr/bin/docker create --name $CONTAINER_NAME $IMAGE
+      ExecStartPre=-/usr/bin/docker create --name $CONTAINER_NAME $IMAGE /kubectl
```
tfussell, line 282: `hyperkube` doesn't define a command so we have to specify one to be able to create the container. We're not running the container so it doesn't matter otherwise.

pkg/template/release.go
```diff
@@ -16,7 +16,7 @@ func BuildImages(registryDomain string, versions Versions) Images {
 		CalicoKubeControllers:        buildImage(registryDomain, "giantswarm/kube-controllers", versions.Calico, ""),
 		CalicoNode:                   buildImage(registryDomain, "giantswarm/node", versions.Calico, ""),
 		Etcd:                         buildImage(registryDomain, "giantswarm/etcd", versions.Etcd, ""),
-		Hyperkube:                    buildImage(registryDomain, "giantswarm/hyperkube", versions.Kubernetes, ""),
+		Hyperkube:                    buildImage(registryDomain, "giantswarm/hyperkube", strings.TrimPrefix(versions.Kubernetes, "v"), ""),
```
tfussell, line 19: The new images we're building don't have a "v" prefix: https://quay.io/repository/giantswarm/hyperkube?tab=tags

MarcelMue: I know a little late, but can we add the `v` on the images? :P

tfussell: We're building the image with architect-orb which doesn't seem to allow adding a prefix. We'd have to do it for all of our images I think.

MarcelMue: Okay . fine for now then :)

MarcelMue (approved):

PR #863, opened 2021-01-05
Retrieve Calico CRDs externally

tfussell:
Closes https://github.com/giantswarm/giantswarm/issues/16785

On KVM we have been seeing new workload cluster masters failing to start because ignition takes too long (>15 minutes). I determined that the time is proportional to the ignition size and the Calico V1 CRDs doubled the ignition to over 200KB. I first tried uploading the CRDs to S3 and using an `s3://` URL in ignition, but I found that this was blocked in China. My next idea was to pull the CRDs from GitHub and use an HTTPS URL in ignition. I tested this on KVM and AWS China and it worked well for both. This isn't the most elegant approach, but as we are retiring our non-KVM provider operators, I think it's adequate for now. We will probably install Calico as an App at some point to improve this.

Tested in https://github.com/giantswarm/kvm-operator/pull/1116 and https://github.com/giantswarm/aws-operator/pull/2923

Unblocks https://github.com/giantswarm/roadmap/issues/247 https://github.com/giantswarm/roadmap/issues/176 https://github.com/giantswarm/giantswarm/issues/14791 https://github.com/giantswarm/giantswarm/issues/16594 https://github.com/giantswarm/giantswarm/issues/15266 https://github.com/giantswarm/giantswarm/issues/15101

## Checklist


tfussell:
I will ensure AWS China access in the v9 cherry-pick https://github.com/giantswarm/k8scloudconfig/pull/864

pkg/template/master_template.go
```diff
@@ -427,7 +427,8 @@ storage:
       filesystem: root
       mode: 0644
       contents:
-        source: "data:text/plain;charset=utf-8;base64,{{  index .Files "k8s-resource/calico-crds.yaml" }}"
+        source: "s3://gs-tc-crds/calico/v3.16.5/crds.yaml"
```
kopiczko, line 430: Is this public object? How are we going to upload those?

tfussell: It's a public object. I created the bucket. I don't expect this to change often, but if we need to change it, we would upload a new one with a different version and update this line here to match. It's clunky, but shouldn't need to happen more than a few times a year.

kopiczko: Are those unchanged upstream CRDs? Do we have instructions how to download them?

pkg/template/master_template.go
```diff
@@ -427,7 +427,8 @@ storage:
       filesystem: root
       mode: 0644
       contents:
-        source: "data:text/plain;charset=utf-8;base64,{{  index .Files "k8s-resource/calico-crds.yaml" }}"
+        source: "https://raw.githubusercontent.com/giantswarm/k8scloudconfig/master/files/k8s-resource/calico-crds.yaml"
```
tfussell, line 430: Not sure how to version this file (or if it's really necessary). I could use a commit hash to point to this commit, or release one version and then a second version which has the first version in the URL. Maybe it's fine like this.

tfussell:
@giantswarm/team-rocket Please review.

pkg/template/master_template.go
MarcelMue: Please at least use the commit. This is bound to blow up if we ever restructure the k8scloudconfig repo. You can maybe use the SHA first and use the version structure later.

tfussell: Done. I went with the CRDs from the latest release, v10.2.0.

pkg/template/master_template.go
```diff
@@ -427,7 +427,8 @@ storage:
       filesystem: root
       mode: 0644
       contents:
-        source: "data:text/plain;charset=utf-8;base64,{{  index .Files "k8s-resource/calico-crds.yaml" }}"
+        source: "https://raw.githubusercontent.com/giantswarm/k8scloudconfig/v10.2.0/files/k8s-resource/calico-crds.yaml"
```
MarcelMue, line 430: Additional thought: Is this guarded for KVM? What happens e.g. in aws china?

tfussell: I tested cluster creation in AWS China with the aws-operator PR linked above. The CRDs were able to be pulled from GitHub in my test cluster. I know that the rules for what is blocked there can change over time, but we don't plan to have any more aws-operator releases and it does currently work. Do you think this is safe to rely on @giantswarm/team-firecracker-engineers?

tfussell: Our helm charts are hosted on GitHub Pages which is a GitHub domain and I don't think we've had problems with that so far.

MarcelMue (approved):
Works for me.

yulianedyalkova (approved):

pkg/template/master_template.go
yulianedyalkova: The hardcoded value worries me a bit, I have a bad feeling that we are going to change the crds in the future and then spend a lot of time trying to figure out why they're not being applied. But not sure how we can template this though so it always gets translated to the current version.

tfussell: True. I do feel like we need to take some time to improve this setup since installing CRDs and Calico feels quite outside the scope of ignition to me. We were able to install it an app during the hive sprint, so maybe we can try that soon. I'll open an issue so we don't forget.

rossf7: > True. I do feel like we need to take some time to improve this setup since installing CRDs and Calico feels quite outside the scope of ignition to me. We were able to install it an app during the hive sprint, so maybe we can try that soon. 

I can add some of the history here. When we moved the apps manifests out of cloudconfig to helm charts we stopped at CoreDNS and didn't remove Calico because at the time we couldn't get it to bootstrap in a new cluster. 

But this was a long time ago and many things will have changed. It's good news that Calico was installable as an app during the hive sprint and yes a helm chart is probably easier to manage.

calvix: I think we will have an issues in installations like buffalo where we need to allow each  domain we access

PR #867, opened 2021-01-15
Set stream timeout for kubelets

stone-z:
Towards https://github.com/giantswarm/giantswarm/issues/3418

## Checklist


MarcelMue (approved):
Works for me. I don't think there is any usecase to keep it longer than an hour

stone-z:
Agreed. The only concrete use case I've seen for streaming connections to the kubelet are for port-forwards, and I'm not sure what the use case is for port-forwarding which has no usage for 1+ hours

PR #869, opened 2021-01-18
Set kubelet runtime request timeouts

stone-z:
Towards https://github.com/giantswarm/giantswarm/issues/3398

Explicitly sets the kubelet request timeout to 2 minutes (this is the default)

## Checklist


stone-z:
This is actually a different flag than the one in the issue.

PR #879, opened 2021-02-15
Enable AWS CSI migration

njuettner:
Starting to phase out in-tree EBS volumes and start with EBS CSI 🚀

## Checklist


njuettner:
This PR adds to additional flags

- CSIMigrationAWS
-  CSIMigrationAWSComplete

CSIMigrationAWS will be activated and later once we migrated everything to CSI we enable the second one which removes the transition once everything is completed

files/manifests/k8s-controller-manager.yaml
```diff
@@ -32,7 +32,7 @@ spec:
     - --terminated-pod-gc-threshold=10
     - --use-service-account-credentials=true
     - --kubeconfig=/etc/kubernetes/kubeconfig/controller-manager.yaml
-    - --feature-gates=TTLAfterFinished=true
+    - --feature-gates=TTLAfterFinished=true,CSIMigrationAWS={{ .EnableCSIMigrationAWS }},CSIMigrationAWSComplete={{ .EnableCSIMigrationAWSComplete }}
```
anvddriesch, line 35: So I can understand it better, where do we set the value of `EnableCSIMigrationAWSComplete`?

njuettner: Sure. It will be set in the aws-operator, there's a second PR https://github.com/giantswarm/aws-operator/pull/2979/files#diff-86e71e19ce47e86816692e26dcc5f757d6361c59da4c695766d11d6f3309aa91R351

njuettner: And the `EnableCSIMigrationAWSComplete` is set by default to false once the `Params`struct in `k8scloudconfig` has been initiated by `aws-operator`

anvddriesch (approved):
LGTM if tested. Don't forget the changelog. :)

files/manifests/k8s-controller-manager.yaml
anvddriesch: Thank you!

PR #887, opened 2021-03-19
Enable `anonymous-auth` in API server 

whites11:
This is needed by kubeadm

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
+### Changed
+
+- Enable `anonymous-auth` in API server to comply with CAPI (needed by `kubeadm`).
+
 ## [10.1.0] - 2021-02-23
 
 ### Changed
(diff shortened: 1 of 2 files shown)


puja108 (approved):
I'm fine with enabling this as upstream and CIS benchmark also are ok with this as long as RBAC is on and tight and it does enable some features. We just need to keep in mind that our CVE exposure might be a bit higher with this cc @giantswarm/sig-security

MarcelMue:
Why did we merge this before checking CIS?

Pretty sure we removed it BECAUSE it fails CIS:
https://github.com/giantswarm/k8scloudconfig/pull/224

whites11:
I though we made the CIS tests already @puja108 ?

puja108:
from most recent CIS:

> If you are using RBAC authorization, it is generally considered reasonable to allow anonymous access to the API Server for health checks and discovery purposes, and hence this recommendation is not scored. However, you should consider whether anonymous discovery is an acceptable risk for your purposes.

It is not scored as long as you have RBAC in place

puja108:
For reference, see page 51/52 [CIS_Kubernetes_Benchmark_v1.6.1.pdf](https://github.com/giantswarm/k8scloudconfig/files/6170147/CIS_Kubernetes_Benchmark_v1.6.1.pdf)

PR #896, opened 2021-04-22
Add --bind-address-hard-fail flag to kube-proxy

ericgraf:
## Checklist

- [ x ] Update changelog in CHANGELOG.md.

fixes: https://github.com/giantswarm/giantswarm/issues/10953

This flag was introduced in k8s 1.19; therefore k8s 1.19 is required.


ericgraf:
Might be worth making this change configurable . WDYT ?

stone-z:
> Might be worth making this change configurable . WDYT ?

Since `kube-proxy` is our responsibility I think it's ok for us to decide to get paged for it. It is possible that whatever has already bound the port is also important, but as we've seen, when it gets into the current soft fail state it eventually causes weird problems we have to debug and resolve manually. _Hopefully_ we don't have too many more releases based on k8scloudconfig anyway.

files/config/kube-proxy.yaml
```diff
@@ -11,3 +11,4 @@ conntrack:
   maxPerCore: {{ .Cluster.Kubernetes.NetworkSetup.KubeProxy.ConntrackMaxPerCore }}
 {{- end }}
 metricsBindAddress: 0.0.0.0:10249
+bind-address-hard-fail: true
```
stone-z, line 14: This is the flag name but the config file has a different naming convention https://kubernetes.io/docs/reference/config-api/kube-proxy-config.v1alpha1/

```suggestion
bindAddressHardFail: true
```

ericgraf:
After trying for many days the sonobuoy jobs finally passed on `ghost`
```
PLUGIN     STATUS   RESULT   COUNT
   e2e   complete   passed       1
  systemd-logs   complete   passed       4

Sonobuoy has completed. Use `sonobuoy retrieve` to get results.

```
Here is the content of kube-proxy config map which includes the new `bindAddressHardFail` flag.
```
apiVersion: v1
data:
  kube-proxy.yaml: |
    apiVersion: kubeproxy.config.k8s.io/v1alpha1
    clientConnection:
    kind: KubeProxyConfiguration
    mode: iptables
    bindAddressHardFail: true
kind: ConfigMap
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"v1","data":{"kube-proxy.yaml":"apiVersion: kubeproxy.config.k8s.io/v1alpha1\nclientConnection:\n  kubeconfig: /etc/kubernetes/config/proxy-kubeconfig.yaml\nkind: KubeProxyConfiguration\nmode: iptables\nmetricsBindAddress: 0.0.0.0:10249\nbindAddressHardFail: true\n"},"kind":"ConfigMap","metadata":{"annotations":{},"creationTimestamp":null,"name":"kube-proxy","namespace":"kube-system"}}
  creationTimestamp: "2021-04-29T20:08:31Z"
  managedFields:
  - apiVersion: v1
    fieldsType: FieldsV1
    fieldsV1:
      f:data:
        .: {}
        f:kube-proxy.yaml: {}
      f:metadata:
        f:annotations:
          .: {}
          f:kubectl.kubernetes.io/last-applied-configuration: {}
    manager: kubectl-client-side-apply
    operation: Update
    time: "2021-04-29T20:08:31Z"
  name: kube-proxy
  namespace: kube-system
  resourceVersion: "326"
  selfLink: /api/v1/namespaces/kube-system/configmaps/kube-proxy
  uid: 788dd44b-6f9b-4ae2-bdf4-2062cfca7444

```

MarcelMue (approved):
Let's goooo

stone-z (approved):

PR #909, opened 2021-05-17
Add proxies to the master and worker templates

tfussell:
## Checklist

### DIFF
--- a/pkg/template/master_template.go
+++ b/pkg/template/master_template.go
@@ -138,12 +138,15 @@ systemd:
           Slice=kubereserved.slice
           Environment="DOCKER_CGROUPS=--exec-opt native.cgroupdriver=cgroupfs --cgroup-parent=/kubereserved.slice --log-opt max-size=25m --log-opt max-file=2 --log-opt labels=io.kubernetes.container.hash,io.kubernetes.container.name,io.kubernetes.pod.name,io.kubernetes.pod.namespace,io.kubernetes.pod.uid"
           Environment="DOCKER_OPT_BIP=--bip={{.Cluster.Docker.Daemon.CIDR}}"
-					{{ if ne .Proxy.HTTP"" }}
-					Environment="HTTP_PROXY={{ .Proxy.HTTP }}"
-					{{ end }}
-					{{ if ne .Proxy.HTTPS"" }}
-					Environment="HTTPS_PROXY={{ .Proxy.HTTPS }}"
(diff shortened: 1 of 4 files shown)


pkg/template/master_template.go
```diff
@@ -138,6 +138,15 @@ systemd:
...
           Environment="DOCKER_CGROUPS=--exec-opt native.cgroupdriver=cgroupfs --cgroup-parent=/kubereserved.slice --log-opt max-size=25m --log-opt max-file=2 --log-opt labels=io.kubernetes.container.hash,io.kubernetes.container.name,io.kubernetes.pod.name,io.kubernetes.pod.namespace,io.kubernetes.pod.uid"
           Environment="DOCKER_OPT_BIP=--bip={{.Cluster.Docker.Daemon.CIDR}}"
+          {{- if .Proxy.HTTP }}
+          Environment="HTTP_PROXY={{ .Proxy.HTTP }}"
+          {{- end }}
+          {{- if .Proxy.HTTPS }}
+          Environment="HTTPS_PROXY={{ .Proxy.HTTPS }}"
+          {{- end }}
+          {{- if .Proxy.NoProxy }}
+          Environment="NO_PROXY={{ .Proxy.NoProxy }}"
```
glitchcrab, line 148: so we're passing the values as a single comma-delimited string? i assume this string will be created from the installation config by kvm-operator?

yulianedyalkova: > so we're passing the values as a single comma-delimited string?

Yes

> i assume this string will be created from the installation config by kvm-operator?

`kvm-operator` will pass the values from its environment ([link](https://github.com/giantswarm/kvm-operator/blob/be5d19eddf8cae39b11976c35b79185d6d645e6d/service/controller/cloudconfig/worker_template.go#L45-L47)). So flow is `installations` -> `kyverno` -> `kvm-operator` -> `k8scloudconfig`.

tfussell: I thought about making it a list and joining it as we do elsewhere, but it's already a comma-delimited string so it wouldn't gain us much. We'll pass it in via config in a nicer way in the future: https://github.com/giantswarm/giantswarm/issues/17252

tfussell:
I can't approve as I opened the PR for Yulia, but we've tested it and it looks good to me.

glitchcrab (approved):
Lgtm

PR #914, opened 2021-05-20
Add KVM worker host volumes

mazzy89:
Signed-off-by: Salvatore Mazzarino <dev@mazzarino.cz>

## Checklist


pkg/template/types.go
```diff
@@ -45,8 +45,9 @@ type Params struct {
 	// Kubernetes components allow the passing of extra `docker run` and
 	// `command` arguments to image commands. This allows, for example,
 	// the addition of cloud provider extensions.
-	Kubernetes Kubernetes
-	Node       v1alpha1.ClusterNode
+	Kubernetes           Kubernetes
+	KVMWorkerHostVolumes []v1alpha1.KVMConfigSpecKVMNodeHostVolumes
```
tfussell, line 49: Even though `k8scloudconfig` is soon going to be KVM-only, I'd prefer to use a provider-independent type for this. This could be a slice of structs, but if we're only using `MountTag`, it could just be a slice of `string`s.

pkg/template/worker_template.go
```diff
@@ -45,6 +45,19 @@ systemd:
...
       {{range .Content}}{{.}}
       {{end}}{{end}}
+  {{ range .KVMWorkerHostVolumes }}
+  - name: data-{{ .MountTag }}.mount
+	enabled: true
+	contents: |
+      Description=Guest mount for {{ .MountTag }} host volume
+      [Mount]
+      What={{ .MountTag }}
+      Where=/data/{{ .MountTag }}
```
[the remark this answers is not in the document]
mazzy89: In this case `.HostPath` is not needed.

pkg/template/types.go
mazzy89: ok

mazzy89: fixed.

PR #926, opened 2021-07-13
Update Calico to 3.19

tfussell:
Towards https://github.com/giantswarm/giantswarm/issues/17954

Based on https://docs.projectcalico.org/manifests/calico.yaml

## Checklist

### DIFF
--- a/files/k8s-resource/calico-all.yaml
+++ b/files/k8s-resource/calico-all.yaml
@@ -2,6 +2,7 @@
 #  - Added resource limits to calico-node and calico-kube-controller.
 #  - Added resource limits to install-cni.
 #  - Added 'priorityClassName: system-node-critical' to calico daemonset.
+#  - Added `runAsUser: 0` to calico-kube-controllers deployment.
 
 # Source: calico/templates/calico-config.yaml
 # This ConfigMap is used to configure a self-hosted Calico installation.
--- a/files/k8s-resource/calico-all.yaml
+++ b/files/k8s-resource/calico-all.yaml
@@ -304,8 +304,8 @@ spec:
               value: "true"
             - name: FELIX_PROMETHEUSMETRICSENABLED
               value: "true"
-            - name: IP_AUTODETECTION_METHOD
-              value: "interface=eth.*"
+            - name: FELIX_LOGSEVERITYFILE
+              value: "Debug"
           securityContext:
             privileged: true
           resources:
--- a/files/k8s-resource/calico-all.yaml
+++ b/files/k8s-resource/calico-all.yaml
@@ -265,7 +265,7 @@ spec:
               value: "autodetect"
             # Enable IPIP
             - name: CALICO_IPV4POOL_IPIP
-              value: "Never"
+              value: "Always"
             # Set MTU for tunnel device used if ipip is enabled
             - name: FELIX_IPINIPMTU
               valueFrom:
--- a/files/k8s-resource/calico-all.yaml
+++ b/files/k8s-resource/calico-all.yaml
@@ -489,8 +489,7 @@ spec:
             - mountPath: /calico-secrets
               name: etcd-certs
           securityContext:
-            runAsUser: 1000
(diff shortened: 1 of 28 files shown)


pkg/key/key.go
```diff
@@ -3,7 +3,8 @@ package key
 // These consts are used by template.Params.Validate() to check that provided versions
 // fall within the known-good ranges.
 const (
-	CalicoVersionConstraint                       = ">= 3.15.0 < 3.16.0"
+	CalicoPolicyOnlyVersionConstraint             = ">= 3.15.0 < 3.16.0"
```
tfussell, line 6: I don't want to update and test `calico-policy-only.yaml` with Calico 3.19 so I'm splitting the version constraint depending on `CalicoPolicyOnly` parameter passed into the cloudconfig template.

files/k8s-resource/calico-all.yaml
```diff
@@ -218,13 +265,25 @@ spec:
               value: "autodetect"
             # Enable IPIP
             - name: CALICO_IPV4POOL_IPIP
-              value: "Always"
+              value: "Never"
```
tfussell, line 268: Discussed here https://gigantic.slack.com/archives/C01BYMF6RN0/p1629140646152200 and here https://gigantic.slack.com/archives/C3S482Z98/p1629144487000400

files/k8s-resource/calico-all.yaml
```diff
@@ -399,15 +489,26 @@ spec:
...
+          securityContext:
+            runAsUser: 0
           resources:
             requests:
               cpu: 250m
               memory: 100Mi
+          livenessProbe:
+            exec:
+              command:
+                - /usr/bin/check-status
```
tfussell, line 501: New in 3.19

tfussell:
I have tested cluster creation with this several times. I will test CNCF and upgrades today.

files/k8s-resource/calico-all.yaml
MarcelMue: Seems reasonable to me - only question:
Is there something we can set as a target to verify it works as intended? E.g. we upgrade some dev clusters and monitor how they are behaving more closely?

MarcelMue (approved):
Not blocking to try this out and give a proper release a spin :rocket:

files/k8s-resource/calico-all.yaml
```diff
@@ -399,15 +489,26 @@ spec:
             # Mount in the etcd TLS secrets.
             - mountPath: /calico-secrets
               name: etcd-certs
+          securityContext:
+            runAsUser: 0
```
pipo02mix, line 493: What does it change to need it now?

pipo02mix (approved):
LGTMIIWFY

files/k8s-resource/calico-all.yaml
tfussell: Yulia made this change. I assume it was necessary to make it run, but I don't know for sure. Interestingly, in 3.18 there was this change so it would run as non-root https://github.com/projectcalico/kube-controllers/pull/565 I can try to revert it.

files/k8s-resource/calico-all.yaml
tfussell: Like prometheus monitoring? I think `net-exporter` should already alert us if some pods can't reach other pods.

MarcelMue: Yes - just keeping a closer eye on the before & after.

tfussell:
Thanks for the reviews. I decided to take a different approach so I'm closing this in favor of https://github.com/giantswarm/k8scloudconfig/pull/941

files/k8s-resource/calico-all.yaml
yulianedyalkova: Changes I did were documented here: https://github.com/giantswarm/giantswarm/issues/17954#issuecomment-898328797

PR #928, opened 2021-07-30
Removed EnableCSIMigrationAWSComplete feature gate flag

ericgraf:
## Checklist

Related to: https://github.com/giantswarm/giantswarm/issues/18284

### DIFF
--- a/files/manifests/k8s-controller-manager.yaml
+++ b/files/manifests/k8s-controller-manager.yaml
@@ -32,7 +32,7 @@ spec:
     - --terminated-pod-gc-threshold=10
     - --use-service-account-credentials=true
     - --kubeconfig=/etc/kubernetes/kubeconfig/controller-manager.yaml
-    - --feature-gates=TTLAfterFinished=true,CSIMigrationAWS={{ .EnableCSIMigrationAWS }}
+    - --feature-gates=TTLAfterFinished=true,CSIMigrationAWS={{ .EnableCSIMigrationAWS }},InTreePluginAWSUnregister={{ .InTreePluginAWSUnregister }}
     - --profiling=false
     - --root-ca-file=/etc/kubernetes/ssl/apiserver-ca.pem
     - --service-account-private-key-file=/etc/kubernetes/ssl/service-account-key.pem
--- a/files/manifests/k8s-controller-manager.yaml
+++ b/files/manifests/k8s-controller-manager.yaml
@@ -32,7 +32,7 @@ spec:
     - --terminated-pod-gc-threshold=10
     - --use-service-account-credentials=true
     - --kubeconfig=/etc/kubernetes/kubeconfig/controller-manager.yaml
-    - --feature-gates=TTLAfterFinished=true,CSIMigrationAWS={{ .EnableCSIMigrationAWS }},CSIMigrationAWSComplete={{ .EnableCSIMigrationAWSComplete }}
+    - --feature-gates=TTLAfterFinished=true,CSIMigrationAWS={{ .EnableCSIMigrationAWS }}
     - --profiling=false
     - --root-ca-file=/etc/kubernetes/ssl/apiserver-ca.pem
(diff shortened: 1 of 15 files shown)


ericgraf:
FYI... Still working on testing change.

njuettner:
But from looking at the PR it seems Kubernetes decided to use `InTreePluginXXUnregister`? So we might only need to change that to `InTreePluginAWSUnregister`?

ericgraf:
Sounds like @njuettner you are right, only need to replace the flag with InTreePluginAWSUnregister. 

> `CSIMigrationAWSComplete` has been replaced with new flag called `InTreePluginAWSUnregister.` This new flag `InTreePluginAWSUnregister` will serve the same purpose as `CSIMigrationAWSComplete` except that it is decoupled from migration. So if you do not want to support in-tree at all. You can turn it on without migration. This flag is still in alpha and remain in alpha.

Ref: https://github.com/kubernetes/kops/issues/10777#issuecomment-776537706

ericgraf:
@njuettner I only see `EnableCSIMigrationAWS` but not `EnableCSIMigrationAWSCompleted` anywhere. Is the flag enabled anywhere else ? 
https://github.com/giantswarm/aws-operator/blob/master/service/internal/cloudconfig/tcnp.go#L184

files/manifests/k8s-controller-manager.yaml
```diff
@@ -32,7 +32,7 @@ spec:
     - --terminated-pod-gc-threshold=10
     - --use-service-account-credentials=true
     - --kubeconfig=/etc/kubernetes/kubeconfig/controller-manager.yaml
-    - --feature-gates=TTLAfterFinished=true,CSIMigrationAWS={{ .EnableCSIMigrationAWS }},CSIMigrationAWSComplete={{ .EnableCSIMigrationAWSComplete }}
+    - --feature-gates=TTLAfterFinished=true,CSIMigrationAWS={{ .EnableCSIMigrationAWS }},InTreePluginAWSUnregister={{ .InTreePluginAWSUnregister }}
```
MarcelMue, line 35: Do we actually still need this flag? I am trying to understand what is stopping us from eventually dropping these :D

njuettner: Customer might still use an old storage class which uses the in-tree plugin to create EBS volumes.

CSIMigrationAWS will forward EBS traffic to EBS CSI Driver in case customer requests it from old in-tree plugin.

When InTreePluginAWSUnregister enabled it will reject sending traffic to in-tree plugin. In case customer doesn't change storage-class it might be a problem.

I already applied with ebs-csi-driver a new storage class `gp3` which will be as the default storage class in release `v15.0.0`. So in case customer doesn't specify the storage-class we are safe, but there might be a custom storage classes from customer still using in-tree plugin.

We should give customers some time and at some point we can drop in-tree.

njuettner:
> @njuettner I only see `EnableCSIMigrationAWS` but not `EnableCSIMigrationAWSCompleted` anywhere. Is the flag enabled anywhere else ?

Yes it is, see https://github.com/giantswarm/k8scloudconfig/blob/master/pkg/template/types.go#L21, but not mentioned specifically in  https://github.com/giantswarm/aws-operator/blob/master/service/internal/cloudconfig/tcnp.go#L184 because by default it is set to `false`

files/manifests/k8s-controller-manager.yaml
MarcelMue: I see - that makes sense to me.

Would be cool if there was even a metric to show how many customers are using the InTreePlugin thing. But maybe something for the future :)

ericgraf:
@njuettner I'm not sure the best way to thoroughly test this change before it gets submitted. How do you recommend I test it?

njuettner (approved):
Good for me, for now we only need to set the flag. We can decide later when we want to stop supporting in-tree requests

PR #937, opened 2021-08-20
only add k8s controller flag if set to true

ericgraf:
## Checklist

---
InTreePluginAWSUnregister was only introduced in k8s 1.21. This change is to allow k8scloudconfig to still work on older versions by only setting the flag if set true,

### DIFF
--- a/files/manifests/k8s-controller-manager.yaml
+++ b/files/manifests/k8s-controller-manager.yaml
@@ -33,7 +33,7 @@ spec:
     - --use-service-account-credentials=true
     - --kubeconfig=/etc/kubernetes/kubeconfig/controller-manager.yaml
     {{ if .InTreePluginAWSUnregister }}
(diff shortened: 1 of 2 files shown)


files/manifests/k8s-controller-manager.yaml
```diff
@@ -32,7 +32,11 @@ spec:
     - --terminated-pod-gc-threshold=10
     - --use-service-account-credentials=true
     - --kubeconfig=/etc/kubernetes/kubeconfig/controller-manager.yaml
+    {{ if .InTreePluginAWSUnregister }}
     - --feature-gates=TTLAfterFinished=true,CSIMigrationAWS={{ .EnableCSIMigrationAWS }},InTreePluginAWSUnregister={{ .InTreePluginAWSUnregister }}
```
tfussell, line 36:
```suggestion
    - --feature-gates=TTLAfterFinished=true,CSIMigrationAWS={{ .EnableCSIMigrationAWS }},InTreePluginAWSUnregister=true
```

tfussell (approved):
Tested on KVM with k8s 1.20 and Calico 3.19.

PR #940, opened 2021-08-24
Renable ssh-rsa CASignatureAlgorithms in sshd-config

ericgraf:
## Checklist

### DIFF
--- a/files/conf/sshd_config
+++ b/files/conf/sshd_config
@@ -13,3 +13,4 @@ MaxAuthTries 5
 LoginGraceTime 60
 AllowTcpForwarding no
 AllowAgentForwarding no
+CASignatureAlgorithms ecdsa-sha2-nistp256,ecdsa-sha2-nistp384,ecdsa-sha2-nistp521,ssh-ed25519,rsa-sha2-512,rsa-sha2-256,ssh-rsa
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
+### Changed
+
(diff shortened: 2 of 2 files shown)


ericgraf:
Note* this only fixes WCs not MCs

files/conf/sshd_config
```diff
@@ -13,3 +13,4 @@ MaxAuthTries 5
 LoginGraceTime 60
 AllowTcpForwarding no
 AllowAgentForwarding no
+CASignatureAlgorithms ecdsa-sha2-nistp256,ecdsa-sha2-nistp384,ecdsa-sha2-nistp521,ssh-ed25519,rsa-sha2-512,rsa-sha2-256,ssh-rsa
```
stone-z, line 16: Assuming the others are just the defaults, so `ssh-rsa` is the only deviation?

stone-z (approved):

files/conf/sshd_config
ericgraf: Correct , it's a list of default I found based on openbsd, ubuntu and debian.
- https://manpages.ubuntu.com/manpages/focal/en/man5/sshd_config.5.html
- https://man.openbsd.org/sshd_config.5#CASignatureAlgorithms
- https://manpages.debian.org/experimental/openssh-server/sshd_config.5.en.html

Arch linux specifies `sk` keys ( `sk-ssh-ed25519` and `sk-ecdsa-sha2-nistp256` ) which are for FIDO2 security keys.  Do you think it's worth adding those? 
- https://man.archlinux.org/man/ssh_config.5.en#CASignatureAlgorithms

Theoretically we could narrow it down even more to the list of keys used in employees repo and ssh-rsa.

calvix (approved):
yeah

MarcelMue (approved):
Can we still have a tracking issue to figure out how to proceed once the flag is removed?

PR #950, opened 2021-09-13
Avoid running Calico datastore migration pods when Calico is in policy only mode

whites11:
## Checklist

### DIFF
--- a/files/conf/k8s-addons
+++ b/files/conf/k8s-addons
@@ -88,14 +88,15 @@ echo "kube-proxy successfully installed"
 # restart ds to apply config from configmap
 kubectl delete pods -l app.kubernetes.io/name=kube-proxy -n kube-system
 
+kubectl apply -f /srv/calico-crd-installer-rbac.yaml
+
+kubectl apply -f /srv/calico-crd-installer.yaml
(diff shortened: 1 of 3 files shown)


files/conf/k8s-addons
```diff
@@ -88,13 +88,15 @@ echo "kube-proxy successfully installed"
 # restart ds to apply config from configmap
 kubectl delete pods -l app.kubernetes.io/name=kube-proxy -n kube-system
 
+{{- if not .CalicoPolicyOnly }}
 # TODO: remove migrator once all clusters have been migrated to kubernetes datastore
 kubectl apply -f /srv/calico-crd-installer-rbac.yaml
 for job in calico-crd-installer calico-datastore-migrator-pre
```
tfussell, line 94: We need `calico-crd-installer` for policy-only and full Calico so we'll have to drop the `for` loop and split this in two parts.

whites11: good point

whites11: done

tfussell (approved):
LGTM.

whites11:
> LGTM.

Thanks. Testing on Azure, and then merging

whites11:
> > LGTM.
> 
> Thanks. Testing on Azure, and then merging

tests green

PR #954, opened 2021-09-28
fix-aws-cni-file

calvix:
this part was recently refactorized and aws was without calico-policy only, fixing the bug


njuettner (approved):
wow 😄, good catch

anvddriesch (approved):

PR #958, opened 2021-10-05
Bump calico

yulianedyalkova:
## Checklist

### DIFF
--- a/pkg/key/key.go
+++ b/pkg/key/key.go
@@ -5,7 +5,7 @@ package key
 const (
 	CalicoPolicyOnlyVersionConstraint             = ">= 3.15.0 < 3.16.0"
 	CalicoVersionConstraint                       = ">= 3.20.0 < 3.21.0"
-	EtcdVersionConstraint                         = ">= 3.4.0 < 3.5.0"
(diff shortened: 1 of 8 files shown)


MarcelMue (approved):

pkg/key/key.go
```diff
@@ -4,8 +4,8 @@ package key
 // fall within the known-good ranges.
 const (
 	CalicoPolicyOnlyVersionConstraint             = ">= 3.15.0 < 3.16.0"
-	CalicoVersionConstraint                       = ">= 3.19.0 < 3.20.0"
-	EtcdVersionConstraint                         = ">= 3.4.0 < 3.5.0"
+	CalicoVersionConstraint                       = ">= 3.20.0 < 3.21.0"
+	EtcdVersionConstraint                         = ">= 3.4.0 < 3.6.0"
```
kopiczko, line 8: What is that doing?

yulianedyalkova: When creating a cluster we [validate](https://github.com/giantswarm/k8scloudconfig/blob/e33f10cf37f867802be348106b0d2b7d1f9cd9db/pkg/template/validation.go) if we have the right component versions. We need to bump both calico and etcd in the `v15.1.0` release so these constraints need to get changed.

PR #976, opened 2022-01-14
Add new EC2 instance types

njuettner:
Add new EC2 instance types for AWC CNI max pod calculation.

Addresses: https://github.com/giantswarm/roadmap/issues/672

## Checklist


calvix (approved):

whites11:
any way to get this data automatically from an AWS endpoint? It would be nice to have this automated to avoid the need for new releases when we add an instance type

njuettner:
> any way to get this data automatically from an AWS endpoint? It would be nice to have this automated to avoid the need for new releases when we add an instance type

good point I would suggest creating a new issue into our backlog to automate this

njuettner:
Ok will open an issue and merge this for now

PR #1021, opened 2022-03-24
Cgroupsv1 noreboot

whites11:
Towards: https://github.com/giantswarm/roadmap/issues/944

Leverage flatcar feature instead of our systemd unit + script for enabling cgroups v1.
No change needed in provider operators, it is enough to bump to this version of k8scc

## Checklist


pkg/template/worker_template.go
```diff
@@ -329,28 +314,13 @@ storage:
...
   files:
     - path: /boot/coreos/first_boot
       filesystem: root
     {{ if .ForceCGroupsV1 }}
-    - filesystem: "OEM"
-      path: "/grub.cfg"
-      mode: 0644
-      append: true
-      contents:
-        source: "data:text/plain;base64,{{ index .Files "conf/grub.cfg" }}"
```
pothos, line 348: Just a note from the off: I recommend to keep this because it helps to avoid the bind mount over /proc/cmdline on subsequent boots.

whites11: Thanks for your comment @pothos 
Can you please explain how having the `grub.cfg` file in place avoids one bind mount? I'm not following

pothos: The flag file under `/etc/` sets up a bind mount over `/proc/cmdline` to avoid the reboot. When the right parameters are set in `grub.cfg`, even this bind mount is not needed anymore.
In our testing the bind mount worked well, but it certainly is no long-term solution.

whites11: ok thank you

PR #1029, opened 2022-04-04
Kube proxy vpa service

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/21494

This PR adds a new systemd unit to master nodes that runs after `k8s-addons` and waits until `verticalpodautoscaler` CR exists before it creates the VPA for kube-proxy.
Hacky solution maybe, but I tested it and it works just fine.

## Checklist

### DIFF
--- a/files/conf/kube-proxy-vpa
+++ b/files/conf/kube-proxy-vpa
@@ -1,0 +1,22 @@
+apiVersion: autoscaling.k8s.io/v1
+kind: VerticalPodAutoscaler
+metadata:
+  labels:
+    app.kubernetes.io/name: kube-proxy
+    giantswarm.io/service-type: system
+    k8s-app: kube-proxy
+    kubernetes.io/cluster-service: "true"
+  name: kube-proxy
+  namespace: kube-system
+spec:
+  resourcePolicy:
+    containerPolicies:
+    - containerName: kube-proxy
+      controlledValues: RequestsOnly
+      mode: Auto
+  targetRef:
+    apiVersion: apps/v1
+    kind: DaemonSet
+    name: kube-proxy
+  updatePolicy:
+    updateMode: Auto
--- a/pkg/template/master_template.go
+++ b/pkg/template/master_template.go
@@ -399,6 +399,19 @@ systemd:
       ExecStartPost=/usr/bin/systemctl restart k8s-kubelet.service
       [Install]
       WantedBy=multi-user.target
+  - name: ensure-kube-proxy-vpa.service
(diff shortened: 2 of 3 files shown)


pkg/template/master_template.go
```diff
@@ -399,6 +399,19 @@ systemd:
...
+    enabled: true
+    contents: |
+      [Unit]
+      Description=Enable VPA for kube-proxy
+      After=k8s-addons.service
+      [Service]
+      Type=oneshot
+      Restart=on-failure
+      ExecStartPre=/bin/bash -c "while ! /opt/bin/kubectl --kubeconfig=/etc/kubernetes/kubeconfig/addons.yaml get crd verticalpodautoscalers.autoscaling.k8s.io; do echo 'Waiting for VPA CRD to exists' && sleep 10; done"
+      ExecStart=/opt/bin/kubectl --kubeconfig=/etc/kubernetes/kubeconfig/addons.yaml apply -f /srv/kube-proxy-vpa.yaml
```
fiunchinho, line 412: This may be stupid from my side but, why wait until CRD exist instead of just applying the VPA CR until it succeeds? won't `kubectl` fails if trying to apply the VPA CR without the CRD being present?

The benefit I see is that if there is a problem with the API at that moment, it would retry to apply it, while on this approach it only tries to apply it once.

whites11: It is basically the same.

```
Restart=on-failure
```

makes it so that if the `apply` fails it will retry a few times.
The real downside (in my opinion) of your approach is that the unit might run forever, while with mine it tries a few times and eventually gives up. Not a big deal either way. Happy to hear other opinions about this.

files/conf/kube-proxy-vpa
```diff
@@ -0,0 +1,22 @@
...
+    giantswarm.io/service-type: system
+    k8s-app: kube-proxy
+    kubernetes.io/cluster-service: "true"
+  name: kube-proxy
+  namespace: kube-system
+spec:
+  resourcePolicy:
+    containerPolicies:
+    - containerName: kube-proxy
+      controlledValues: RequestsOnly
```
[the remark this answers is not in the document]
whites11: don't know the internals, but I guess VPA monitors the resource usage (querying k8s for metrics) and adjustes the request over time to be around the needed value.

fiunchinho (approved):

whites11:
@calvix you ok with the PR?

calvix:
sure, if it works I am fine :)

PR #1032, opened 2022-04-19
Add extra sa key args for irsa

njuettner:
Issue: https://github.com/giantswarm/giantswarm/issues/21736

TL;DR 

Key files for new IRSA service account which signs JWT need to be after older service account key files.

## Checklist


calvix (approved):
not the best way but lets make this a quick fix

whites11 (approved):

PR #1035, opened 2022-05-11
For HA clusters, use the Load Balancer endpoint for etcd rather than localhost

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/21946#issuecomment-1123557966

Performance wise, the latency introduced by ELB and going to the LAN rather than using localhost is much less than using 1/3 of etcd power. This is a bottleneck for etcd performance on busy clusters

## Checklist


files/manifests/k8s-api-server.yaml
```diff
@@ -42,7 +42,11 @@ spec:
     - --enable-admission-plugins=NamespaceLifecycle,LimitRanger,ServiceAccount,ResourceQuota,DefaultStorageClass,PersistentVolumeClaimResize,PodSecurityPolicy,Priority,DefaultTolerationSeconds,MutatingAdmissionWebhook,ValidatingAdmissionWebhook
     - --cloud-provider={{.Cluster.Kubernetes.CloudProvider}}
     - --service-cluster-ip-range={{.Cluster.Kubernetes.API.ClusterIPRange}}
-    - --etcd-servers=https://127.0.0.1:2379
+    {{- if .Etcd.HighAvailability -}}
+    - --etcd-servers=https://{{ .Cluster.Etcd.Domain }}:2379
+    {{- else -}}
+    - --etcd-servers=https://{{ .Cluster.Etcd.Domain }}:2379
+    {{- end -}}
```
AverageMarcus, line 49: These are the same. 

(Is there any reason _not_ to do the same on non-HA?)

AverageMarcus (changes requested):

files/manifests/k8s-api-server.yaml
whites11: > These are the same.

-_-

> (Is there any reason _not_ to do the same on non-HA?)

yes. On azure we can't do this. Azure LBs are network load balancers. Traffic leaving the node and coming back to the same node will have src IP == dst IP and the kernel drops the packet ("martian source" error). So we simply can't do the same there

AverageMarcus: Damn that sucks!

AverageMarcus (approved):

PR #1057, opened 2022-07-26
Change preferred address type

whites11:
On AWS, master nodes have 2 ENIs (network interfaces), one used just for etcd, one used for the rest.
The kubelet binds to the first ethernet card, and does not listen to the other.
When prometheus tries to scrape stuff going through the API server's proxy, it selects the first private IP address as listed in the `Node` resource's status.
There is no guarantee over the order of IP addresses in the node status, and thus scraping can fail or succeed due to randomness.

This PR changes the way the kubelet exposes itself through the proxy, making it prefer the hostname over the private IP address  and hopefully fixing scraping troubles.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


whites11:
Pinging @AverageMarcus as I know he might be around at this time of day 😅

AverageMarcus (approved):
You know me so well @whites11 😆

AverageMarcus (approved):

PR #1070, opened 2022-08-30
Hostname check

whites11:
Sometimes, azure VMs take some time to have their hostname properly set.
This PR adds a sanity check that ensures the hostname is correct before starting the kubelet.
Without the check, node can't bootstrap correctly

## Checklist


mnitchev (approved):

zewolfe (approved):
LGTMIIWFY

PR #1091, opened 2022-11-18
Add registry mirror

njuettner:
Fixes https://github.com/giantswarm/giantswarm/issues/24615

We currently missing to set the registry mirror for workload clusters.

The registry mirrors is an array which comes from our config repo, https://github.com/giantswarm/config/blob/main/default/config.yaml#L7.

We should pick the first item to add this into our containerd configuration.
I double checked it on all installations, our China installations won't include this (we overwrite an empty array in those cases).

## Checklist


whites11 (approved):
<3

PR #1096, opened 2023-01-11
Fix apiserver requests

whites11:
The previous strategy we used to allocate memory request for API server static pod was too aggressive.
On azure, that left not enough room for pods to be scheduled in master nodes and lead to broken clusters.

This PR changes how we calculate requests and limits for API server pod and clarifies the reasoning behind the choice.

Tradeoff: fairness limits will be lower on the same hardware

## Checklist

### DIFF
--- a/files/conf/setup-apiserver-environment
+++ b/files/conf/setup-apiserver-environment
@@ -22,5 +22,7 @@ echo "MAX_REQUESTS_INFLIGHT=$((dedicated / 1000 * 200 * 2 / 3))" >$env_file
 echo "MAX_MUTATING_REQUESTS_INFLIGHT=$((dedicated / 1000 * 200 * 1 / 3 ))" >>$env_file
 echo "CPU_LIMIT=${dedicated}m" >>$env_file
 echo "MEMORY_LIMIT=$((memory * 3 / 4))Gi" >>$env_file
-echo "CPU_REQUEST=$((cpus / 2))m" >>$env_file
+
+# We request 1/3 of the available CPUs and 1/2 of the system memory just for api server.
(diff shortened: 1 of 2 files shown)


files/conf/setup-apiserver-environment
```diff
@@ -7,11 +7,22 @@ memory="$(awk '/MemTotal/ { printf "%d \n", $2/1024/1024 }' /proc/meminfo)"
...
+# for every core we have dedicated to api server we allow 200 requests for fairness.
+# 1/3 of such number is for mutating requests, 2/3 are for all requests.
+echo "MAX_REQUESTS_INFLIGHT=$((dedicated / 1000 * 200 * 2 / 3))" >$env_file
+echo "MAX_MUTATING_REQUESTS_INFLIGHT=$((dedicated / 1000 * 200 * 1 / 3 ))" >>$env_file
+echo "CPU_LIMIT=${dedicated}m" >>$env_file
 echo "MEMORY_LIMIT=$((memory * 3 / 4))Gi" >>$env_file
-echo "CPU_REQUEST=$((cpus / 2))" >>$env_file
+
+# We request 1/3 of the available CPUs and 1/2 of the system memory just for api server.
+echo "CPU_REQUEST=$((cpus / 3))m" >>$env_file
```
njuettner, line 27: Is the "m" correct or is this a typo because it was not used before?

whites11: we switched from full cores (no suffix) to millicores.

2 became 2000m

so the m is needed

njuettner: Cool that make sense 👍🏻

njuettner (approved):

PR #1098, opened 2023-01-11
draft pull request to check some branch behaviour

Strigix:
test with seccomp wildcard in restricted whether runtime/default was too strict to deploy clusters.

## Checklist

### DIFF
--- a/files/k8s-resource/psp_policies.yaml
+++ b/files/k8s-resource/psp_policies.yaml
@@ -31,6 +31,8 @@ apiVersion: policy/v1beta1
 kind: PodSecurityPolicy
 metadata:
   name: restricted
+  annotations:
+    seccomp.security.alpha.kubernetes.io/allowedProfileNames: '*'
 spec:
   privileged: false
   allowPrivilegeEscalation: false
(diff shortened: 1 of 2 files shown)


files/k8s-resource/psp_policies.yaml
```diff
@@ -31,6 +31,8 @@ apiVersion: policy/v1beta1
 kind: PodSecurityPolicy
 metadata:
   name: restricted
+  annotations:
+    seccomp.security.alpha.kubernetes.io/allowedProfileNames: 'runtime/default'
```
stone-z, line 35: For https://github.com/giantswarm/giantswarm/issues/22947 this probably needs to be a wildcard:
```
  annotations:
    seccomp.security.alpha.kubernetes.io/allowedProfileNames: '*'
```
Because we want to _allow_ `runtime/default` but not require it. I don't know of anywhere where Giant Swarm sets a profile other than runtime/default, but customers in theory could set one. For backward-compatibility I think we can't tighten this further (and PSPs are gone soon anyway). The issue was that our kyverno policies suggest setting runtime/default, but no profile is currently allowed by this PSP.

If you want to go a level further, you could also [set the default profile](https://v1-24.docs.kubernetes.io/docs/concepts/security/pod-security-policy/#seccomp) if a Pod doesn't set one. (If you do that, let's also make sure it ends up in the provider release notes)

stone-z (approved):
Please add a changelog entry but LGTM once tested

PR #1114, opened 2023-03-06
Enable `CronJobTimeZone` feature gate through feature flag.

whites11:
The `CronJobTimeZone` feature flag is only available in k8s 1.24+
This PR makes the feature flag configurable so we can bump k8scc on 1.23 clusters

## Checklist

### DIFF
--- a/pkg/template/types.go
+++ b/pkg/template/types.go
@@ -19,7 +19,7 @@ type Params struct {
 	AWSCNISubnetPrefixMode bool
 	// EnableCSIMigrationAWS flag. When set to true will use in-tree EBS volumes will be migrated to CSI.
 	EnableCSIMigrationAWS bool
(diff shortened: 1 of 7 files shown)


pkg/template/types.go
```diff
@@ -19,6 +19,8 @@ type Params struct {
 	AWSCNISubnetPrefixMode bool
 	// EnableCSIMigrationAWS flag. When set to true will use in-tree EBS volumes will be migrated to CSI.
 	EnableCSIMigrationAWS bool
+	// EnableCronJobTimeZone flag. When set to true the `EnableCronJobTimeZone` will be enabled.
```
fiunchinho, line 22: I guess?

```suggestion
	// EnableCronJobTimeZone flag. When set to true the `CronJobTimeZone` will be enabled.
```

whites11: yes thanks

fiunchinho (approved):

PR #1120, opened 2023-05-02
Disable ETCD compaction request from apiserver

njuettner:
Issue: https://github.com/giantswarm/giantswarm/issues/26781

## Checklist


whites11 (approved):
ok but this should be tested long term with big clusters before merging it blindly

njuettner:
Tested with a big cluster all good 🎉, merging it now.

PR #1125, opened 2023-05-23
Consider new control-plane taint

mogottsch:
This PR adds the new control-plane taint to occurrences of the old one.

## Checklist


mogottsch:
Not tested.

pkg/template/master_template.go
```diff
@@ -350,7 +350,7 @@ systemd:
         {{ end -}}
         --pod-infra-container-image={{ .Images.Pause }} \
         --register-node=true \
-        --register-with-taints=node-role.kubernetes.io/master=:NoSchedule \
+        --register-with-taints=node-role.kubernetes.io/master=:NoSchedule,node-role.kubernetes.io/control-plane=:NoSchedule\
         --kubeconfig=/etc/kubernetes/kubeconfig/kubelet.yaml \
         --node-labels="node.kubernetes.io/master,role=master,ip=${DEFAULT_IPV4},{{.Cluster.Kubernetes.Kubelet.Labels}}" \
```
whites11, line 355: maybe want to add the control-plane label as well?

mogottsch: good catch, I've added it :+1:

files/conf/k8s-addons
```diff
@@ -26,7 +26,6 @@ done
 SECURITY_FILES=""
 SECURITY_FILES="${SECURITY_FILES} rbac_bindings.yaml"
 SECURITY_FILES="${SECURITY_FILES} rbac_roles.yaml"
-SECURITY_FILES="${SECURITY_FILES} psp_policies.yaml"
 SECURITY_FILES="${SECURITY_FILES} psp_roles.yaml"
```
calvix, line 29: is there a reason why we dont remove all PSP related resources?

files/conf/k8s-addons
```diff
@@ -26,7 +26,6 @@ done
...
 SECURITY_FILES="${SECURITY_FILES} psp_binding.yaml"
```
calvix, line 30: is there a reason why we dont remove all PSP related resources?

PR #1139, opened 2023-08-17
Update containerd-config.toml

njuettner:
This needs to change with Kernel 6.x.

It will break loading the configuration.

```
containerd: failed to load TOML from /etc/containerd/config.toml: invalid plugin key URI "containerd.runtime.v1.linux" expect io.containerd.x.vxInvalid 
```

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,11 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 - Enable auditd unit.
 
+### Changed
+
+- Changed plugin key containerd.runtime.v1.linux in containerd configuration for release 1.7.x.
+
+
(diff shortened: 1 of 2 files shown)


calvix:
is this compatible with older kernels? If yes we should do the same change in cluster-aws so we dont hit this later on
https://github.com/giantswarm/cluster-aws/blob/master/helm/cluster-aws/files/etc/containerd/config.toml#L9

calvix (approved):

njuettner:
yes it is compatible

njuettner:
I tested it with containerd 1.6.x and 1.7.x in kernel 5.x and kernel 6.x

whites11 (approved):

PR #1155, opened 2023-10-17
K8s 1.25

whites11:
This PR will be needed when we'll need legacy release v20 with k8s 1.25.

Please note the changes in this PR were already released as part of k8scc v17 release series and master branch's history was rewritten afterwards.

see https://gigantic.slack.com/archives/C0559SH3RJ4/p1697464874504409

## Checklist


files/manifests/k8s-api-server.yaml.tmpl
```diff
@@ -38,8 +38,8 @@ spec:
     - --profiling=false
     - --service-account-lookup=true
     - --authorization-mode=RBAC
-    - --feature-gates=TTLAfterFinished=true{{ if .EnableCronJobTimeZone }},CronJobTimeZone=true{{ end }}
-    - --enable-admission-plugins=NamespaceLifecycle,LimitRanger,ServiceAccount,ResourceQuota,DefaultStorageClass,PersistentVolumeClaimResize,PodSecurityPolicy,Priority,DefaultTolerationSeconds,MutatingAdmissionWebhook,ValidatingAdmissionWebhook
+    {{ if .EnableCronJobTimeZone }}- --feature-gates=CronJobTimeZone=true{{ end }}
```
njuettner, line 41: I think we don't need to add it anymore, it's by default in 1.25
image

njuettner (approved):

files/manifests/k8s-api-server.yaml.tmpl
whites11: good catch, but this and other vintage repos will be a thing of the past soon so I guess I'll ignore it

